PR #7, opened 2018-07-30
ref chameleon-system/chameleon-system#5: hide old media properties field

pixeljunkie:
| Q             | A
| ------------- | ---
| Branch?       | 6.2.x
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | yes
| Tests pass?   | yes
| Fixed issues  | chameleon-system/chameleon-system#5
| License       | MIT

hides old media properties field, because the same information is already shown in the detail view of media files


umulmrum:
The new media manager is intended as drop-in replacement, but also with the option to "drop out" so that the previous implementation can still be used. This change would remove essential information for the old implementation - can you think of a somewhat simple solution that does not change database state?

src/CoreBundle/Resources/config/services.xml
```diff
@@ -11,6 +11,7 @@
         <parameter key="chameleon_system_core.database.estimation_lower_limit">0</parameter>
         <parameter key="chameleon_system_core.controller_invoke_listener.class">ChameleonSystem\CoreBundle\Controller\ChameleonControllerInvokeListener</parameter>
         <parameter key="chameleon_system_core.tpl_page_controller.class">TPkgCmsRouteControllerCmsTplPage</parameter>
+        <parameter key="chameleon_system_media_manager.render_media_properties_field">true</parameter>
```
umulmrum: This is a parameter of chameleon_system_core and only modified in the media manager. Could you set the prefix to chameleon_system_core?

pixeljunkie: You are right. Renamed the parameter.

umulmrum:
Thanks @pixeljunkie

PR #14, opened 2018-08-09
Fix/17 replace mappers

umulmrum:
| Q             | A
| ------------- | ---
| Branch?       | 6.2.x
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed issues  | chameleon-system/chameleon-system#17
| License       | MIT

With this PR we can handle incomplete mapper configuration. I'd prefer to avoid invalid data being written to database, but that would mean quite a bit of effort.


umulmrum:
@rnuernberg I think logging isn't the right thing to do here. The log message is a very indirect way of telling the user that there is a problem; the log could even be read by another person as the one that introduced the wrong config (apart from cluttering the log with semi-relevant information). A better solution would be to validate input on save and present an error message there; as time is short, this fix simply leaves invalid input as is and writes invalid entries back.

umulmrum:
As I now know, having an entry that only contains a name without further config is required for legacy modules in a theme. Therefore this is not a configuration error and should not be logged or aborted.

Line 85 should also be OK - when this line is called, the config line only consists of the name so that no information is lost.

PR #27, opened 2018-08-29
Fix/46 case insensitive portal prefix

umulmrum:
| Q             | A
| ------------- | ---
| Branch        | 6.2.x
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed issues  | chameleon-system/chameleon-system#46
| License       | MIT

There will be a separate PR for Chameleon 6.3 which will also contain a unit test (and minor BC breaks).


src/CoreBundle/EventListener/CaseInsensitivePortalExceptionListener.php
```diff
@@ -0,0 +1,121 @@
...
+                $url = $this->replacePortalPrefix($relativePath, $portalPrefixCandidate, $prefix);
+                $code = 'POST' === $request->getMethod() ? Response::HTTP_TEMPORARY_REDIRECT : Response::HTTP_MOVED_PERMANENTLY;
+                $event->setResponse(new RedirectResponse($url, $code));
+                break;
+            }
+        }
+    }
+
+    /**
+     * Returns the path fragment between first and second backslash (or everything after the first slash if there is no
```
bestform: Don't you mean "slash"?

umulmrum: Yep :-) - fixed

PR #28, opened 2018-08-30
ref #48: Restructure portal and domain initialization

umulmrum:
| Q             | A
| ------------- | ---
| Branch        | master
| Bug fix?      | no
| New feature?  | ni
| BC breaks?    | yes
| Deprecations? | no
| Tests pass?   | yes
| Fixed issues  | chameleon-system/chameleon-system#48
| License       | MIT

Database access moved to the data access class; removed some code that was superfluous.

PR #29, opened 2018-08-30
Feature/49 variable replacement

umulmrum:
| Q             | A
| ------------- | ---
| Branch        | master
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | yes
| Tests pass?   | yes
| Fixed issues  | chameleon-system/chameleon-system#49
| License       | MIT

Branch is based on #48, so please review this first.

### DIFF
--- a/UPGRADE-6.3.md
+++ b/UPGRADE-6.3.md
@@ -35,3 +35,55 @@ Note that ONLY BC breaking changes are listed, according to our backwards compat
 - New method `getActivePortalCandidate()`.
 - New method `getDomainDataByName()`.
 - New method `getPortalPrefixListForDomain()`.
+
+# Deprecated Code Entities
+
+It is recommended that all references to the classes, interfaces, properties, constants, methods and services in the
+following list are removed from the project, as they will be removed in Chameleon 7.0. The deprecation notices in the
+code will tell if there are replacements for the deprecated entities or if the functionality is to be entirely removed.
+
+To search for deprecated code usage, [SensioLabs deprecation detector](https://github.com/sensiolabs-de/deprecation-detector)
+is recommended (although this tool may not find database-related deprecations).
+
+## Services
+
+None.
+
+## Container Parameters
+
+None.
+
+## Constants
+
+None.
+
+## Classes and Interfaces
+
+None.
+
+## Properties
+
+- \ChameleonSystem\CoreBundle\Controller\ChameleonController::$sGeneratedPage
+- \ChameleonSystem\CoreBundle\Controller\ChameleonController::$postRenderVariables
+
+## Methods
+
+- \ChameleonSystem\CoreBundle\Controller\ChameleonController::GetPostRenderVariables()
+- \ChameleonSystem\CoreBundle\Controller\ChameleonController::PreOutputCallbackFunctionReplaceCustomVars()
+- \TTools::AddStaticPageVariables()
+
+## JavaScript Files and Functions
+
+None.
+
+## Translations
+
+None.
+
+## Database Tables
+
+None.
+
+## Database Fields
+
+None.
--- a/UPGRADE-6.3.md
+++ b/UPGRADE-6.3.md
@@ -5,32 +5,14 @@ UPGRADE FROM 6.2 TO 6.3
 
 ## Changed Signatures
 
-See the Changed Interfaces and Method Signatures section if changes in signatures affect the project.
+See the Changed Interfaces and Method Signatures section whether changes in signatures affect the project.
 
 # Changed Interfaces and Method Signatures
 
 This section contains information on interface and method signature changes which affect backwards compatibility (BC).
 Note that ONLY BC breaking changes are listed, according to our backwards compatibility policy.
 
-## ChameleonSystem\CoreBundle\DataAccess\CmsPortalDomainsDataAccess
-
-- New method `getActivePortalCandidate()`.
-- New method `getDomainDataByName()`.
-- New method `getPortalPrefixListForDomain()`.
-
-## ChameleonSystem\CoreBundle\DataAccess\CmsPortalDomainsDataAccessCacheDecorator
-
-- New method `getActivePortalCandidate()`.
-- New method `getDomainDataByName()`.
-- New method `getPortalPrefixListForDomain()`.
-
 ## ChameleonSystem\CoreBundle\DataAccess\CmsPortalDomainsDataAccessInterface
-
-- New method `getActivePortalCandidate()`.
-- New method `getDomainDataByName()`.
-- New method `getPortalPrefixListForDomain()`.
-
-## ChameleonSystem\CoreBundle\DataAccess\CmsPortalDomainsDataAccessRequestLevelCacheDecorator
 
 - New method `getActivePortalCandidate()`.
 - New method `getDomainDataByName()`.
--- a/src/CoreBundle/Controller/ChameleonController.php
+++ b/src/CoreBundle/Controller/ChameleonController.php
@@ -13,7 +13,9 @@ namespace ChameleonSystem\CoreBundle\Controller;
 
 use ChameleonSystem\CoreBundle\CoreEvents;
 use ChameleonSystem\CoreBundle\Event\HtmlIncludeEvent;
+use ChameleonSystem\CoreBundle\Response\ResponseVariableReplacerInterface;
 use ChameleonSystem\CoreBundle\Security\AuthenticityToken\AuthenticityTokenManagerInterface;
+use ChameleonSystem\CoreBundle\Security\AuthenticityToken\TokenInjectionFailedException;
 use ChameleonSystem\CoreBundle\Service\ActivePageServiceInterface;
 use ChameleonSystem\CoreBundle\Service\PortalDomainServiceInterface;
 use ChameleonSystem\CoreBundle\Service\RequestInfoServiceInterface;
@@ -27,7 +29,6 @@ use Symfony\Component\HttpFoundation\Request;
 use Symfony\Component\HttpFoundation\RequestStack;
 use Symfony\Component\HttpFoundation\Response;
 use Symfony\Component\HttpKernel\Exception\NotFoundHttpException;
-use TCMSMessageManager;
 use TCMSPageDefinitionFile;
 use TCMSResourceCollection;
 use TGlobal;
@@ -69,6 +70,8 @@ abstract class ChameleonController implements ChameleonControllerInterface
     protected $redirectPageDef;
     /**
      * @var array
+     *
+     * @deprecated since 6.3.0 - not used anymore
      */
     protected $postRenderVariables;
     /**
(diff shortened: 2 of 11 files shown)


UPGRADE-6.3.md
```diff
@@ -0,0 +1,89 @@
+UPGRADE FROM 6.2 TO 6.3
+=======================
+
+# Essentials
+
+## Changed Signatures
+
+See the Changed Interfaces and Method Signatures section if changes in signatures affect the project.
```
bestform: if = whether

UPGRADE-6.3.md
```diff
@@ -0,0 +1,89 @@
...
+
+- New method `getActivePortalCandidate()`.
+- New method `getDomainDataByName()`.
+- New method `getPortalPrefixListForDomain()`.
+
+## ChameleonSystem\CoreBundle\DataAccess\CmsPortalDomainsDataAccessRequestLevelCacheDecorator
+
+- New method `getActivePortalCandidate()`.
+- New method `getDomainDataByName()`.
+- New method `getPortalPrefixListForDomain()`.
```
bestform: The new methods in the implementations do not really break BC, do they? If we just want to list BC breaking changes, only the interfaces are relevant.

src/CoreBundle/Controller/ChameleonController.php
```diff
@@ -542,7 +551,7 @@ public function PreOutputCallbackFunction(&$sPageContent)
                 $bHeaderParsed = true;
             }
         }
-        $sPageContent = $this->PreOutputCallbackFunctionReplaceCustomVars($sPageContent);
+        $sPageContent = $this->responseVariableReplacer->replaceVariables($sPageContent);
```
bestform: `$this->responseVariableReplacer` can be `null` as it is injected via a setter injection.

src/CoreBundle/Controller/ChameleonController.php
```diff
@@ -693,29 +702,14 @@ protected function runExternalResourceCollectorOnPageContent($sPageContent)
...
-            $sPageContent = $this->authenticityTokenManager->addTokenToForms($sPageContent);
-            $aPageVariables = $this->GetPostRenderVariables();
-            if (is_array($aPageVariables) && count($aPageVariables) > 0) {
-                $oStringReplace = new \TPkgCmsStringUtilities_VariableInjection();
-                $sPageContent = $oStringReplace->replace($sPageContent, $aPageVariables);
-            }
-        }
-
-        return $sPageContent;
+        return $this->responseVariableReplacer->replaceVariables($sPageContent);
```
bestform: can be `null`

src/CoreBundle/DataAccess/CmsPortalDomainsDataAccess.php
```diff
@@ -52,4 +50,84 @@ public function getPrimaryDomain($portalId, $languageId)
...
+        $query = 'SELECT `cms_portal`.`identifier`
+                    FROM `cms_portal_domains`
+              INNER JOIN `cms_portal` ON `cms_portal_domains`.`cms_portal_id` = `cms_portal`.`id`
+                   WHERE `cms_portal_domains`.`name` = ? OR `cms_portal_domains`.`sslname` = ?
+                GROUP BY `cms_portal_domains`.`cms_portal_id`
+               ';
+
+        $result = $this->connection->fetchAll($query, [
+            $domainName,
+            $domainName,
```
bestform: Is there reason we do not use named placeholders here?

src/CoreBundle/DataAccess/CmsPortalDomainsDataAccess.php
```diff
@@ -52,4 +50,84 @@ public function getPrimaryDomain($portalId, $languageId)
...
+
+         if (false === $allowInactivePortals) {
+             $query .= " AND `cms_portal`.`deactive_portal` != '1' ";
+         }
+         $query .= ' ORDER BY `identifier` DESC
+                       LIMIT 0,1';
+
+         $portalCandidate = $this->connection->fetchAssoc($query, [
+             $idRestrictionList,
+             $identifierRestriction,
```
bestform: Same question here

src/CoreBundle/DataAccess/CmsPortalDomainsDataAccessRequestLevelCacheDecorator.php
```diff
@@ -47,4 +47,28 @@ public function getPrimaryDomain($portalId, $languageId)
...
+        return $this->subject->getActivePortalCandidate($idRestrictionList, $identifierRestriction, $allowInactivePortals);
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function getDomainDataByName(string $domainName): array
+    {
+        return $this->subject->getDomainDataByName($domainName);
+    }
```
bestform: There is not much request level caching going on in here. Is this intentionally?

src/CoreBundle/Response/ResponseVariableReplacerInterface.php
```diff
@@ -0,0 +1,40 @@
...
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace ChameleonSystem\CoreBundle\Response;
+
+use ChameleonSystem\CoreBundle\Security\AuthenticityToken\TokenInjectionFailedException;
+
+interface ResponseVariableReplacerInterface
```
bestform: We usually add a doc comment to the interface declaration describing an overall purpose.

src/CoreBundle/Response/ResponseVariableReplacerInterface.php
```diff
@@ -0,0 +1,40 @@
...
+{
+    /**
+     * Add a variable that can then be replaced by replaceVariables(). Note that variables MUST alreday be escaped by
```
bestform: *already

src/CoreBundle/Response/ResponseVariableReplacerInterface.php
```diff
@@ -0,0 +1,40 @@
...
+     * the caller, as implementations of this interface do not know the context in which the variables are used and are
+     * therefore unable to perform the correct escaping.
```
bestform: In the light of this comment we could discuss the name of the interface as `Response` kind of sets a context already.

src/CoreBundle/Response/ResponseVariableReplacerInterface.php
```diff
@@ -0,0 +1,40 @@
...
+     *
+     * @param string $key
+     * @param string $value
+     */
+    public function addVariable(string $key, string $value): void;
```
bestform: I do not have an easy solution for this but with this interface we force a state which leads to a stateful service if used as such by an implementation.

src/CoreBundle/Tests/Response/ResponseVariableReplacerTest.php
```diff
@@ -0,0 +1,173 @@
...
+    {
+        $this->actualResult = $this->subject->replaceVariables($input);
+    }
+
+    private function thenTheExpectedResultShouldBeReturned($expectedResult): void
+    {
+        $this->assertEquals($expectedResult, $this->actualResult);
+    }
+
+    public function testTokenInjectionFailedException(): void
```
bestform: Please move this test below the other one so one can see all the tests side by side.

UPGRADE-6.3.md
umulmrum: done

src/CoreBundle/Response/ResponseVariableReplacerInterface.php
umulmrum: done

src/CoreBundle/Response/ResponseVariableReplacerInterface.php
umulmrum: done

src/CoreBundle/Response/ResponseVariableReplacerInterface.php
umulmrum: The name was chosen on purpose - the service adds the authenticity token (and throws a TokenInjectionFailedException on failure), which limits its scope to responses. 

Also having an addVariable() method in the interface does only make sense if the service is called on multiple independent pieces of client code. I wouldn't add the method on any generic replacer service (but leave this responsibility to implementations), so this also emphasizes the strategic (as opposed to generic) use of the interface.

Of course we can discuss if the name can be improved, but I wouldn't remove the purpose indicator.

src/CoreBundle/Response/ResponseVariableReplacerInterface.php
umulmrum: See my comment above - I think this is OK (and suppose we can't get around state here anyway).

src/CoreBundle/Tests/Response/ResponseVariableReplacerTest.php
umulmrum: done (moved it under the data provider of the other test, as test and provider shouldn't be separated). I think we should discuss the desired order again when we find the time.

UPGRADE-6.3.md
umulmrum: I just saw that this change belongs to PR #28 - could you have a look at that PR before merging this one (as requested in the PR description)? Thanks!

src/CoreBundle/DataAccess/CmsPortalDomainsDataAccessRequestLevelCacheDecorator.php
umulmrum: This comment isn't shown in the discussion ..?

This is indeed by intention - these methods are quite specific for initialization and are only called once per request. Adding a request level cache wouldn't improve performance but cost additional memory, so I didn't do this.

src/CoreBundle/Controller/ChameleonController.php
bestform: Then why isn't this dependency injected into the constructor?

src/CoreBundle/Controller/ChameleonController.php
bestform: same

UPGRADE-6.3.md
bestform: OK, I've moved my comment there and will resolve this.

src/CoreBundle/Controller/ChameleonController.php
umulmrum: Subclasses rely on the constructor as-is and would break. This is how we solved it for every new dependency in recent times (and we discuss the topic every time ;-) ) - maybe we should find a better solution in general, but not in this PR.

src/CoreBundle/Controller/ChameleonController.php
umulmrum: same ^^

src/CoreBundle/Controller/ChameleonController.php
bestform: We should :)

bestform:
Could you resolve the conflict? After this is done the PR is ready to be merged

umulmrum:
Done

PR #30, opened 2018-08-30
Feature/46 improved case insensitivity for portal prefix

umulmrum:
| Q             | A
| ------------- | ---
| Branch        | master
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed issues  | chameleon-system/chameleon-system#46
| License       | MIT

Based on #29, so please review that PR first.

An improvement for the fix for chameleon-system/chameleon-system#46, which uses a new method that caused a slight BC break. In 6.3.x only this fix should be used.

### DIFF
--- a/src/CoreBundle/EventListener/CaseInsensitivePortalExceptionListener.php
+++ b/src/CoreBundle/EventListener/CaseInsensitivePortalExceptionListener.php
@@ -62,7 +62,7 @@ class CaseInsensitivePortalExceptionListener
     }
 
     /**
-     * Returns the path fragment between first and second backslash (or everything after the first slash if there is no
+     * Returns the path fragment between first and second slash (or everything after the first slash if there is no
      * second one) - this might be a portal prefix.
      *
      * @param string $relativePath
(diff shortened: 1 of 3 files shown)


src/CoreBundle/EventListener/CaseInsensitivePortalExceptionListener.php
```diff
@@ -0,0 +1,89 @@
...
+                $url = $this->replacePortalPrefix($relativePath, $portalPrefixCandidate, $prefix);
+                $code = 'POST' === $request->getMethod() ? Response::HTTP_TEMPORARY_REDIRECT : Response::HTTP_MOVED_PERMANENTLY;
+                $event->setResponse(new RedirectResponse($url, $code));
+                break;
+            }
+        }
+    }
+
+    /**
+     * Returns the path fragment between first and second backslash (or everything after the first slash if there is no
```
bestform: "backslash"

umulmrum: Ah I copy&pasted that from the 6.2.x fix :-) - changed

PR #33, opened 2018-09-04
ref #55: Check memcache after initialization

UlrichKu:
| Q             | A
| ------------- | ---
| Branch        | 6.2.x
| Bug fix?      | yes
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed issues  | chameleon-system/chameleon-system#55
| License       | MIT

### DIFF
--- a/src/CoreBundle/Exception/MemcachedAccessFailedException.php
+++ b/src/CoreBundle/Exception/MemcachedAccessFailedException.php
@@ -1,0 +1,8 @@
+<?php
+
+namespace ChameleonSystem\CoreBundle\Exception;
+
+class MemcachedAccessFailedException extends \Exception
+{
+
+}
--- a/src/CoreBundle/Exception/SessionAccessFailedException.php
+++ b/src/CoreBundle/Exception/SessionAccessFailedException.php
@@ -1,5 +1,14 @@
 <?php
 
+/*
+ * This file is part of the Chameleon System (https://www.chameleonsystem.com).
+ *
+ * (c) ESONO AG (https://www.esono.de)
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
 namespace ChameleonSystem\CoreBundle\Exception;
 
 class SessionAccessFailedException extends \Exception
--- a/src/CoreBundle/Exception/SessionAccessFailedException.php
+++ b/src/CoreBundle/Exception/SessionAccessFailedException.php
@@ -13,5 +13,4 @@ namespace ChameleonSystem\CoreBundle\Exception;
 
 class SessionAccessFailedException extends \Exception
 {
-
 }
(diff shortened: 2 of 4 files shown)


src/CoreBundle/Exception/MemcachedAccessFailedException.php
```diff
@@ -0,0 +1,8 @@
+<?php
+
+namespace ChameleonSystem\CoreBundle\Exception;
+
+class MemcachedAccessFailedException extends \Exception
```
umulmrum: Just one remark by me: I'd prefer to have a SessionAccessFailedException or similar - the fact that the session uses memcached shouldn't bleed into client code as this is an implementation detail that may change anytime.

UlrichKu: done

src/CoreBundle/Exception/SessionAccessFailedException.php
```diff
@@ -0,0 +1,8 @@
+<?php
+
```
umulmrum: Could you add the default license header?

UlrichKu: ok

src/CoreBundle/Session/ChameleonSessionManager.php
```diff
@@ -174,6 +179,14 @@ public function boot()
         $request->getSession()->start();
         $this->isSessionStarting = false;
         $this->registerSessionInContainer($request->getSession());
+
+        if (null !== $memcache && null !== $memcache->getDriver()) {
+            $sessionInitResult = $memcache->getDriver()->getResultCode();
```
rnuernberg: I don't think, that this is the correct place to check if memcache is available. The user may well have choosen to use the database a session storage.

Instead, I would suggest moving the code into the boot method where the memcache connection is opened.

UlrichKu: This method here is called boot() (?)

rnuernberg: that is true :) but take a look at line 103

UlrichKu: If I remember correctly this location was chosen to be placed "after initialization - which is happening outside of the scope of this method".

It checks if memcache is actually used before doing a check (null !== $memcache).

rnuernberg: I am still not ok with this. for one, checking if memcache is null or not is not the same as checking if the user plans on using memcache for session storage (even if that may currently act the same).

I think we should solve this problem one of two ways:

We have the sanity check bundle and use it to check if the session memcache is available and works
Or we check if the where able to fetch/get data from memcache when accessing memcache - if the request returns an error that indicates, that the server has gone away, we can throw the exception - of course only for session access, not for cache access.

But doing the check here is not helpful. Either the configuration is invalid - which we should not check on every single call, or the memcache server has gone away - which we should check when accessing it.

PR #39, opened 2018-09-06
ref #63: Add configuration to disable script tags in text fields glob…

umulmrum:
…ally

| Q             | A
| ------------- | ---
| Branch        | 6.2.x (although this is a feature, it should be available as fast as possible, and in the default config it doesn't alter system behavior)
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | 
| Tests pass?   | yes
| Fixed issues  | chameleon-system/chameleon-system#63
| License       | MIT

### DIFF
--- a/src/CmsTextFieldBundle/DependencyInjection/Configuration.php
+++ b/src/CmsTextFieldBundle/DependencyInjection/Configuration.php
@@ -1,0 +1,40 @@
+<?php
+
+/*
+ * This file is part of the Chameleon System (https://www.chameleonsystem.com).
+ *
+ * (c) ESONO AG (https://www.esono.de)
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace ChameleonSystem\CmsTextFieldBundle\DependencyInjection;
+
+use Symfony\Component\Config\Definition\Builder\TreeBuilder;
+use Symfony\Component\Config\Definition\ConfigurationInterface;
+
+class Configuration implements ConfigurationInterface
+{
+    public function getConfigTreeBuilder()
+    {
+        $treeBuilder = new TreeBuilder();
+        $root = $treeBuilder->root('chameleon_system_cms_text_field');
+        $root
+            ->isRequired()
+            ->addDefaultsIfNotSet();
+
+        $rootq
+            ->children()
+                ->booleanNode('allow_script_tags')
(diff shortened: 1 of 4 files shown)


umulmrum:
I had to add a container parameter as there is no DI infrastructure for TCMSTextFieldEndPoint. But the feature is configured by container configuration, so users do not get in contact with the parameter and we should be able to remove it in the future.

src/CmsTextFieldBundle/DependencyInjection/Configuration.php
```diff
@@ -0,0 +1,40 @@
...
+        $root
+            ->isRequired()
+            ->addDefaultsIfNotSet();
+
+        $root
+            ->children()
+                ->booleanNode('allow_script_tags')
+                    ->defaultTrue()
+                ->info('If set to false, script tags are removed on display. This is a security feature: 
+                If an attacker somehow manages to gain access to the backend, they still cannot inject scripts.
```
bestform: I'd remove the 'still' from the sentence.

src/CmsTextFieldBundle/objects/TCMSTextFieldEndPoint.class.php
```diff
@@ -1232,4 +1232,9 @@ protected function isForceThumbnailGenerationOnFullSizeImagesEnabled()
     {
         return false;
     }
+
+    private function doesAllowScriptTags(): bool
```
bestform: Why is this method called `doesAllowScriptTags` instead of `allowScriptTags`? This is somewhat unconventional.

umulmrum: `allowScriptTags` would (in code context as opposed to the config parameter name) imply an action that makes script takes allowed - if possible methods returning booleans should have an `is` or `has` prefix. I changed the method name to `isScriptTagAllowed` as the plural isn't really required.

src/CmsTextFieldBundle/DependencyInjection/Configuration.php
umulmrum: done

src/CmsTextFieldBundle/objects/TCMSTextFieldEndPoint.class.php
bestform: perfect

PR #49, opened 2018-09-25
ref #82: Use portal infix also for resource collection files

UlrichKu:
| Q             | A
| ------------- | ---
| Branch        |  6.2.x
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed issues  | chameleon-system/chameleon-system#82
| License       | MIT

The portal (id) is now considered in resource collection file names: Content can be different and there can be links with domain name inside.

### DIFF
--- a/src/CoreBundle/private/library/classes/TCMSResourceCollection.class.php
+++ b/src/CoreBundle/private/library/classes/TCMSResourceCollection.class.php
@@ -11,6 +11,7 @@
 
 use ChameleonSystem\CoreBundle\CoreEvents;
 use ChameleonSystem\CoreBundle\Event\ResourceCollectionJavaScriptCollectedEvent;
+use ChameleonSystem\CoreBundle\Service\PortalDomainServiceInterface;
 use ChameleonSystem\CoreBundle\ServiceLocator;
 use Symfony\Component\EventDispatcher\EventDispatcherInterface;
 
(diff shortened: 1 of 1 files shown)


pixeljunkie:
> I'm not entirely sure if we can do this change in a fix version as it changes the file naming scheme. Client code could in principle depend on the current scheme, but chances are quite high that this is not the case. WDYT @bestform?

I`m very sure that we don`t rely on the file name because the path to the file is inserted by the core itself to the header. I can not imagine a situation where you would import or do anything with the filename. It even changes with every deploy, so this would be impossible anyway imho. I would vote for this change even in a fix release.

bestform:
As the files are generated, there is the possibility that some sort of external script depends on the scheme.

I think though, it is relatively save to release this in a fix release. The use case where this would break an installation is very far fetched.

src/CoreBundle/private/library/classes/TCMSResourceCollection.class.php
```diff
@@ -148,6 +147,23 @@ public function CollectExternalResources($sPageContent)
...
     }
 
+    private function getFilesRefreshAndDomainPrefix(): string
+    {
+        $filesPrefix = ServiceLocator::getParameter('chameleon_system_core.resources.enable_external_resource_collection_refresh_prefix');
+
+        /**
+         * @var $portalDomainService PortalDomainServiceInterface
+         */
+        $portalDomainService = ServiceLocator::get('chameleon_system_core.portal_domain_service');
```
umulmrum: Could you move the ServiceLocator call to a separate method that includes a return type declaration? This improves type safety and keeps the distracting details of service retrieval away.

UlrichKu: ok

umulmrum:
Thanks @UlrichKu

PR #51, opened 2018-09-26
fix/85: undefined variable $_SESSION

TorbenBr:
| Q             | A
| ------------- | ---
| Branch        | 6.2.x
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no     
| Deprecations? | no 
| Tests pass?   | no
| Fixed issues  | chameleon-system/chameleon-system#85
| License       | MIT

### DIFF
--- a/src/CoreBundle/private/library/classes/TCMSLogChange.class.php
+++ b/src/CoreBundle/private/library/classes/TCMSLogChange.class.php
@@ -1604,7 +1604,7 @@ class TCMSLogChange
         try {
             $connection->executeQuery($query);
         } catch (DBALException $e) {
-            throw new ErrorException($e->getMessage(), 0 , $e);
+            throw new ErrorException($e->getMessage(), 0, $e);
         }
 
         $oManager->UpdateVirtualClasses();
--- a/src/CoreBundle/private/library/classes/TCMSLogChange.class.php
+++ b/src/CoreBundle/private/library/classes/TCMSLogChange.class.php
@@ -1599,10 +1599,12 @@ class TCMSLogChange
 
         $connection = self::getDatabaseConnection();
 
-        $query = 'DELETE FROM '.$connection->quoteIdentifier($oExtension->table).' WHERE id = '.$connection->quote($oExtension->id);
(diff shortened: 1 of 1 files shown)


src/CoreBundle/private/library/classes/TCMSLogChange.class.php
```diff
@@ -1585,6 +1585,7 @@ public static function AddVirtualNonDbExtension($iLine, $sEntryPoint, $sClassNam
      * @param string $sClassName
      *
      * @throws ErrorException
+     * @throws DBALException
```
umulmrum: This changes the method signature which is not allowed in a fix release. Could you wrap the exception in an ErrorException (which I normally wouldn't recommend, but here it's the only annotated exception)?

src/CoreBundle/private/library/classes/TCMSLogChange.class.php
```diff
@@ -1597,9 +1598,10 @@ public static function deleteVirtualNonDbExtension($iLine, $sEntryPoint, $sClass
             throw new ErrorException("unable to find virtual extension {$sClassName} for {$sEntryPoint} (called from the running update in line {$iLine})", 0, E_USER_ERROR, __FILE__, __LINE__);
         }
 
-        $tableEditorManager = TTools::GetTableEditorManager($oExtension->table, $oExtension->id);
-        $tableEditorManager->AllowDeleteByAll(true);
-        $tableEditorManager->Delete($oExtension->id);
+        $connection = self::getDatabaseConnection();
+
+        $query = 'DELETE FROM '.$connection->quoteIdentifier($oExtension->table).' WHERE id = '.$connection->quote($oExtension->id);
```
umulmrum: Could we work with a parameter for the ID? Although we need string concatenation for the table name, we should use this as little as possible.

src/CoreBundle/private/library/classes/TCMSLogChange.class.php
TorbenBr: done

umulmrum:
Thanks @TorbenBr

PR #53, opened 2018-10-01
ref #111: Reset modules in case this is some kind of nested request (…

umulmrum:
…e.g. error page as called by ChameleonSystem\CoreBundle\Controller\ExceptionController)

| Q             | A
| ------------- | ---
| Branch        | 6.2.x
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | no
| Fixed issues  | chameleon-system/chameleon-system#111
| License       | MIT

The fix was surprisingly simple, so I suspect I might have missed something. The LoadModules() method is the first one called by ChameleonController and the one in which modules are added, so I think it fits "naturally" without having to care about the context in which the method is called.

### DIFF
--- a/src/CoreBundle/Resources/config/services.xml
+++ b/src/CoreBundle/Resources/config/services.xml
@@ -162,6 +162,8 @@
         </service>
         <service id="chameleon_system_core.usermoduleloader" class="TUserModuleLoader" parent="chameleon_system_core.moduleloader" public="true" />
 
+        <service id="chameleon_system_core.subusermoduleloader" parent="chameleon_system_core.usermoduleloader" shared="false" public="true" />
+
         <service id="chameleon_system_core.base_controller" abstract="true">
             <argument type="service" id="request_stack"/>
             <argument type="service" id="event_dispatcher" />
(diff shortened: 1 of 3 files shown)


umulmrum:
Aaaannd it's too easy :-) - needs additional work.

umulmrum:
I added a second commit that fetches separate module loader instances in TTools. Otherwise modules would get erased when a multi module is loaded.

I'm not sure if this change isn't too bold. I think it is the correct thing to do because it more or less restores the implementation with a separate module loader from a few years ago. I think the current implementation wasn't really intended but crept in when we defined a Symfony service for the module loader.

That said I don't know if client code that relies on the current behavior was added in the last years. It might be better to be conservative and move this change to 6.3.0 and add a notice in the upgrade info. WDYT?

bestform:
Hard to say. The solution is pretty clean and I don't think that any code should depend on the old version. 
That said, there is the possibility so if we want to be strict about such things - which we probably should be - then a 6.3.0 release would be saver.

umulmrum:
OK, maybe it's better to be a bit conservative. I'll propose another PR for master.

umulmrum:
New PR for master: https://github.com/chameleon-system/chameleon-base/pull/59

PR #55, opened 2018-10-08
ref #116: Add command for counting pending updates

UlrichKu:
| Q             | A
| ------------- | ---
| Branch        | 6.2.x
| Bug fix?      | no
| New feature?  | yes| BC breaks?    | no
| Deprecations? | no
| Fixed issues  | chameleon-system/chameleon-system#116
| License       | MIT

### DIFF
--- a/src/CoreBundle/Command/UpdateCountCommand.php
+++ b/src/CoreBundle/Command/UpdateCountCommand.php
@@ -1,0 +1,51 @@
+<?php
+
+/*
+ * This file is part of the Chameleon System (https://www.chameleonsystem.com).
+ *
+ * (c) ESONO AG (https://www.esono.de)
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace ChameleonSystem\CoreBundle\Command;
+
+use Symfony\Component\Console\Command\Command;
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Output\OutputInterface;
+
+/**
+ * Console command for counting updates that have not been executed yet.
+ */
+class UpdateCountCommand extends Command
+{
+    public function __construct($name = null)
+    {
+        parent::__construct('chameleon_system:update:count');
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    protected function configure()
+    {
+        $this
+            ->setDefinition([])
+            ->setDescription('Prints the count of all pending updates')
+        ;
+    }
+
+    /**
+     * {@inheritdoc}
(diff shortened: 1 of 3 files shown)


src/CoreBundle/Command/UpdateCountCommand.php
```diff
@@ -0,0 +1,51 @@
...
+namespace ChameleonSystem\CoreBundle\Command;
+
+use Symfony\Component\Console\Command\Command;
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Output\OutputInterface;
+
+/**
+ * Console command for counting updates that have not been executed yet.
+ */
+class UpdateCountCommand extends Command
```
umulmrum: Could we rename the command to CountUpdateCommand to be in line with the existing commands (although your scheme would be better if we had used it from the beginning; in the future I'd like to move these commands to the DatabaseMigrationBundle where the current scheme would be OK)?

src/CoreBundle/Command/UpdateCountCommand.php
```diff
@@ -0,0 +1,51 @@
...
+        parent::__construct('chameleon_system:update:count');
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    protected function configure()
+    {
+        $this
+            ->setDefinition([])
```
umulmrum: I think this call can be removed.

src/CoreBundle/Command/UpdateCountCommand.php
```diff
@@ -0,0 +1,51 @@
...
+            ->setDescription('Prints the count of all pending updates')
```
umulmrum: Please begin the description with "The <info>%command.name%</info> command " and end with a dot.

src/CoreBundle/Command/UpdateCountCommand.php
UlrichKu: ok

src/CoreBundle/Command/UpdateCountCommand.php
UlrichKu: html help added

src/CoreBundle/Command/UpdateCountCommand.php
UlrichKu: ok

umulmrum:
I forgot another thing, sorry. Could you add a service definition for the command? Although it works as is, we always define commands as services (which will also be required in future Symfony versions).

UlrichKu:
Service definition added

umulmrum:
Thanks @UlrichKu

PR #56, opened 2018-10-09
117 configure less path

UlrichKu:
| Q             | A
| ------------- | ---
| Branch        | 6.2.x
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Fixed issues  | chameleon-system/chameleon-system#117
| License       | MIT

### DIFF
--- a/src/ViewRendererBundle/DependencyInjection/Configuration.php
+++ b/src/ViewRendererBundle/DependencyInjection/Configuration.php
@@ -1,0 +1,28 @@
+<?php
+
+namespace ChameleonSystem\ViewRendererBundle\DependencyInjection;
+
+use Symfony\Component\Config\Definition\Builder\TreeBuilder;
+use Symfony\Component\Config\Definition\ConfigurationInterface;
+
+class Configuration implements ConfigurationInterface
+{
+    /**
+     * {@inheritdoc}
+     */
+    public function getConfigTreeBuilder(): TreeBuilder
+    {
+        $treeBuilder = new TreeBuilder();
+        $root = $treeBuilder->root('chameleon_system_view_renderer');
+        $root->addDefaultsIfNotSet();
+
+        $root
+            ->children()
+                ->scalarNode('css_dir')
+                    ->defaultValue(PATH_USER_CMS_PUBLIC.'/outbox/static/less')
+                ->end()
+            ->end();
+
+        return $treeBuilder;
+    }
+}
--- a/src/ViewRendererBundle/DependencyInjection/Configuration.php
+++ b/src/ViewRendererBundle/DependencyInjection/Configuration.php
@@ -20,6 +20,7 @@ class Configuration implements ConfigurationInterface
             ->children()
                 ->scalarNode('css_dir')
                     ->defaultValue(PATH_USER_CMS_PUBLIC.'/outbox/static/less')
+                    ->info('Directory path to where the less compiler caches work and the output files are stored.')
                 ->end()
             ->end();
 
--- a/src/ViewRendererBundle/objects/helper/TPkgViewRendererLessCompiler.php
+++ b/src/ViewRendererBundle/objects/helper/TPkgViewRendererLessCompiler.php
@@ -289,7 +289,7 @@ class TPkgViewRendererLessCompiler
     private function createDirectoryIfNeeded(string $dir): bool
     {
         if (false === \is_dir($dir)) {
-            if (!mkdir($dir, 0777, true) && !\is_dir($dir)) {
+            if (false === \mkdir($dir, 0777, true) && false === \is_dir($dir)) {
                 throw new ViewRenderException(sprintf('Cannot create directory %s', $dir));
             }
         }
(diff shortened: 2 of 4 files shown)


src/ViewRendererBundle/objects/helper/TPkgViewRendererLessCompiler.php
```diff
@@ -270,4 +283,15 @@ private function getMinifyParameter()
         return \ChameleonSystem\CoreBundle\ServiceLocator::getParameter(
             'chameleon_system_core.resources.enable_external_resource_collection_minify');
     }
+
+    private function checkAndCreateDir(string $dir): bool
```
umulmrum: Could we rename this method to something like `createOutputDirIfRequired`? The method's purpose isn't really to check anything but to assure that the directory exists.

Also I'd remove the return value and simply throw an exception instead of returning false. Then `getGeneratedCssForPortal` could simply call the method without having to check the result.

src/ViewRendererBundle/objects/helper/TPkgViewRendererLessCompiler.php
```diff
@@ -270,4 +283,15 @@ private function getMinifyParameter()
...
+    {
+        if (false === \is_dir($dir)) {
+            if (!mkdir($dir, 0777, true) && !\is_dir($dir)) {
```
umulmrum: Please use `false === ` instead of `!`.

src/ViewRendererBundle/objects/helper/TPkgViewRendererLessCompiler.php
```diff
@@ -227,11 +243,8 @@ public function writeCssFileForPortal($css, $portal)
     {
         $lessDir = $this->getLocalPathToCompiledLess();
 
-        if (false === \is_dir($lessDir)) {
-            if (!mkdir($lessDir, 0777, true) && !\is_dir($lessDir)) {
-                return false;
-            }
-        }
+        // NOTE this is probably already done in getGeneratedCssForPortal() for the cache of Less itself
```
umulmrum: I think this comment is a bit cryptic. What we should say is (correct me if I'm wrong) that this call ensures the directory is created even if a) `getGeneratedCssForPortal` wasn't called before (which reduces temporal coupling) or b) the cache directory is no subdir of the output dir for any reasons.

src/ViewRendererBundle/objects/helper/TPkgViewRendererLessCompiler.php
```diff
@@ -227,11 +243,8 @@ public function writeCssFileForPortal($css, $portal)
...
+        $this->checkAndCreateDir($lessDir);
```
umulmrum: If the method is changed to throw an exception, we should catch it here and return false (don't change the method signature in a fix release).

src/ViewRendererBundle/DependencyInjection/Configuration.php
```diff
@@ -0,0 +1,28 @@
...
+     */
+    public function getConfigTreeBuilder(): TreeBuilder
+    {
+        $treeBuilder = new TreeBuilder();
+        $root = $treeBuilder->root('chameleon_system_view_renderer');
+        $root->addDefaultsIfNotSet();
+
+        $root
+            ->children()
+                ->scalarNode('css_dir')
```
umulmrum: The name of this option could be more expressive. Perhaps even `css_output_dir` is enough?

Please also add a short description of the option using the `info` method.

src/ViewRendererBundle/objects/helper/TPkgViewRendererLessCompiler.php
```diff
@@ -19,14 +19,24 @@
 
 class TPkgViewRendererLessCompiler
 {
+    /**
+     * @var string
+     */
+    private $dir;
+
+    public function __construct(string $dir)
```
umulmrum: Could you rename this var to resemble the configuration option ($cssOutputDir if you follow my suggestion above)?

src/ViewRendererBundle/objects/helper/TPkgViewRendererLessCompiler.php
UlrichKu: ok, ok

src/ViewRendererBundle/objects/helper/TPkgViewRendererLessCompiler.php
UlrichKu: ok

src/ViewRendererBundle/objects/helper/TPkgViewRendererLessCompiler.php
UlrichKu: removed

src/ViewRendererBundle/objects/helper/TPkgViewRendererLessCompiler.php
UlrichKu: ok

src/ViewRendererBundle/objects/helper/TPkgViewRendererLessCompiler.php
UlrichKu: ok

src/ViewRendererBundle/DependencyInjection/Configuration.php
UlrichKu: info added

The name is unchanged: Its not only an output dir but also a working dir.
(So output, working, storage, ... is not correct)

umulmrum:
Thanks @UlrichKu

PR #57, opened 2018-10-09
118 maintenance mode command

UlrichKu:
| Q             | A
| ------------- | ---
| Branch        | 6.2.x
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Fixed issues  | chameleon-system/chameleon-system#118
| License       | MIT

### DIFF
--- a/src/CoreBundle/Bridge/Chameleon/Migration/Script/update-1539085816.inc.php
+++ b/src/CoreBundle/Bridge/Chameleon/Migration/Script/update-1539085816.inc.php
@@ -5,18 +5,19 @@
 </div>
 <?php
 
+const ERROR_MESSAGE_TYPE = '4';
+
 TCMSLogChange::AddBackEndMessage(
     'TABLEEDITOR_CMS_CONFIG_MAINTENANCE_ERROR',
     "Einstellung für 'Alle Webseiten abschalten' konnte nicht gespeichert werden: [{exceptionMessage}]",
-    TCMSLogChange::GetMessageTypeByName('Fehler'),
+    ERROR_MESSAGE_TYPE,
     '', 'de'
 );
 
 TCMSLogChange::AddBackEndMessage(
     'TABLEEDITOR_CMS_CONFIG_MAINTENANCE_ERROR',
     "Setting for 'Turn off all websites' could not be saved: [{exceptionMessage}]",
-    TCMSLogChange::GetMessageTypeByName('Fehler'),
+    ERROR_MESSAGE_TYPE,
     '', 'en'
 );
-
 
--- a/src/CoreBundle/Bridge/Chameleon/Migration/Script/update-1539085816.inc.php
+++ b/src/CoreBundle/Bridge/Chameleon/Migration/Script/update-1539085816.inc.php
@@ -5,18 +5,18 @@
 </div>
 <?php
 
-TCMSLogChange::AddFrontEndMessage(
+TCMSLogChange::AddBackEndMessage(
     'TABLEEDITOR_CMS_CONFIG_MAINTENANCE_ERROR',
     "Einstellung für 'Alle Webseiten abschalten' konnte nicht gespeichert werden: [{exceptionMessage}]",
     TCMSLogChange::GetMessageTypeByName('Fehler'),
-    '', '', 'Core', 'standard', 'de'
+    '', 'de'
 );
 
-TCMSLogChange::AddFrontEndMessage(
+TCMSLogChange::AddBackEndMessage(
     'TABLEEDITOR_CMS_CONFIG_MAINTENANCE_ERROR',
     "Setting for 'Turn off all websites' could not be saved: [{exceptionMessage}]",
     TCMSLogChange::GetMessageTypeByName('Fehler'),
-    '', '', 'Core', 'standard', 'en'
+    '', 'en'
 );
 
 
--- a/src/CoreBundle/Bridge/Chameleon/Migration/Script/update-1539085816.inc.php
+++ b/src/CoreBundle/Bridge/Chameleon/Migration/Script/update-1539085816.inc.php
@@ -1,0 +1,22 @@
+<h1>Build #1539085816</h1>
+<h2>Date: 2018-10-09</h2>
+<div class="changelog">
+    - #118: Add error message for maintenance mode problems
+</div>
+<?php
+
+TCMSLogChange::AddFrontEndMessage(
+    'TABLEEDITOR_CMS_CONFIG_MAINTENANCE_ERROR',
+    "Einstellung für 'Alle Webseiten abschalten' konnte nicht gespeichert werden: [{exceptionMessage}]",
+    TCMSLogChange::GetMessageTypeByName('Fehler'),
+    '', '', 'Core', 'standard', 'de'
+);
+
+TCMSLogChange::AddFrontEndMessage(
+    'TABLEEDITOR_CMS_CONFIG_MAINTENANCE_ERROR',
+    "Setting for 'Turn off all websites' could not be saved: [{exceptionMessage}]",
+    TCMSLogChange::GetMessageTypeByName('Fehler'),
+    '', '', 'Core', 'standard', 'en'
+);
+
+
--- a/src/CoreBundle/Bridge/Chameleon/Migration/Script/update-1539085816.inc.php
+++ b/src/CoreBundle/Bridge/Chameleon/Migration/Script/update-1539085816.inc.php
@@ -20,4 +20,3 @@ TCMSLogChange::AddBackEndMessage(
     ERROR_MESSAGE_TYPE,
     '', 'en'
 );
-
--- a/src/CoreBundle/Command/ActivateMaintenanceModeCommand.php
+++ b/src/CoreBundle/Command/ActivateMaintenanceModeCommand.php
@@ -13,16 +13,22 @@ namespace ChameleonSystem\CoreBundle\Command;
 
 use ChameleonSystem\CoreBundle\Exception\MaintenanceModeErrorException;
 use ChameleonSystem\CoreBundle\Service\MaintenanceModeServiceInterface;
-use ChameleonSystem\CoreBundle\ServiceLocator;
 use Symfony\Component\Console\Command\Command;
 use Symfony\Component\Console\Input\InputInterface;
 use Symfony\Component\Console\Output\OutputInterface;
 
 class ActivateMaintenanceModeCommand extends Command
 {
-    public function __construct($name = null)
+    /**
+     * @var MaintenanceModeServiceInterface
+     */
+    private $maintenanceModeService;
+
+    public function __construct(MaintenanceModeServiceInterface $maintenanceModeService)
     {
         parent::__construct('chameleon_system:maintenance_mode:activate');
+
+        $this->maintenanceModeService = $maintenanceModeService;
     }
 
     /**
@@ -44,13 +50,8 @@ EOF
      */
     protected function execute(InputInterface $input, OutputInterface $output)
     {
-        /**
-         * @var $maintenanceModeService MaintenanceModeServiceInterface
-         */
-        $maintenanceModeService = ServiceLocator::get('chameleon_system_core.maintenance_mode_service');
-
         try {
-            $maintenanceModeService->activate();
+            $this->maintenanceModeService->activate();
         } catch (MaintenanceModeErrorException $exception) {
             $output->writeln(sprintf('Maintenance mode could not be activated: %s', $exception->getMessage()));
 
--- a/src/CoreBundle/Command/ActivateMaintenanceModeCommand.php
+++ b/src/CoreBundle/Command/ActivateMaintenanceModeCommand.php
@@ -12,7 +12,7 @@
 namespace ChameleonSystem\CoreBundle\Command;
 
 use ChameleonSystem\CoreBundle\Exception\MaintenanceModeErrorException;
-use ChameleonSystem\CoreBundle\MaintenanceMode\MaintenanceModeServiceInterface;
+use ChameleonSystem\CoreBundle\Maintenance\MaintenanceMode\MaintenanceModeServiceInterface;
 use Symfony\Component\Console\Command\Command;
 use Symfony\Component\Console\Input\InputInterface;
 use Symfony\Component\Console\Output\OutputInterface;
--- a/src/CoreBundle/Command/ActivateMaintenanceModeCommand.php
+++ b/src/CoreBundle/Command/ActivateMaintenanceModeCommand.php
@@ -12,7 +12,7 @@
 namespace ChameleonSystem\CoreBundle\Command;
 
 use ChameleonSystem\CoreBundle\Exception\MaintenanceModeErrorException;
-use ChameleonSystem\CoreBundle\Service\MaintenanceModeServiceInterface;
+use ChameleonSystem\CoreBundle\MaintenanceMode\MaintenanceModeServiceInterface;
 use Symfony\Component\Console\Command\Command;
 use Symfony\Component\Console\Input\InputInterface;
 use Symfony\Component\Console\Output\OutputInterface;
--- a/src/CoreBundle/Command/ActivateMaintenanceModeCommand.php
+++ b/src/CoreBundle/Command/ActivateMaintenanceModeCommand.php
@@ -1,0 +1,62 @@
+<?php
+
+/*
+ * This file is part of the Chameleon System (https://www.chameleonsystem.com).
+ *
+ * (c) ESONO AG (https://www.esono.de)
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace ChameleonSystem\CoreBundle\Command;
+
+use ChameleonSystem\CoreBundle\Exception\MaintenanceModeErrorException;
+use ChameleonSystem\CoreBundle\Service\MaintenanceModeServiceInterface;
+use ChameleonSystem\CoreBundle\ServiceLocator;
+use Symfony\Component\Console\Command\Command;
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Output\OutputInterface;
+
+class ActivateMaintenanceModeCommand extends Command
+{
+    public function __construct($name = null)
+    {
+        parent::__construct('chameleon_system:maintenance_mode:activate');
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    protected function configure()
+    {
+        $this
+            ->setDescription('Activates the maintenance mode')
+            ->setHelp(<<<EOF
+The <info>%command.name%</info> command activates the maintenance mode.
+EOF
+            )
+        ;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    protected function execute(InputInterface $input, OutputInterface $output)
+    {
+        /**
+         * @var $maintenanceModeService MaintenanceModeServiceInterface
+         */
+        $maintenanceModeService = ServiceLocator::get('chameleon_system_core.maintenance_mode_service');
+
+        try {
+            $maintenanceModeService->activate();
+        } catch (MaintenanceModeErrorException $exception) {
+            $output->writeln(sprintf('Maintenance mode could not be activated: %s', $exception->getMessage()));
+
+            return 1;
+        }
+
+        return 0;
+    }
+}
--- a/src/CoreBundle/Command/DeactivateMaintenanceModeCommand.php
+++ b/src/CoreBundle/Command/DeactivateMaintenanceModeCommand.php
@@ -1,0 +1,62 @@
+<?php
+
+/*
+ * This file is part of the Chameleon System (https://www.chameleonsystem.com).
+ *
+ * (c) ESONO AG (https://www.esono.de)
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace ChameleonSystem\CoreBundle\Command;
+
+use ChameleonSystem\CoreBundle\Exception\MaintenanceModeErrorException;
+use ChameleonSystem\CoreBundle\Service\MaintenanceModeServiceInterface;
+use ChameleonSystem\CoreBundle\ServiceLocator;
+use Symfony\Component\Console\Command\Command;
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Output\OutputInterface;
+
+class DeactivateMaintenanceModeCommand extends Command
+{
+    public function __construct($name = null)
+    {
+        parent::__construct('chameleon_system:maintenance_mode:deactivate');
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    protected function configure()
+    {
+        $this
+            ->setDescription('Deactivates the maintenance mode')
+            ->setHelp(<<<EOF
+The <info>%command.name%</info> command deactivates the maintenance mode.
+EOF
+            )
+        ;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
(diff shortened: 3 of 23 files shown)


src/CoreBundle/Resources/config/defaults.inc.php
```diff
@@ -1192,12 +1192,9 @@
 }
 
 /**
- * Path to the file that marks if maintenance mode is active. The current directory's hash is added to
- * distinguish multiple Chameleon instances on a shared host.
- * We use the SCRIPT_FILENAME variable because we do NOT want to resolve symlinks. Otherwise it would not be possible
- * to activate maintenance mode and then perform a deployment using atomic deploys, because the maintenance marker
- * would then change its name.
+ * Path to the file that marks if maintenance mode is active. The default value with "cmsdata" should take care
```
umulmrum: I think the new comment doesn't address enough the requirements for the marker file, which are:
- Marker must survive deployment (if maintenance mode is active before deployment, it must still be active after deployment). I think this was the reason we used the temp dir before; we should check if this is still the case now and which requirements must be met by any deployment process.
- There must be one distinguishable marker per Chameleon installation on the same host (which is fulfilled if we store it in cmsdata, but it should be noted in the comment).
- In multi-node setups the same physical marker must be accessible to all nodes (which is only fulfilled for the default value if cmsdata is shared between nodes; the old description wasn't flawless, we need to make sure that our assumptions about file sharing and deployment are made explicit).

src/CoreBundle/Service/MaintenanceModeService.php
```diff
@@ -0,0 +1,92 @@
+<?php
+
+namespace ChameleonSystem\CoreBundle\Service;
```
umulmrum: I'm not happy about the Service namespace and would like to use it as little as possible. Could we use a sub-namespace of CoreBundle/Maintenance?

src/CoreBundle/Service/MaintenanceModeService.php
```diff
@@ -0,0 +1,92 @@
...
+
+use ChameleonSystem\CoreBundle\Exception\MaintenanceModeErrorException;
+use Doctrine\DBAL\Connection;
+use Doctrine\DBAL\DBALException;
+
+class MaintenanceModeService implements MaintenanceModeServiceInterface
```
umulmrum: Likewise I'm not that happy about our massive usage of the Service suffix; this doesn't express anything and I'd like to avoid using it. How about something like MaintenanceModeActivator?

src/CoreBundle/Service/MaintenanceModeService.php
```diff
@@ -0,0 +1,92 @@
...
+     * @var Connection
+     */
+    private $connection;
+
+    public function __construct(Connection $connection)
+    {
+        $this->connection = $connection;
+    }
+
+    public function isActivated(): bool
```
umulmrum: --> isActive

src/CoreBundle/Service/MaintenanceModeService.php
```diff
@@ -0,0 +1,92 @@
...
+    {
+        return file_exists(PATH_MAINTENANCE_MODE_MARKER);
```
umulmrum: Please use "true === ".

src/CoreBundle/Service/MaintenanceModeService.php
```diff
@@ -0,0 +1,92 @@
...
+    }
+
+    public function isActivatedInDb(): bool
```
umulmrum: How about something like "isRequestedActive"? I'd like to avoid the interface exposing a) which storage type is used and b) that there even are different storage types.

src/CoreBundle/Service/MaintenanceModeService.php
```diff
@@ -0,0 +1,92 @@
...
+    {
+        $col = $this->connection->fetchColumn("SELECT `shutdown_websites` FROM `cms_config`");
```
umulmrum: Why not use `TdbCmsConfig::GetInstance()->fieldShutdownWebsites`?

src/CoreBundle/Service/MaintenanceModeService.php
```diff
@@ -0,0 +1,92 @@
...
+            return false;
+        }
+
+        return '1' === $col['shutdown_websites'];
+    }
+
+    public function activate(): void
+    {
+        try {
+            $this->connection->executeUpdate("UPDATE `cms_config` SET `shutdown_websites` = '1'");
```
umulmrum: We need to invalidate the cache for cms_config afterwards.

src/CoreBundle/Service/MaintenanceModeService.php
```diff
@@ -0,0 +1,92 @@
...
+        } catch (DBALException $exception) {
+            throw new MaintenanceModeErrorException('Cannot save maintenance mode flag in database', 0, $exception);
```
umulmrum: Mini detail: Please end sentences with a dot.

src/CoreBundle/Service/MaintenanceModeService.php
```diff
@@ -0,0 +1,92 @@
...
+    }
+
+    public function deactivate(): void
+    {
+        $this->removeMarkerFile();
+
+        try {
+            $this->connection->executeUpdate("UPDATE `cms_config` SET `shutdown_websites` = '0'");
+        } catch (DBALException $exception) {
+            throw new MaintenanceModeErrorException('Cannot reset maintenance mode flag in database', 0, $exception);
```
umulmrum: Mini detail: Please end sentences with a dot (also in further exception messages in this class).

src/CoreBundle/Service/MaintenanceModeService.php
```diff
@@ -0,0 +1,92 @@
...
+
+        $this->addMarkerFile();
+    }
+
+    public function deactivate(): void
+    {
+        $this->removeMarkerFile();
+
+        try {
+            $this->connection->executeUpdate("UPDATE `cms_config` SET `shutdown_websites` = '0'");
```
umulmrum: We need to invalidate the cache for cms_config afterwards.

src/CoreBundle/Service/MaintenanceModeService.php
```diff
@@ -0,0 +1,92 @@
...
+        } catch (DBALException $exception) {
+            throw new MaintenanceModeErrorException('Cannot save maintenance mode flag in database', 0, $exception);
+        }
+
+        $this->addMarkerFile();
+    }
+
+    public function deactivate(): void
+    {
+        $this->removeMarkerFile();
```
umulmrum: Why remove the file before the DB change is done? Could you add a short comment?

src/CoreBundle/Service/MaintenanceModeService.php
```diff
@@ -0,0 +1,92 @@
...
+
+    /**
+     * @throws MaintenanceModeErrorException
+     */
+    protected function addMarkerFile(): void
+    {
+        $markerDir = \dirname(PATH_MAINTENANCE_MODE_MARKER);
+
+        if (false === \is_dir($markerDir)) {
+            if (!mkdir($markerDir, 0777, true) && !\is_dir($markerDir)) {
```
umulmrum: Please use "false === " instead of `!`.

src/CoreBundle/Service/MaintenanceModeService.php
```diff
@@ -0,0 +1,92 @@
...
+            $this->connection->executeUpdate("UPDATE `cms_config` SET `shutdown_websites` = '0'");
+        } catch (DBALException $exception) {
+            throw new MaintenanceModeErrorException('Cannot reset maintenance mode flag in database', 0, $exception);
+        }
+    }
+
+    /**
+     * @throws MaintenanceModeErrorException
+     */
+    protected function addMarkerFile(): void
```
umulmrum: The name implies that an item is added to a list. How about `createMarkerFile` instead?

src/CoreBundle/Service/MaintenanceModeService.php
```diff
@@ -0,0 +1,92 @@
...
+            throw new MaintenanceModeErrorException('Cannot save maintenance mode flag in file system');
+        }
+    }
+
+    /**
+     * @throws MaintenanceModeErrorException
+     */
+    protected function removeMarkerFile(): void
+    {
+        if (true === file_exists(PATH_MAINTENANCE_MODE_MARKER)) {
```
umulmrum: Early return instead of huge blocks would improve readability.

src/CoreBundle/Service/MaintenanceModeService.php
```diff
@@ -0,0 +1,92 @@
+<?php
+
```
umulmrum: Could you add the standard copyright file header?

src/CoreBundle/Service/MaintenanceModeServiceInterface.php
```diff
@@ -0,0 +1,35 @@
+<?php
+
```
umulmrum: Could you add the standard copyright file header?

src/CoreBundle/Service/MaintenanceModeServiceInterface.php
```diff
@@ -0,0 +1,35 @@
...
+namespace ChameleonSystem\CoreBundle\Service;
+
+use ChameleonSystem\CoreBundle\Exception\MaintenanceModeErrorException;
+
+interface MaintenanceModeServiceInterface
+{
+    /**
+     * Returns if the maintenance mode is activated (by file).
```
umulmrum: Client code should not need to care how maintenance mode works internally, so I'd remove the reference to the file.

src/CoreBundle/Service/MaintenanceModeServiceInterface.php
```diff
@@ -0,0 +1,35 @@
...
+     * Method should not query the database.
```
umulmrum: The database is an implementation detail and should not be relevant in the interface. What's your intention behind this suggestion?

src/CoreBundle/Service/MaintenanceModeServiceInterface.php
```diff
@@ -0,0 +1,35 @@
...
+     *
+     * @return bool
+     */
+    public function isActivated(): bool;
+
+    /**
+     * Returns if the maintenance mode is activated (in database).
```
umulmrum: As written above I would change the method name and then also the description.

src/CoreBundle/private/library/classes/TCMSTableEditor/TCMSTableEditorCMSConfig.class.php
```diff
@@ -65,16 +69,34 @@ public function UpdateTranslationFields()
...
-         * @var TdbCmsConfig $config
+         * @var $maintenanceModeService MaintenanceModeServiceInterface
          */
-        $config = $this->oTable;
-        if ($config->fieldShutdownWebsites) {
-            touch(PATH_MAINTENANCE_MODE_MARKER);
-        } else {
-            if (file_exists(PATH_MAINTENANCE_MODE_MARKER)) {
-                unlink(PATH_MAINTENANCE_MODE_MARKER);
+        $maintenanceModeService = \ChameleonSystem\CoreBundle\ServiceLocator::get('chameleon_system_core.maintenance_mode_service');
```
umulmrum: Could you outsource the service locator call into another method? This moves the ugly call to where we don't see it and we can add the typehint there to further simplify the "real" code.

src/CoreBundle/private/library/classes/TCMSTableEditor/TCMSTableEditorCMSConfig.class.php
```diff
@@ -65,16 +69,34 @@ public function UpdateTranslationFields()
...
+            if ($newConfig->fieldShutdownWebsites) {
+                $maintenanceModeService->activate();
+            } else {
+                $maintenanceModeService->deactivate();
             }
+        } catch (MaintenanceModeErrorException $exception) {
+            /**
+             * @var $flashMessageService FlashMessageServiceInterface
+             */
+            $flashMessageService = \ChameleonSystem\CoreBundle\ServiceLocator::get('chameleon_system_core.flash_messages');
```
umulmrum: See above (outsource).

src/CoreBundle/private/library/classes/TCMSTableEditor/TCMSTableEditorCMSConfig.class.php
```diff
@@ -65,16 +69,34 @@ public function UpdateTranslationFields()
...
-                unlink(PATH_MAINTENANCE_MODE_MARKER);
+        $maintenanceModeService = \ChameleonSystem\CoreBundle\ServiceLocator::get('chameleon_system_core.maintenance_mode_service');
+
+        /**
+         * @var TdbCmsConfig $newConfig
+         */
+        $newConfig = $this->oTable;
+
+        try {
+            if ($newConfig->fieldShutdownWebsites) {
```
umulmrum: "true === "

src/CoreBundle/Bridge/Chameleon/Migration/Script/update-1539085816.inc.php
```diff
@@ -0,0 +1,22 @@
+<h1>Build #1539085816</h1>
+<h2>Date: 2018-10-09</h2>
+<div class="changelog">
+    - #118: Add error message for maintenance mode problems
+</div>
+<?php
+
+TCMSLogChange::AddFrontEndMessage(
```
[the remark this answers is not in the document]
UlrichKu: ok

src/CoreBundle/Command/ActivateMaintenanceModeCommand.php
```diff
@@ -0,0 +1,62 @@
...
+
+    /**
+     * {@inheritdoc}
+     */
+    protected function execute(InputInterface $input, OutputInterface $output)
+    {
+        /**
+         * @var $maintenanceModeService MaintenanceModeServiceInterface
+         */
+        $maintenanceModeService = ServiceLocator::get('chameleon_system_core.maintenance_mode_service');
```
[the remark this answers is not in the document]
UlrichKu: ok

src/CoreBundle/Command/DeactivateMaintenanceModeCommand.php
```diff
@@ -0,0 +1,62 @@
...
+
+    /**
+     * {@inheritdoc}
+     */
+    protected function execute(InputInterface $input, OutputInterface $output)
+    {
+        /**
+         * @var $maintenanceModeService MaintenanceModeServiceInterface
+         */
+        $maintenanceModeService = ServiceLocator::get('chameleon_system_core.maintenance_mode_service');
```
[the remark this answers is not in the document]
UlrichKu: ok

src/CoreBundle/Service/MaintenanceModeService.php
UlrichKu: Manager...

src/CoreBundle/Exception/MaintenanceModeErrorException.php
```diff
@@ -0,0 +1,8 @@
+<?php
+
+namespace ChameleonSystem\CoreBundle\Exception;
```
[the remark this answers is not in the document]
UlrichKu: ok (3x)

src/CoreBundle/Resources/config/defaults.inc.php
UlrichKu: ok, comment extended

src/CoreBundle/Service/MaintenanceModeService.php
UlrichKu: ok; Service->MaintenanceMode (Maintenance already exists with a different meaning (?))

src/CoreBundle/Service/MaintenanceModeService.php
UlrichKu: Because it references a Tdb

src/CoreBundle/private/library/classes/TCMSTableEditor/TCMSTableEditorCMSConfig.class.php
UlrichKu: ok

src/CoreBundle/private/library/classes/TCMSTableEditor/TCMSTableEditorCMSConfig.class.php
UlrichKu: ok

src/CoreBundle/private/library/classes/TCMSTableEditor/TCMSTableEditorCMSConfig.class.php
UlrichKu: ok

src/CoreBundle/Service/MaintenanceModeService.php
UlrichKu: ok

src/CoreBundle/Service/MaintenanceModeService.php
UlrichKu: ok

src/CoreBundle/Service/MaintenanceModeService.php
UlrichKu: Changed

src/CoreBundle/Service/MaintenanceModeService.php
UlrichKu: After some discussion: "Service" fits the current interface contract best.

src/CoreBundle/Service/MaintenanceModeService.php
UlrichKu: It's the reverse order of the activation.

(The db info is more authoritative and thus changed first or last.)

src/CoreBundle/Service/MaintenanceModeServiceInterface.php
UlrichKu: Removed from interface for now: will be changed for #119

src/CoreBundle/Service/MaintenanceModeServiceInterface.php
UlrichKu: Removed from interface for now: will be changed for #119

src/CoreBundle/Service/MaintenanceModeServiceInterface.php
UlrichKu: Removed from interface for now: will be changed for #119

src/CoreBundle/Service/MaintenanceModeService.php
UlrichKu: see below

src/CoreBundle/Service/MaintenanceModeService.php
UlrichKu: see below

src/CoreBundle/Bridge/Chameleon/Migration/Script/update-1539085816.inc.php
```diff
@@ -0,0 +1,22 @@
...
+<h2>Date: 2018-10-09</h2>
+<div class="changelog">
+    - #118: Add error message for maintenance mode problems
+</div>
+<?php
+
+TCMSLogChange::AddBackEndMessage(
+    'TABLEEDITOR_CMS_CONFIG_MAINTENANCE_ERROR',
+    "Einstellung für 'Alle Webseiten abschalten' konnte nicht gespeichert werden: [{exceptionMessage}]",
+    TCMSLogChange::GetMessageTypeByName('Fehler'),
```
umulmrum: I'm afraid this won't work if updates are executed while using the backend with another language as German. This is a conceptual flaw in the GetMessageTypeByName method - I opened a new issue: chameleon-system/chameleon-system/issues/126

For now the best thing to do (*shudder*) is to hard-code the ID (as it is already done e.g. for the default value of AddFrontEndMessage).

src/CoreBundle/Bridge/Chameleon/Migration/Script/update-1539085816.inc.php
```diff
@@ -0,0 +1,22 @@
...
+    '', 'de'
+);
+
+TCMSLogChange::AddBackEndMessage(
+    'TABLEEDITOR_CMS_CONFIG_MAINTENANCE_ERROR',
+    "Setting for 'Turn off all websites' could not be saved: [{exceptionMessage}]",
+    TCMSLogChange::GetMessageTypeByName('Fehler'),
```
umulmrum: See above.

src/CoreBundle/MaintenanceMode/MaintenanceModeService.php
```diff
@@ -0,0 +1,108 @@
...
+/*
+ * This file is part of the Chameleon System (https://www.chameleonsystem.com).
+ *
+ * (c) ESONO AG (https://www.esono.de)
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace ChameleonSystem\CoreBundle\MaintenanceMode;
```
umulmrum: I intended to use the existing Maintenance namespace. Could you move again (maybe ChameleonSystem\CoreBundle\Maintenance\MaintenanceMode)?

src/CoreBundle/MaintenanceMode/MaintenanceModeService.php
```diff
@@ -0,0 +1,108 @@
...
+    }
+
+    public function activate(): void
+    {
+        try {
+            $this->connection->executeUpdate("UPDATE `cms_config` SET `shutdown_websites` = '1'");
+
+            $this->cache->callTrigger('cms_config');
+        } catch (DBALException $exception) {
+            throw new MaintenanceModeErrorException('Cannot save maintenance mode flag in database', 0, $exception);
```
umulmrum: I can't answer to outdated conversations, so I start anew: Anything looking like a sentence should end with a dot.

src/CoreBundle/MaintenanceMode/MaintenanceModeService.php
```diff
@@ -0,0 +1,108 @@
...
+     */
+    private $cache;
+
+    public function __construct(Connection $connection, CacheInterface $cache)
+    {
+        $this->connection = $connection;
+        $this->cache = $cache;
+    }
+
+    public function isActivated(): bool
```
umulmrum: Now this method and `isActivatedInDb` are public in the implementation but gone in the interface. What's the intention behind this (you mentioned a follow-up PR, but I can't see the big plan - PRs should be meaningful for themselves)? The implementation should not have additional public methods.

UlrichKu: Only alternative would be to remove these methods now and re-add them with the next to come PR.

Removing them from the interface was the idea to "not break the interface" with the next PR.

src/CoreBundle/Bridge/Chameleon/Migration/Script/update-1539085816.inc.php
UlrichKu: ok

src/CoreBundle/MaintenanceMode/MaintenanceModeService.php
UlrichKu: ok, changed

It's now stuttering: Maintenance\MaintenanceMode\MaintenanceModeService

src/CoreBundle/MaintenanceMode/MaintenanceModeService.php
umulmrum: We're running out of time for the current release, so I suggest to move this PR to 6.2.4, so that we can manage both PRs at once in case they depend on each other (as long as an interface is not published in a release we can change it anytime).

umulmrum:
Looks good to me, but PrettyCI isn't happy yet. Could you fix those issues (and ignore those that aren't caused by your changes)?

umulmrum:
Thanks @UlrichKu

PR #65, opened 2018-10-17
ref #140: support tls protocol in host definition.

rnuernberg:
| Q             | A
| ------------- | ---
| Branch        | 6.2.x
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes 
| Fixed issues  | chameleon-system/chameleon-system#140
| License       | MIT

### DIFF
--- a/src/CoreBundle/DependencyInjection/ChameleonSystemCoreExtension.php
+++ b/src/CoreBundle/DependencyInjection/ChameleonSystemCoreExtension.php
@@ -174,6 +174,7 @@ class ChameleonSystemCoreExtension extends Extension
             return;
         }
         $mailerDefinition = $container->getDefinition('chameleon_system_core.mailer');
+
         $mailerDefinition->addMethodCall('setSmtpHost', array($mailerConfig['host']));
         $mailerDefinition->addMethodCall('setSmtpUser', array($mailerConfig['user']));
         $mailerDefinition->addMethodCall('setSmtpPassword', array($mailerConfig['password']));
--- a/src/CoreBundle/DependencyInjection/Configuration.php
+++ b/src/CoreBundle/DependencyInjection/Configuration.php
@@ -103,6 +103,7 @@ class Configuration implements ConfigurationInterface
         $subTree->addDefaultsIfNotSet();
         $subTree->children()
             ->scalarNode('host')
+                ->info('The mailer host. To use a port other than 25, add the port to the host with a colon (Example: myhost:123). To use TLS, prefix the host with "tls://".')
             ->end()
             ->scalarNode('user')
             ->end()
--- a/src/CoreBundle/private/library/classes/TCMSMail.class.php
+++ b/src/CoreBundle/private/library/classes/TCMSMail.class.php
@@ -459,13 +459,16 @@ class TCMSMail extends PHPMailer
      */
     public function setSmtpHost($smtpHost)
     {
-        if (false === $colonPos = strpos($smtpHost, ':')) {
-            $this->smtpHost = $smtpHost;
-            $this->smtpPort = 25;
-        } else {
-            $this->smtpHost = substr($smtpHost, 0, $colonPos);
-            $this->smtpPort = (int) substr($smtpHost, $colonPos + 1);
+        $port = 25;
+        if (false !== ($portPosition = \mb_strrpos($smtpHost, ':'))) {
+            $portInHost = \mb_substr($smtpHost, $portPosition + 1);
+            if (true === \is_numeric($portInHost)) {
+                $port = (int) $portInHost;
+                $smtpHost = \mb_substr($smtpHost, 0, $portPosition);
+            }
         }
+        $this->smtpHost = $smtpHost;
+        $this->smtpPort = $port;
     }
 
     /**


src/CoreBundle/DependencyInjection/ChameleonSystemCoreExtension.php
```diff
@@ -174,7 +174,23 @@ private function addMailerConfig(array $mailerConfig, ContainerBuilder $containe
...
         }
         $mailerDefinition = $container->getDefinition('chameleon_system_core.mailer');
-        $mailerDefinition->addMethodCall('setSmtpHost', array($mailerConfig['host']));
+
+        $host = $mailerConfig['host'];
+        $port = (int) $mailerConfig['port'];
+        if (false !== ($portPosition = \mb_strrpos($host, ':'))) {
+            $portInHost = \mb_substr($host, $portPosition + 1);
+            if (true === \is_numeric($portPosition) && (25 === $port || $port === (int) $portInHost)) {
+                @trigger_error(
```
umulmrum: We can't add deprecations in fix versions, only in minor versions. But do we even need to deprecate this? Maybe it makes sense to support both ways of setting the port (then we need to define precedence if both values are set).

src/CoreBundle/DependencyInjection/ChameleonSystemCoreExtension.php
```diff
@@ -174,7 +174,23 @@ private function addMailerConfig(array $mailerConfig, ContainerBuilder $containe
             return;
         }
         $mailerDefinition = $container->getDefinition('chameleon_system_core.mailer');
-        $mailerDefinition->addMethodCall('setSmtpHost', array($mailerConfig['host']));
+
+        $host = $mailerConfig['host'];
+        $port = (int) $mailerConfig['port'];
+        if (false !== ($portPosition = \mb_strrpos($host, ':'))) {
+            $portInHost = \mb_substr($host, $portPosition + 1);
+            if (true === \is_numeric($portPosition) && (25 === $port || $port === (int) $portInHost)) {
```
umulmrum: The first comparison is superfluous, isn't it? We checked in line 180 if it is false, so it must be numeric here. Maybe you wanted to check `\is_numeric($portInHost)`?

src/CoreBundle/DependencyInjection/Configuration.php
```diff
@@ -103,6 +103,10 @@ private function getMailerConfig()
         $subTree->addDefaultsIfNotSet();
         $subTree->children()
             ->scalarNode('host')
+                ->info('The mailer host. If you need to use tls, you can prefix the host with "tls://".')
```
umulmrum:
```suggestion
                ->info('The mailer host. To use TLS, prefix the host with "tls://".')
```

src/CoreBundle/private/library/classes/TCMSMail.class.php
```diff
@@ -459,13 +459,15 @@ public function setMailPeerSecurity(HttpsContextInterface $mailPeerSecurity)
...
-            $this->smtpHost = substr($smtpHost, 0, $colonPos);
-            $this->smtpPort = (int) substr($smtpHost, $colonPos + 1);
-        }
+        $this->smtpHost = $smtpHost;
+    }
+
+    /**
+     * @param int $port
+     */
+    public function setSmtpPort($port)
```
umulmrum: Please add a type declaration instead of the docblock, and avoid return type declaration.

src/CoreBundle/DependencyInjection/ChameleonSystemCoreExtension.php
rnuernberg: I can understand not wanting to remove deprecated code in a minor or patch version. I see no reason to not to deprecate code in a patch version.

src/CoreBundle/DependencyInjection/ChameleonSystemCoreExtension.php
rnuernberg: true - that was a silly mistake

src/CoreBundle/private/library/classes/TCMSMail.class.php
rnuernberg: I am keeping with the style of the class here - change it anyway?

src/CoreBundle/DependencyInjection/ChameleonSystemCoreExtension.php
rnuernberg: I also don't see any advantage supporting both configuration options. I think this only adds potential confusion - so I would like to remove that option as soon as possible.

In a later version we should really move the TLS option into a separate configuration parameter - I don't think it should be part of the host.

Perhaps we want to add that option at this point so we can keep the current behavior but deprecate it and remove it in the next version.

bestform: Deprecations should only be introduced in minor versions. See https://semver.org/#how-should-i-handle-deprecating-functionality

umulmrum: What @bestform says :-) - if we like to deprecated the port within the host, we need a separate PR targetting master.

src/CoreBundle/private/library/classes/TCMSMail.class.php
umulmrum: I'd say we add typing anyway - if we keep the style, it is really hard to improve (we already deliberately broke the legacy conventions of capital letters for method names and the Hungarian notation).

rnuernberg: ok

src/CoreBundle/DependencyInjection/ChameleonSystemCoreExtension.php
rnuernberg: ok - I have moved the addition of the port parameter out of this PR. I think fixing the port handling in the host will be enough for the patch

umulmrum:
Sorry that I need to add quite a fundamental point after all your work, but I fear it's required. `TCMSMail::setSmtpHost` now has no longer the ability to receive host and port combined, which violates its (implicit) contract and is therefore a BC break. We can't rely on the class being exclusively instantiated through the extension class. Could you move the fix from the extension to setSmtpHost and remove setSmtpPort? Now that the fix stays quite close to the original implementation, we should not risk the break.

rnuernberg:
@umulmrum ok moved the code. I do think having a method called setSmtpHost also setting the port is quite unexpected - but I agree with your BC argument.

umulmrum:
Thanks @rnuernberg 

If I remember correctly the method was designed after the old DB-based configuration scheme where host and port were saved in the same field. Just open a new issue if you'd like to change it :-)

PR #68, opened 2018-10-22
ref #145: Consider a "field" for a multi table lookup

UlrichKu:
| Q             | A
| ------------- | ---
| Branch        | 6.2.x
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Fixed issues  | chameleon-system/chameleon-system#145
| License       | MIT

Make sure the cache key is different for this popup compared to the one for a normal "extended lookup".


umulmrum:
Thanks @UlrichKu

PR #75, opened 2018-10-25
Feature/77 symfony 3.4

umulmrum:
| Q             | A
| ------------- | ---
| Branch        | master 
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | yes
| Deprecations? | yes
| Tests pass?   | yes
| Fixed issues  | chameleon-system/chameleon-system#77
| License       | MIT

Some remarks:
- Some services are declared public now, see the note in UPGRADE-6.3.md. For mappers there is a separate ticket chameleon-system/chameleon-system#170, so they will still cause lots of deprecation warnings for now.
- Changes in ModuleResolver and CronJobFactory were also necessary because of these deprecations; we pass a Symfony service locator now instead of injecting the full container, so the contained services are neither read from the container nor should there be problems in Symfony 4.0+ (as the services are injected, they should not be removed by container optimization). These changes lead to BC breaks, but I don't expect any problems here.


src/CoreBundle/CronJob/CronJobFactory.php
```diff
@@ -23,24 +18,19 @@ public function __construct(ContainerInterface $container)
 
     /**
      * @param string[] $idList
+     *
+     * @deprecated since 6.3.0 - available cron jobs are injected by constructor now
```
UlrichKu: Wikipedia calls it "cron job" but the majority of the internet uses "cronjob" (?)

src/CoreBundle/DependencyInjection/Compiler/AddCronJobsPass.php
```diff
@@ -23,14 +25,16 @@ public function process(ContainerBuilder $container)
     {
         $cronJobFactoryDefinition = $container->getDefinition('chameleon_system_core.cronjob.cronjob_factory');
         $cronJobServiceIds = array_keys($container->findTaggedServiceIds('chameleon_system.cronjob'));
+        $services = [];
 
-        foreach ($cronJobServiceIds as $cronjobServiceId) {
-            $cronJobDefinition = $container->getDefinition($cronjobServiceId);
+        foreach ($cronJobServiceIds as $cronJobServiceId) {
```
UlrichKu: Wording maybe here as well.

src/CoreBundle/DependencyInjection/Compiler/ChameleonModulePass.php
```diff
@@ -28,14 +29,16 @@ public function process(ContainerBuilder $container)
...
             $moduleDefinition = $container->getDefinition($moduleServiceId);
-            if (true === $moduleDefinition->isShared() && ContainerInterface::SCOPE_PROTOTYPE !== $moduleDefinition->getScope()) {
+            if (true === $moduleDefinition->isShared()) {
                 throw new \LogicException('Chameleon modules must not be shared service instances. This module is shared: '.$moduleServiceId);
             }
+            $services[$moduleServiceId] = new Reference($moduleServiceId);
         }
 
-        $moduleServiceDefinition->addMethodCall('addModules', array($moduleServiceIds));
+        $moduleServiceDefinition->replaceArgument(0, ServiceLocatorTagPass::register($container, $services));
```
UlrichKu: Why is "register()" used as method argument here?

src/CoreBundle/Resources/config/cronjobs.xml
```diff
@@ -16,7 +16,6 @@
         </service>
 
         <service class="TCMSCronJob_ClearCompleteCache" id="chameleon_system_core.cronjob.clear_complete_cache_cronjob" shared="false">
-            <argument type="service" id="chameleon_system_cms_cache.cache"/>
```
UlrichKu: Isn't there / shouldn't there be  a different issue/PR for this?

src/CoreBundle/Resources/config/services.xml
```diff
@@ -133,16 +133,16 @@
...
-        <service id="session" synthetic="true" />
+        <service id="session" synthetic="true" public="true"/>
 
-        <service id="chameleon_system_core.chameleon_controller" synthetic="true" />
+        <service id="chameleon_system_core.chameleon_controller" synthetic="true" public="true"/>
 
-        <service id="chameleon_system_core.module_resolver" class="ChameleonSystem\CoreBundle\ModuleService\ModuleResolver">
-            <argument type="service" id="service_container" />
+        <service id="chameleon_system_core.module_resolver" class="ChameleonSystem\CoreBundle\ModuleService\ModuleResolver" public="true">
+            <argument /> <!-- Service locator will be set by ChameleonModulePass -->
```
UlrichKu: The Service locator?

src/CoreBundle/Resources/config/services.xml
```diff
@@ -753,21 +753,21 @@
             <argument type="service" id="chameleon_system_core.field.provider.class_from_table_field_request_level_cache.inner" />
         </service>
 
-        <service id="chameleon_system_core.cronjob.cronjob_factory" class="ChameleonSystem\CoreBundle\CronJob\CronJobFactory">
-            <argument type="service" id="service_container"/>
+        <service id="chameleon_system_core.cronjob.cronjob_factory" class="ChameleonSystem\CoreBundle\CronJob\CronJobFactory" public="true">
+            <argument /> <!-- Service locator will be set by AddCronJobsPass -->
```
UlrichKu: Service locator?

src/ViewRendererBundle/Bridge/Twig/Loader/TwigStringLoader.php
```diff
@@ -0,0 +1,49 @@
+<?php
+
+namespace ChameleonSystem\ViewRendererBundle\Bridge\Twig\Loader;
+
+/**
+ * Loads a template from a string.
+ *
+ * This loader should NEVER be used. It only exists for Twig internal purposes.
```
UlrichKu: Why is this class in our bundle?

src/ViewRendererBundle/Bridge/Twig/Loader/TwigStringLoader.php
```diff
@@ -0,0 +1,49 @@
...
+ *
+ * When using this loader with a cache mechanism, you should know that a new cache
+ * key is generated each time a template content "changes" (the cache key being the
+ * source code of the template). If you don't want to see your cache grows out of
+ * control, you need to take care of clearing the old cache file by yourself.
+ *
+ * @author Fabien Potencier <fabien@symfony.com>
+ *
+ *
+ * In the Chameleon System this class is only used in case a template could not be found. We can then display the
```
UlrichKu: How does this overwrite / relate to the normal implementation?

src/ViewRendererBundle/objects/helper/TPkgViewRendererSnippetResourceCollectorEndPoint.class.php
```diff
@@ -24,7 +26,7 @@ public function getLessResources($oPortal = null, $snippetPath = null)
         $aSnippetTree = $oTree->getConfigTree($oPortal, $snippetPath);
         $aTmpResources = $this->getConfigsFromTree($aSnippetTree);
         foreach ($aTmpResources as $sPathToConfig) {
-            $aConfig = Symfony\Component\Yaml\Yaml::parse($sPathToConfig);
+            $aConfig = Yaml::parse(\file_get_contents($sPathToConfig));
```
UlrichKu: Can there be any errors from \file_get_contents()?

src/CmsCacheBundle/Resources/config/services.xml
```diff
@@ -5,9 +5,9 @@
 
     <services>
 
-        <service id="chameleon_system_cms_cache.cache_database_connection" alias="database_connection" />
+        <service id="chameleon_system_cms_cache.cache_database_connection" alias="database_connection"  public="true"/>
 
-        <service id="chameleon_system_cms_cache.cache" class="esono\pkgCmsCache\Cache">
+        <service id="chameleon_system_cms_cache.cache" class="esono\pkgCmsCache\Cache" public="true">
```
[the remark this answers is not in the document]
umulmrum: It's because services will be private by default in Symfony 4 and it will no longer be possible to retrieve private services from the service container (which affects our ServiceLocator).

src/CoreBundle/CronJob/CronJobFactory.php
umulmrum: Both notations are common, so I decided for "cron job" in newer code which seems more "correct" to me.

src/CoreBundle/DependencyInjection/Compiler/ChameleonModulePass.php
umulmrum: See Symfony docs: https://symfony.com/doc/current/service_container/service_subscribers_locators.html#defining-a-service-locator (last example)

src/CoreBundle/Resources/config/cronjobs.xml
umulmrum: I couldn't call any SanityChecks (seems Symfony 3 is more strict here), so I fixed it directly. Shouldn't affect a separate bugfix PR.

src/CoreBundle/Resources/config/services.xml
umulmrum: Not the Chameleon ServiceLocator but a Symfony service locator which is a subset of the full service container (see https://symfony.com/doc/current/service_container/service_subscribers_locators.html).

src/CoreBundle/Resources/config/services.xml
umulmrum: See above.

src/ViewRendererBundle/Bridge/Twig/Loader/TwigStringLoader.php
umulmrum: It's because it was removed from Twig in version 2.0 and we still need it (could have re-implemented it, but didn't want to hide the fact that it is basically the same class as was used before and written by others).

src/ViewRendererBundle/Bridge/Twig/Loader/TwigStringLoader.php
umulmrum: The loaders are chained - this loader is only used if other loaders didn't find the template.

UlrichKu: I'd move this block to the top of the doc block.

And finish with something like: "Taken from xx and this is the original doc:"

src/CoreBundle/Resources/config/cronjobs.xml
UlrichKu: I mean mustn't this be fixed before in 6.2 (and master)?

src/CoreBundle/Resources/config/services.xml
UlrichKu: Then maybe call it "Symfony service locator"?

src/CoreBundle/Resources/config/cronjobs.xml
umulmrum: Yes, the PR you proposed will fix the problem in 6.2.x and we will merge the fix into master. With this separate fix (which is the same fix and cost close to no time) master works before the merge and the merge should recognize it's the same change and not raise a conflict. So it's better to have it in I think.

src/CoreBundle/Resources/config/services.xml
umulmrum: I think it's self-explanatory as soon as we're a bit more familiar with Symfony 3, as it makes neither sense nor is it possible to inject the Chameleon ServiceLocator (there is no service definition for it).

src/ViewRendererBundle/Bridge/Twig/Loader/TwigStringLoader.php
umulmrum: OK - done

PR #78, opened 2018-10-31
171 validate before save

UlrichKu:
| Q             | A
| ------------- | ---
| Branch        | master
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Fixed issues  | chameleon-system/chameleon-system#171
| License       | MIT

### DIFF
--- a/src/CoreBundle/private/library/classes/TCMSTableEditor/TCMSTableWriter.class.php
+++ b/src/CoreBundle/private/library/classes/TCMSTableEditor/TCMSTableWriter.class.php
@@ -9,9 +9,10 @@
  * file that was distributed with this source code.
  */
 
+use ChameleonSystem\AutoclassesBundle\CacheWarmer\AutoclassesCacheWarmer;
 use ChameleonSystem\CoreBundle\ServiceLocator;
 use ChameleonSystem\DatabaseMigration\DataModel\LogChangeDataModel;
-use ChameleonSystem\AutoclassesBundle\CacheWarmer\AutoclassesCacheWarmer;
+use Doctrine\DBAL\DBALException;
 
 /**
  * manages creation and deletion of tables.
@@ -30,6 +31,11 @@ class TCMSTableWriter extends TCMSTableEditor
 
     protected $aOldData = null;
 
+    /**
+     * @var string|null
+     */
+    private $oldTableComment;
+
     protected function LoadDataFromDatabase()
     {
         parent::LoadDataFromDatabase();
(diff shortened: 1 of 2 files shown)


src/CoreBundle/private/library/classes/TCMSTableEditor/TCMSTableWriter.class.php
```diff
@@ -127,19 +135,76 @@ protected function PostSaveHook(&$oFields, &$oPostTable)
...
+        $this->adaptTableComment($newTableName, $this->oldTableComment);
+    }
+
+    /**
+     * @param TCMSRecord $oPostTable
+     * @param string     $tableName
+     *
+     * @throws DBALException
+     */
+    private function adaptTableEngine(&$oPostTable, string $tableName): void
```
umulmrum: I'd like to change two things here:
- Move this code to changeTableEngine, so that there is only one method dealing with the table engine (the difference between `adaptTableEngine` and `changeTableEngine` can only be seen by reading the method bodies); I'm quite sure that we never want to have the table using another engine than is saved in the table config, so we should enforce writing both at the same time. This changes the method contract and we should add a small note in the upgrade guide.
- Rename `adaptTableName` and `adaptTableComment` to `changeXyz`. The `change` prefix is way more common.

src/CoreBundle/private/library/classes/TCMSTableEditor/TCMSTableWriter.class.php
```diff
@@ -127,19 +135,76 @@ protected function PostSaveHook(&$oFields, &$oPostTable)
...
+
+    /**
+     * @param string $oldTableName
+     * @param string $newTableName
+     *
+     * @throws DBALException
+     */
+    private function adaptTableName(string $oldTableName, string $newTableName): void
+    {
+        if ($oldTableName !== $newTableName) {
```
umulmrum: Please use an early return statement to improve readability.

src/CoreBundle/private/library/classes/TCMSTableEditor/TCMSTableWriter.class.php
```diff
@@ -127,19 +135,76 @@ protected function PostSaveHook(&$oFields, &$oPostTable)
...
+            $databaseConnection = $this->getDatabaseConnection();
+
+            $query = sprintf('ALTER TABLE %s RENAME %s', $databaseConnection->quoteIdentifier($oldTableName), $databaseConnection->quoteIdentifier($newTableName));
+            $databaseConnection->executeQuery($query);
+
+            $aQuery = array(new LogChangeDataModel($query));
+            TCMSLogChange::WriteTransaction($aQuery);
+
+            $this->_RenameRelatedTables($this->sOldTblName, $newTableName);
```
umulmrum: The old table name is both passed as argument and read from internal state - can we remove the argument?

src/CoreBundle/private/library/classes/TCMSTableEditor/TCMSTableWriter.class.php
```diff
@@ -127,19 +135,76 @@ protected function PostSaveHook(&$oFields, &$oPostTable)
...
+     * @param string $tableName
+     * @param string $oldTableComment
+     *
+     * @throws DBALException
+     */
+    private function adaptTableComment(string $tableName, string $oldTableComment): void
+    {
+        $newTableComment = $this->getTableComment($this->oTable->sqlData['translation'], $this->oTable->sqlData['notes']);
+
+        if ($oldTableComment !== $newTableComment) {
```
umulmrum: Please use an early return statement to improve readability.

src/CoreBundle/private/library/classes/TCMSTableEditor/TCMSTableWriter.class.php
```diff
@@ -127,19 +135,76 @@ protected function PostSaveHook(&$oFields, &$oPostTable)
...
             $_SESSION['_listObjCache'] = array();
         }
 
         $this->getAutoclassesCacheWarmer()->updateTableById($this->sId);
 
-        // check table engine
-        $realEngine = $this->getRealTableEngine($oPostTable->sqlData['name']);
+        $this->adaptTableEngine($oPostTable, $newTableName);
+        $this->adaptTableName($this->sOldTblName, $newTableName);
+        $this->adaptTableComment($newTableName, $this->oldTableComment);
```
umulmrum: I wouldn't pass an internal property to a private method, as the method can read it as well.

src/CoreBundle/private/library/classes/TCMSTableEditor/TCMSTableWriter.class.php
```diff
@@ -171,6 +236,8 @@ public function _RenameRelatedTables($sOldName, $sNewName)
      *
      * @param  $sOldTableName
      * @param  $sNewTableName
+     *
+     * @throws TPkgCmsException_Log
```
umulmrum: Technically it's correct to add these declarations, but I'm very unsure if we should do this. The exception class is deprecated and usages should be removed. The database methods that throw this exception are deprecated as well, so they should be replaced by modern database calls. If we add the exception declarations, we entrench this unsatisfying intermediate state, so I'm inclined to say that it's better to let these methods throw undeclared exceptions instead.

src/CoreBundle/private/library/classes/TCMSTableEditor/TCMSTableWriter.class.php
```diff
@@ -417,6 +490,8 @@ protected function getTableEngine($postData)
      * @param string $tableName
      *
      * @return string
+     *
+     * @throws DBALException
```
umulmrum: It's OK to add these declarations. Overall I'd like to ask if we should really let DBAL bleed into that much business code as it is now, but this is a separate and global topic and not really worsened by this change.

src/CoreBundle/private/library/classes/TCMSTableEditor/TCMSTableWriter.class.php
```diff
@@ -441,13 +516,15 @@ protected function getRealTableEngine($tableName)
...
-     * @throws ErrorException
-     * @throws TPkgCmsException_Log
+     * @throws DBALException
      */
     protected function changeTableEngine($tableName, $newEngine)
     {
-        $query = 'ALTER TABLE  `'.MySqlLegacySupport::getInstance()->real_escape_string($tableName)."` ENGINE = {$newEngine}";
-        MySqlLegacySupport::getInstance()->query($query);
+        $databaseConnection = $this->getDatabaseConnection();
+
```
umulmrum: Unnecessary newline.

src/CoreBundle/private/library/classes/TCMSTableEditor/TCMSTableWriter.class.php
UlrichKu: ok

src/CoreBundle/private/library/classes/TCMSTableEditor/TCMSTableWriter.class.php
umulmrum: No changes yet - forgot to push?

src/CoreBundle/private/library/classes/TCMSTableEditor/TCMSTableWriter.class.php
umulmrum: Also no change yet.

src/CoreBundle/private/library/classes/TCMSTableEditor/TCMSTableWriter.class.php
umulmrum: No change yet.

src/CoreBundle/private/library/classes/TCMSTableEditor/TCMSTableWriter.class.php
umulmrum: No changes yet.

UlrichKu: Removed (3 more)

umulmrum:
Thanks @UlrichKu

PR #79, opened 2018-11-05
173 validate before save field

UlrichKu:
| Q             | A
| ------------- | ---
| Branch        | master
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Fixed issues  | chameleon-system/chameleon-system#173
| License       | MIT

Uses now validation before save: most functionality is moved to "PostSaveHook()".
Table fields are correctly created and deleted if needed.

### DIFF
--- a/src/CoreBundle/private/library/classes/TCMSFields/TCMSField.class.php
+++ b/src/CoreBundle/private/library/classes/TCMSFields/TCMSField.class.php
@@ -846,12 +846,12 @@ class TCMSField implements TCMSFieldVisitableInterface
     /**
      * Renames existing related table.
      *
-     * @param string $sNewTableName
-     * @param bool   $returnDDL
+     * @param array $newFieldData
+     * @param bool  $returnDDL
      *
      * @return string|null
      */
-    public function RenameRelatedTables($sNewTableName, $returnDDL = false)
+    public function RenameRelatedTables($newFieldData, $returnDDL = false)
     {
         if ($returnDDL) {
             return '';
--- a/src/CoreBundle/private/library/classes/TCMSTableEditor/TCMSTableFieldWriter.class.php
+++ b/src/CoreBundle/private/library/classes/TCMSTableEditor/TCMSTableFieldWriter.class.php
@@ -10,6 +10,7 @@
  */
 
 use ChameleonSystem\AutoclassesBundle\CacheWarmer\AutoclassesCacheWarmer;
+use Doctrine\DBAL\DBALException;
 
 /**
  * manages creating, changing, and deleting fields from a table.
@@ -31,6 +32,16 @@ class TCMSTableFieldWriter extends TCMSTableEditor
     protected $_oParentRecord = null;
 
     protected $_sqlFieldName = null;
+
+    /**
+     * @var array|null
+     */
+    private $oldData;
+
+    /**
+     * @var bool
+     */
+    private $isChangeFromTable = false;
 
     /**
      * @return TCMSStdClass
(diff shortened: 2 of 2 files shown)


src/CoreBundle/private/library/classes/TCMSFields/TCMSField.class.php
```diff
@@ -846,12 +846,12 @@ public function AllowRenameRelatedTablesBeforeFieldSave($aNewFieldData, $aOldFie
...
      *
-     * @param string $sNewTableName
-     * @param bool   $returnDDL
+     * @param array $newFieldData
+     * @param bool  $returnDDL
      *
      * @return string|null
      */
-    public function RenameRelatedTables($sNewTableName, $returnDDL = false)
+    public function RenameRelatedTables($newFieldData, $returnDDL = false)
```
umulmrum: Although this is more a fix than a change, we should a short note in the UPGRADE guide.

src/CoreBundle/private/library/classes/TCMSTableEditor/TCMSTableFieldWriter.class.php
```diff
@@ -120,66 +136,118 @@ public function Save(&$postData, $bDataIsInSQLForm = false)
...
+     *
+     * @throws DBALException
+     */
+    protected function adaptFieldAndRelatedTables(array $postData): TCMSField
+    {
+        $oldTypeId = $this->oldData['cms_field_type_id'];
+        $newTypeId = $postData['cms_field_type_id'];
+        $newName = $this->oTable->sqlData['name'];
+
+        $bTypeChanged = $oldTypeId !== $newTypeId;
```
umulmrum: --> `typeChanged` or `doesTypeChange`

src/CoreBundle/private/library/classes/TCMSTableEditor/TCMSTableFieldWriter.class.php
```diff
@@ -120,66 +136,118 @@ public function Save(&$postData, $bDataIsInSQLForm = false)
...
+        $currentFieldTypeRow = $this->getFieldTypeDefinition($oldTypeId);
```
umulmrum: I'd use the prefix `old` instead of `current` to for consistency.

src/CoreBundle/private/library/classes/TCMSTableEditor/TCMSTableFieldWriter.class.php
```diff
@@ -120,66 +136,118 @@ public function Save(&$postData, $bDataIsInSQLForm = false)
...
+        if (true === $bTypeChanged) {
+            $oldField->ChangeFieldTypePreHook();
         }
-        if (!$bTargetTableChangeForMLTField) {
-            if ($this->_oField->AllowDeleteRelatedTablesBeforeFieldSave($postData, $currentFieldTypeRow, $newFieldTypeRow)) {
-                $this->_oField->DeleteRelatedTables();
-            } elseif ($this->_oField->AllowRenameRelatedTablesBeforeFieldSave($postData, $currentFieldTypeRow, $newFieldTypeRow)) {
-                $this->_oField->RenameRelatedTables($postData);
+
+        if (false === $this->isChangeFromTable) {
```
umulmrum: Can we invert the `if` clause? Negated conditions together with `else` are hard to read.

src/CoreBundle/private/library/classes/TCMSTableEditor/TCMSTableFieldWriter.class.php
```diff
@@ -120,66 +136,118 @@ public function Save(&$postData, $bDataIsInSQLForm = false)
...
-        $this->_oField->oDefinition = &$this->oTable;
-        if ($bTypeChanged) {
-            $this->_oField->ChangeFieldTypePostHook();
+        $oldField->RemoveFieldIndex();
+
+        $this->oTable->sqlData['cms_field_type_id'] = $newTypeId;
+        /**
+         * @var $newField TCMSField
+         */
+        $newField = &$this->oTable->GetFieldObject();
```
umulmrum: We can remove the `&` from assignments (also a few lines below).

src/CoreBundle/private/library/classes/TCMSTableEditor/TCMSTableFieldWriter.class.php
```diff
@@ -32,6 +33,21 @@ class TCMSTableFieldWriter extends TCMSTableEditor
...
+
+    /**
+     * @var string|null
+     */
+    private $oldTableName;
+
+    /**
+     * @var bool
+     */
+    private $isChangeFromTable = false;
```
umulmrum: I think this variable name could be improved. What does it really want to say?

src/CoreBundle/private/library/classes/TCMSTableEditor/TCMSTableFieldWriter.class.php
```diff
@@ -120,66 +136,118 @@ public function Save(&$postData, $bDataIsInSQLForm = false)
...
+                $newField->CreateFieldDefinition(); // is missing: old one was virtual
+            } else {
+                $oldField->DeleteFieldDefinition();
+            }
         }
-        $this->_oField->CreateFieldIndex();
 
-        $this->_oField->oDefinition->UpdateFieldTranslationKeys($aOriginalData);
-        $this->getAutoclassesCacheWarmer()->updateTableById($this->oTable->sqlData['cms_tbl_conf_id']);
+        // NOTE here was Save() before the reorganization - hence the method names
```
umulmrum:
```suggestion
        // NOTE Save() was called here before the code was reorganized - hence the method names
```

src/CoreBundle/private/library/classes/TCMSTableEditor/TCMSTableFieldWriter.class.php
```diff
@@ -120,66 +136,118 @@ public function Save(&$postData, $bDataIsInSQLForm = false)
...
+     * @throws DBALException
+     */
+    private function getFieldTypeDefinition(string $typeId)
+    {
+        $databaseConnection = $this->getDatabaseConnection();
+
+        $query = 'SELECT * FROM `cms_field_type` WHERE `id` = :typeId';
 
-        return $this->GetObjectShortInfo($postData);
+        return $databaseConnection->fetchAssoc($query, ['typeId' => $typeId]);
```
umulmrum: Can we return null instead of false in case no rows are found? This would be consistent with other code from recent years.

Also the callers of this method don't expect false or null yet.

UlrichKu: ok: null returned

However this is a rather basic system error (missing type). For example subclasses will work with this data.
Should this then be some sort of RuntimeException?

src/CoreBundle/private/library/classes/TCMSTableEditor/TCMSTableFieldWriter.class.php
umulmrum: Sounds good.

src/CoreBundle/private/library/classes/TCMSTableEditor/TCMSTableFieldWriter.class.php
umulmrum: This has not been addressed yet.

src/CoreBundle/private/library/classes/TCMSTableEditor/TCMSTableFieldWriter.class.php
umulmrum: OK

src/CoreBundle/private/library/classes/TCMSTableEditor/TCMSTableFieldWriter.class.php
UlrichKu: ok

src/CoreBundle/private/library/classes/TCMSTableEditor/TCMSTableFieldWriter.class.php
UlrichKu: ok

src/CoreBundle/private/library/classes/TCMSTableEditor/TCMSTableFieldWriter.class.php
UlrichKu: ok

src/CoreBundle/private/library/classes/TCMSTableEditor/TCMSTableFieldWriter.class.php
umulmrum: Maybe a `\LogicException`? Or what do you suggest (I'd like to avoid both `\Exception` and `\RuntimeException` as they don't express anything)?

umulmrum:
Thanks @UlrichKu

PR #81, opened 2018-11-07
WIP ref #97: Provide session as non-synthetic service

umulmrum:
| Q             | A
| ------------- | ---
| Branch        | master
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | yes (mosty in parts not covered by the BC promise)
| Deprecations? | no
| Tests pass?   | yes
| Fixed issues  | chameleon-system/chameleon-system#97
| License       | MIT

This PR basically solves the problem of the non-injectable session service. ChameleonSessionManager serves as a factory, allowing the same custom session functionality as before.

While the change itself is relatively simple, its implications are not. The session is now constructed before the request is available, so we can't inject the session into the request during construction. Unfortunately we already need the request during session start for some objects that get deserialized from the session, like the extranet user or the shop basket. This PR therefore also moves some part of the user deserialisation to a kernel.request listener.

The problem is that we get deep into some rabbit hole: I managed to decouple part of the shop basket deserialization in a similar way (no PR yet), but when switching the language during checkout, there is still a crash. I fear there will be other crashes hidden anywhere in the system, so I'd question if it is currently worth the time to figure out everything.

### DIFF
--- a/src/CoreBundle/Resources/config/services.xml
+++ b/src/CoreBundle/Resources/config/services.xml
@@ -113,9 +113,6 @@
 
         <service id="chameleon_system_core.service_initializer.request" class="ChameleonSystem\CoreBundle\Service\Initializer\RequestInitializer">
             <argument type="service" id="chameleon_system_core.response.response_variable_replacer" />
-            <call method="setSessionManager">
-                <argument type="service" id="chameleon_system_core.session.chameleon_session_manager" />
-            </call>
         </service>
 
         <service id="chameleon_system_core.InitializeRequestListener" class="ChameleonSystem\CoreBundle\EventListener\InitializeRequestListener" public="true">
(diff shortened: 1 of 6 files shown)


src/CoreBundle/Service/Initializer/RequestInitializer.php
```diff
@@ -71,14 +59,6 @@ protected function registerErrorHandler()
         }
     }
 
-    /**
-     * @param ChameleonSessionManagerInterface $sessionManager
-     */
```
UlrichKu: This must be deprecated?

src/CoreBundle/Session/ChameleonSessionManager.php
```diff
@@ -152,44 +129,11 @@ public function boot()
...
         $this->isSessionStarting = true;
-        $request->getSession()->start();
+        $session->start();
         $this->isSessionStarting = false;
-        $this->registerSessionInContainer($request->getSession());
-    }
-
-    /**
-     * @return Request
-     */
```
UlrichKu: Deprecated?

src/CoreBundle/Resources/config/services.xml
```diff
@@ -124,20 +121,19 @@
...
-            <argument type="service" id="chameleon_system_core.util.input_filter" />
         </service>
 
         <service id="chameleon_system_core.ChameleonSessionManager" alias="chameleon_system_core.session.chameleon_session_manager" public="true">
             <deprecated>The "%service_id%" service is deprecated since Chameleon 6.2.0 - use chameleon_system_core.session.chameleon_session_manager instead.</deprecated>
         </service>
 
-        <service id="session" synthetic="true" public="true"/>
+        <service id="session" class="TPkgCmsSession" public="true">
+            <factory service="chameleon_system_core.session.chameleon_session_manager" method="boot" />
```
UlrichKu: That would be rather surprising to have "boot" be a factory methode.

src/ExtranetBundle/objects/db/TDataExtranetUser.class.php
```diff
@@ -2450,12 +2450,8 @@ public function setFakedShippingAddressForUser($oShippingAddress)
         }
     }
 
-    /**
-     * {@inheritdoc}
-     */
```
UlrichKu: deprecation?

UlrichKu:
Seems to risky and doesn't consider all cases. Benefit is unclear.

PR #87, opened 2018-11-13
ref #117: Improve update notice

umulmrum:
| Q             | A
| ------------- | ---
| Branch        | 6.2.x
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed issues  | chameleon-system/chameleon-system#117
| License       | MIT

### DIFF
--- a/src/ViewRendererBundle/Bridge/Chameleon/Migration/Script/update-1539324183.inc.php
+++ b/src/ViewRendererBundle/Bridge/Chameleon/Migration/Script/update-1539324183.inc.php
@@ -27,5 +27,8 @@ $data = TCMSLogChange::createMigrationQueryData('cms_config', 'de')
 ;
 TCMSLogChange::update(__LINE__, $data);
 
-TCMSLogChange::addInfoMessage('Removing "chameleon/outbox/static/less/cached" from additionally to clean cache directories.
-Is now handled as part of the system - path can be configured.');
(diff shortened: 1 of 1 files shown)


src/ViewRendererBundle/Bridge/Chameleon/Migration/Script/update-1539324183.inc.php
```diff
@@ -27,5 +27,8 @@
 ;
 TCMSLogChange::update(__LINE__, $data);
 
-TCMSLogChange::addInfoMessage('Removing "chameleon/outbox/static/less/cached" from additionally to clean cache directories.
-Is now handled as part of the system - path can be configured.');
+TCMSLogChange::addInfoMessage('The path for generated CSS can now be configured using bundle config
```
UlrichKu: Why two messages? The second one is a (more or less) technical detail of the first.

umulmrum: Exactly because of this - the fact that the path is now configurable shouldn't be hidden as some detail within another message. I think both pieces of information are important and distinct enough to justify separate messages.

UlrichKu: I meant it the other way around...

PR #88, opened 2018-11-15
ref #119: Do not redirect when maintenance mode late check is true

UlrichKu:
| Q             | A
| ------------- | ---
| Branch        | 6.2.x
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Fixed issues  | chameleon-system/chameleon-system#119
| License       | MIT

Fixes the problem that on very slow shared file system between multiple nodes the maintenance mode marker file might really be missing for seconds.
During this time a redirect would not work and redirect again.


umulmrum:
Thanks @UlrichKu

PR #91, opened 2018-11-20
ref #206: Use the proper paths (absolute where needed)

UlrichKu:
| Q             | A
| ------------- | ---
| Branch        | 6.2.x
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed issues  | chameleon-system/chameleon-system#206
| License       | MIT

Any relevant position is now prefixed with PATH_WEB - which is also set correctly for a console environment.

### DIFF
--- a/src/ViewRendererBundle/objects/helper/TPkgViewRendererLessCompiler.php
+++ b/src/ViewRendererBundle/objects/helper/TPkgViewRendererLessCompiler.php
@@ -32,7 +32,7 @@ class TPkgViewRendererLessCompiler
     /**
      * Local path to less directory - this is where the chameleon_?.css files live.
      *
-     * @return string - relative path without trailing slash
+     * @return string - relative path (to PATH_WEB) without trailing slash
      */
     public function getLocalPathToCompiledLess()
     {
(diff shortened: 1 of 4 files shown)


UlrichKu:
Changed to absolute paths again.
The only "relative" user the routing now uses a new method for this.

src/ViewRendererBundle/objects/helper/TPkgViewRendererLessCompiler.php
```diff
@@ -81,8 +81,18 @@ protected function getCompiledCssFilename(TdbCmsPortal $portal = null)
         return 'chameleon'.$fileSuffix.'.css';
     }
 
+    /**
+     * @return string - the relative (to PATH_WEB) path pattern to the chameleon css file, without leading slash
```
umulmrum:
```suggestion
     * @return string - the path pattern to the generated CSS file, relative to PATH_WEB, without leading slash
```

src/ViewRendererBundle/objects/helper/TPkgViewRendererLessCompiler.php
```diff
@@ -81,8 +81,18 @@ protected function getCompiledCssFilename(TdbCmsPortal $portal = null)
...
+     */
+    public function getCssRoutingPattern(): string
+    {
+        return $this->cssDir.'/'.'chameleon_{portalId}.css';
+    }
+
     /**
      * @return string - the file part for route generation; without a leading slash
+     *
+     * @deprecated since 6.2.4 - use getCssRoutingPattern() which includes the relative path
```
umulmrum: Deprecation is only allowed in minor versions, so we need to move this to 6.3.0.

src/ViewRendererBundle/objects/helper/TPkgViewRendererLessCompiler.php
```diff
@@ -81,8 +81,18 @@ protected function getCompiledCssFilename(TdbCmsPortal $portal = null)
...
+     */
+    public function getCssRoutingPattern(): string
+    {
+        return $this->cssDir.'/'.'chameleon_{portalId}.css';
+    }
+
     /**
      * @return string - the file part for route generation; without a leading slash
+     *
+     * will deprecate with 6.3.0 - use getCssRoutingPattern() which includes the relative path
```
umulmrum:
```suggestion
     * will be deprecated in 6.3.0 - use getCssRoutingPattern() which includes the relative path
```

src/ViewRendererBundle/objects/helper/TPkgViewRendererLessCompiler.php
umulmrum: The message. Could you file a PR for 6.3 that adds the @deprecated annotation and a notice in the migration guide?

umulmrum:
Thanks @UlrichKu

PR #92, opened 2018-11-22
WIP Feature/94 symfony flex

umulmrum:
| Q             | A
| ------------- | ---
| Branch        | master
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | yes
| Deprecations? | yes
| Tests pass?   | yes
| Fixed issues  | chameleon-system/chameleon-system#94
| License       | MIT

A WIP PR for introducing Symfony Flex. Another PR was filed for chameleon-system/chameleon-system (https://github.com/chameleon-system/chameleon-system/pull/209).

We should first review the changes to see if they are OK, i.e. are the BC breaks tolerable for 6.3 or if we should wait for 7.0. 
Then there are further tasks:
- Write deprecation notes to the UPGRADING file.
- Write migration notes to the UPGRADING file. Some of the required Composer and file changes should be handled by a migration script to simplify the migration.
- Write notes to the CHANGELOG file.

Mostly this is a "simple" switch to Flex, but also I use the Doctrine bundle so that we can get rid of some services in CoreBundle's services.xml, be closer to the Symfony standard, avoid some environment variables, and gain the Doctrine debug toolbar profiler. The old Chameleon database profiler is now disabled by default, but otherwise left untouched, as it might provide other information than the Doctrine profiler.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -28,6 +28,7 @@
         "symfony/polyfill-apcu": "^1.0",
         "symfony/console": "~3.4.0",
         "symfony/dependency-injection": "~3.4.0",
+        "symfony/expression-language": "~3.4.0",
         "symfony/framework-bundle": "~3.4.0",
         "symfony/routing": "~3.4.0",
         "symfony/security-csrf": "~3.4.0",
--- a/composer.json
+++ b/composer.json
@@ -26,7 +26,13 @@
         "sensio/distribution-bundle": "^5.0.19",
         "symfony/monolog-bundle": "^3.1.0",
         "symfony/polyfill-apcu": "^1.0",
-        "symfony/symfony": "~3.4.0",
+        "symfony/console": "~3.4.0",
+        "symfony/dependency-injection": "~3.4.0",
+        "symfony/framework-bundle": "~3.4.0",
+        "symfony/routing": "~3.4.0",
+        "symfony/templating": "~3.4.0",
+        "symfony/validator": "~3.4.0",
+        "symfony/yaml": "~3.4.0",
         "symfony-cmf/routing-bundle": "~2.1.0",
         "tedivm/jshrink": "~1.0",
         "twig/twig": "^2.0"
@@ -34,7 +40,10 @@
     "require-dev": {
         "chameleon-system/code-style-config": "dev-master@dev",
         "chameleon-system/sanitycheck": "dev-master",
-        "phpunit/phpunit": "~7.0"
+        "phpunit/phpunit": "~7.0",
+        "symfony/debug-bundle": "~3.4.0",
+        "symfony/stopwatch": "~3.4.0",
+        "symfony/web-profiler-bundle": "~3.4.0"
     },
     "autoload": {
         "classmap": [
--- a/composer.json
+++ b/composer.json
@@ -15,6 +15,7 @@
         "chameleon-system/sanitycheck-bundle": "dev-master",
         "ckeditor/ckeditor": "4.6.2",
         "doctrine/dbal": "~2.5",
+        "doctrine/doctrine-bundle": "^1.6.10",
         "ext-curl": "*",
         "ext-tidy": "*",
         "ext-pdo_mysql": "*",
@@ -23,10 +24,10 @@
         "natxet/CssMin": "~3.0.0",
         "php": "^7.1",
         "phpmailer/phpmailer": "~6.0",
+        "symfony/config": "~3.4.0",
         "symfony/console": "~3.4.0",
         "symfony/dependency-injection": "~3.4.0",
         "symfony/expression-language": "~3.4.0",
-        "symfony/flex": "^1.1.0",
         "symfony/framework-bundle": "~3.4.0",
         "symfony/monolog-bundle": "^3.1.0",
         "symfony/polyfill-apcu": "^1.0",
@@ -44,8 +45,7 @@
     "require-dev": {
         "chameleon-system/code-style-config": "dev-master@dev",
         "chameleon-system/sanitycheck": "dev-master",
-        "phpunit/phpunit": "~7.0",
-        "symfony/dotenv": "~3.4.0"
+        "phpunit/phpunit": "~7.0"
     },
     "autoload": {
         "classmap": [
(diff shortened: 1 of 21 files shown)


composer.json
```diff
@@ -23,10 +24,20 @@
...
-        "sensio/distribution-bundle": "^5.0.19",
+        "symfony/config": "~3.4.0",
+        "symfony/console": "~3.4.0",
+        "symfony/dependency-injection": "~3.4.0",
+        "symfony/expression-language": "~3.4.0",
+        "symfony/framework-bundle": "~3.4.0",
         "symfony/monolog-bundle": "^3.1.0",
         "symfony/polyfill-apcu": "^1.0",
-        "symfony/symfony": "~3.4.0",
+        "symfony/routing": "~3.4.0",
```
UlrichKu: These 10 do no have a unifying dependency (eg symfony/symfony)?

src/CmsCacheBundle/Resources/config/services.xml
```diff
@@ -11,7 +11,7 @@
             <argument type="service" id="request_stack" />
             <argument type="service" id="chameleon_system_cms_cache.cache_database_connection" />
             <argument type="service" id="chameleon_system_cms_cache.storage.null" />
-            <argument>%secret%</argument>
+            <argument>%env(APP_SECRET)%</argument>
```
UlrichKu: Who sets this / where?

src/CoreBundle/Resources/config/checks.xml
```diff
@@ -46,26 +46,26 @@
...
-                <argument>../web/chameleon/outbox/static/js</argument>
-                <argument>../web/chameleon/outbox/static/less</argument>
+                <argument>var/cmsdata</argument>
+                <argument>var/cache</argument>
+                <argument>var/cache/autoclasses</argument>
+                <argument>var/log</argument>
+                <argument>public/chameleon/mediapool</argument>
+                <argument>public/chameleon/outbox/static/css</argument>
+                <argument>public/chameleon/outbox/static/js</argument>
+                <argument>public/chameleon/outbox/static/less</argument>
```
UlrichKu: This can/will be generated automatically if needed - and at a configurable location.

src/CoreBundle/Resources/config/project-config.yml
```diff
@@ -57,28 +57,20 @@ parameters:
 
   chameleon_system_external_tracker_google_analytics.enable_campaign_tracking: false # if true, the product exports will add their google campaign parameters (if they support them) to the detail links of the products
 
-  database_port: 3306
+  database_host: "" # @deprecated since 6.3.0 - no longer used
```
UlrichKu: It's an environment variable now? Maybe link/document here.

src/CoreBundle/Resources/config/project-config_test.yml
```diff
@@ -1,22 +1,23 @@
 imports:
-   - { resource: project-config.yml }
+     - { resource: project-config.yml }
```
UlrichKu: Some files are renamed to "yaml" in the other PR (https://github.com/chameleon-system/chameleon-system/pull/209)?

src/CoreBundle/bin/console
```diff
@@ -1,23 +1,36 @@
...
 
 define('PATH_WEB', $_SERVER['DOCUMENT_ROOT']);
 require_once __DIR__.'/../Resources/config/const.inc.php';
 
-if (file_exists(PATH_PROJECT_BASE.'/app/autoload.php')) {
-    include_once PATH_PROJECT_BASE.'/app/autoload.php';
-} else {
-    include_once __DIR__.'/../../../../../autoload.php';
+$input = new ArgvInput();
+if (null !== $_ENV['APP_ENV'] = $input->getParameterOption(['--env', '-e'], null, true)) {
```
UlrichKu: use brackets?

null !== (X = y)

or maybe even better two lines

src/CoreBundle/FrontController/index.php
```diff
@@ -29,21 +33,20 @@
 
 require_once PATH_PROJECT_CONFIG.'/config.inc.php';
 
+define('_DEVELOPMENT_MODE', $_SERVER['APP_DEBUG']);
+
+if (true === $_SERVER['APP_DEBUG']) {
+    umask(0000);
```
bestform: Why on earth?

src/CoreBundle/bin/console
```diff
@@ -1,23 +1,36 @@
...
+    putenv('APP_ENV='.$_ENV['APP_ENV']);
+    // force loading .env files when --env is defined
+    $_SERVER['APP_ENV'] = null;
+}
```
bestform: This whole block is a little confusing. First we set a value inside $_ENV by hand, then we transfer it to the outer environment - where it should have come from in the first place - and after that, we set a value in $_SERVER with the same name to `null`. I am sure this produces certain side effects we want at this point, but this code is not obvious about that at all. If we need to do it this way, a small doc paragraph explaining the thoughts behind it would be neat.

src/CoreBundle/bin/console
```diff
@@ -1,23 +1,36 @@
...
+
+if ($input->hasParameterOption('--no-debug', true)) {
+    putenv('APP_DEBUG='.$_SERVER['APP_DEBUG'] = $_ENV['APP_DEBUG'] = '0');
```
bestform: I would avoid this assignment + concatenation shortcut and make it more explicit what we are doing here.

src/CoreBundle/bin/console
```diff
@@ -27,20 +40,14 @@ spl_autoload_register('ChameleonSystem\AutoclassesBundle\Loader\AutoClassLoader:
 
 TGlobal::setMode(TGlobal::MODE_BACKEND);
 
-require_once PATH_PROJECT_BASE.'/app/AppKernel.php';
-
-use Symfony\Bundle\FrameworkBundle\Console\Application;
-use Symfony\Component\Console\Input\ArgvInput;
+if (true === $_SERVER['APP_DEBUG']) {
+    umask(0000);
```
bestform: Am I missing something here? Why are we doing this?

umulmrum:
@bestform @UlrichKu My intention for the first step was this:

"We should first review the changes to see if they are OK, i.e. are the BC breaks tolerable for 6.3 or if we should wait for 7.0."

We should clarify details afterwards.

PR #93, opened 2018-11-22
Fix/Unmanaged List Table Handling

ApricumMedia:
| Q             | A
| ------------- | ---
| Branch        | master
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Fixed issues  | no
| License       | MIT

Provides supplementary changes provided for overlay- and inline editing introduced by commits up to `77d134fe1f9dfdeb96353d5dcd621fe14588f667` of the `chameleon-system/inline-editing-bundle`, currently in review.

Add `TCMSListManager` support-indicative function `usesManagedTables(): bool`. Returned value is forwarded to all created tables (`TGroupTable`, `TFullGroupTable`, ...) as `isManagedTable` and passed along to created table fields with `tableIsManaged`.

The event emitted by table fields includes a reference to the field, allowing determining if it is owned by a managed table. All transformations performed by the inline editing bundle are *no longer applied* when encountering a field that does not belong to a table with the `*isManaged` state.

### DIFF
--- a/src/CoreBundle/private/library/classes/TCMSListManager/TCMSListManagerCmsTableConf.class.php
+++ b/src/CoreBundle/private/library/classes/TCMSListManager/TCMSListManagerCmsTableConf.class.php
@@ -16,9 +16,10 @@ class TCMSListManagerCmsTableConf extends TCMSListManagerFullGroupTable
     }
 
     /**
-     * {@inheritDoc}
+     * {@inheritdoc}
      */
-    protected function usesManagedTables(): bool {
+    protected function usesManagedTables(): bool
+    {
         return false;
     }
 }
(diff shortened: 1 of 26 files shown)


src/CoreBundle/private/library/classes/TCMSListManager/TCMSListManagerCmsTableConf.class.php
```diff
@@ -14,4 +14,11 @@ class TCMSListManagerCmsTableConf extends TCMSListManagerFullGroupTable
     public function _AddFunctionColumn()
     {
     }
+
+    /**
+     * {@inheritDoc}
+     */
+    protected function usesManagedTables(): bool {
```
pixeljunkie: Please run cs-fixer on your changes. The code format is not right here.

PR #95, opened 2018-11-23
ref #211: add missing view path to error message

pSpeck:
| Q             | A
| ------------- | ---
| Branch        | 6.2.x
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed issues  | chameleon-system/chameleon-system#211 
| License       | MIT

### DIFF
--- a/src/CoreBundle/private/library/classes/TViewParser/TViewPathManager.class.php
+++ b/src/CoreBundle/private/library/classes/TViewParser/TViewPathManager.class.php
@@ -102,16 +102,19 @@ class TViewPathManager implements IViewPathManager
         /*
          * @deprecated only used by pkgBlog
          */
+        $sPath = TPkgViewRendererSnippetDirectory::PATH_MODULES;
         if (!TGlobal::IsCMSMode()) {
             // overwrite path with theme path if we find a portal for the current page and a theme is set
             $activePortal = $this->portalDomainService->getActivePortal();
             if (null !== $activePortal) {
-                $sThemePath = $activePortal->GetThemeBaseModuleViewsPath();
+                $sThemePath = $activePortal->GetThemeObjectViewsPath();
                 if (!empty($sThemePath)) {
-                    $sTemplatePath = $sThemePath.'/'.$sModuleName.'/'.$sViewName.'.view.php';
+                    $sPath = $sThemePath;
                 }
             }
         }
+
+        $sTemplatePath = $sPath.'/'.$sViewFileName;
 
         return $sTemplatePath;
     }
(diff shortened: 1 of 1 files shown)


src/CoreBundle/private/library/classes/TViewParser/TViewPathManager.class.php
```diff
@@ -188,15 +191,17 @@ public function getObjectPackageViewPath($sViewName, $sSubType = '', $sType = 'C
             return $sTemplatePathFromTheme;
         }
 
-        if ('Core' != $sType && false == TGlobal::IsCMSMode()) {
+        $sPath = TPkgViewRendererSnippetDirectory::PATH_OBJECTVIEWS;
```
umulmrum: We shouldn't introduce Hungarian notation for new variables (please rename the variable both here and above).

src/CoreBundle/private/library/classes/TViewParser/TViewPathManager.class.php
```diff
@@ -102,17 +102,20 @@ public function getWebModuleViewPath($sViewName, $sModuleName, $sType = 'Custome
         /*
          * @deprecated only used by pkgBlog
          */
+        $sPath = TPkgViewRendererSnippetDirectory::PATH_MODULES;
```
umulmrum: The comment states that this code fragment is deprecated. Did you change it just for completeness' sake (which is OK) or is the code actually used (so that we should think about the deprecation message)?

pSpeck: The code is actually use if the template was not found.

src/CoreBundle/private/library/classes/TViewParser/TViewPathManager.class.php
pSpeck: change here and above

umulmrum:
I have to admit I didn't completely understand the change on the first review. The issue is about displaying better error messages, but this PR changes template resolution logic - could you explain the changes? What are the implications?

src/CoreBundle/private/library/classes/TViewParser/TViewPathManager.class.php
```diff
@@ -102,6 +102,7 @@ public function getWebModuleViewPath($sViewName, $sModuleName, $sType = 'Custome
         /*
          * @deprecated only used by pkgBlog
          */
+        $templateModulePath = TPkgViewRendererSnippetDirectory::PATH_MODULES;
```
umulmrum: This line is no longer needed.

pSpeck: removed

PR #99, opened 2018-11-28
ref #30: Fix width field definition for cms_media

UlrichKu:
| Q             | A
| ------------- | ---
| Branch        | 6.2.x
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Fixed issues  | chameleon-system/chameleon-system#30
| License       | MIT

Note there is a faulty field definition for "width" (with TCMSFieldMediaProperties).


UlrichKu:
Relevant location - which prohibits saving the width field: 
(central)
vendor/chameleon-system/chameleon-base/src/CoreBundle/private/library/classes/TCMSTableEditor/TCMSTableEditorEndPoint.class.php:1664

umulmrum:
I'm a bit confused in multiple ways :-)

First, the issue stated that both width and height don't work as expected. Maybe @proligde can assist in reproducing the issue?

Second, I'm afraid I didn't understand your explanation above. Why does TCMSTableEditorEndPoint prohibit saving the width, and why wasn't this the case in the past (as the manual save of the height and the file size was in place before, but not the save of the width)?

Third, why do we need to allow editing by all? This isn't the case for the other fields above the change, and this shouldn't be required in the backend.

UlrichKu:
> First, the issue stated that both width and height don't work as expected. Maybe @proligde can assist in reproducing the issue?

The described error/test data used the same height for the second image. That worked (and thus was visibly unchanged as well).

> 
> Second, I'm afraid I didn't understand your explanation above. Why does TCMSTableEditorEndPoint prohibit saving the width, and why wasn't this the case in the past (as the manual save of the height and the file size was in place before, but not the save of the width)?

With https://github.com/chameleon-system/chameleon-system/issues/5 this problem was introduced.
This returns "hidden" for the "width" field display mode. Consequently it cannot be saved (by the backend). Later tries to save the field directly would have the same result.

> 
> Third, why do we need to allow editing by all? This isn't the case for the other fields above the change, and this shouldn't be required in the backend.

That is a hack as "edit for all" is checked during writing before the the display mode check and allows saving of the width field.

src/CoreBundle/private/library/classes/TCMSTableEditor/TCMSTableEditorMedia.class.php
```diff
@@ -410,6 +410,14 @@ protected function PostSaveHook(&$oFields, &$oPostTable)
             if (array_key_exists('height', $oPostTable->sqlData)) {
                 $this->SaveField('height', $oPostTable->sqlData['height']);
             }
+            if (array_key_exists('width', $oPostTable->sqlData)) {
+                // See https://github.com/chameleon-system/chameleon-system/issues/30
```
umulmrum:
```suggestion
                /*
                 * The width field is hidden and will therefore not be saved by default. We save it
                 * explicitly and also need to allow editing by all to bypass restrictions of the hidden state.
                 */
```

umulmrum:
Thanks @UlrichKu

PR #101, opened 2018-11-28
ref #217: add media manager iframe page definition and call this header less…

pixeljunkie:
… view for image selection mode

| Q             | A
| ------------- | ---
| Branch        | 6.2.x
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed issues  | chameleon-system/chameleon-system#217
| License       | MIT


src/MediaManagerBundle/Bridge/Chameleon/BackendModule/MediaManagerBackendModule.php
```diff
@@ -48,6 +48,8 @@ class MediaManagerBackendModule extends MTPkgViewRendererAbstractModuleMapper
 {
     const PAGEDEF_NAME = 'mediaManager';
 
+    const PAGEDEF_IFRAME_NAME = 'mediaManagerIframe';
```
umulmrum: "iframe" is a technical term and is also not really what we do here. Is a term referring to "pick image" better here? This would use a domain terminology and also avoid to make hard assumptions about the used technology.

pixeljunkie: renamed the const and pagedef file to pickImage

umulmrum:
Thanks @pixeljunkie

PR #104, opened 2018-11-29
ref #108: Do not set fixed heights for media manager or popups

UlrichKu:
| Q             | A
| ------------- | ---
| Branch        | 6.2.x
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Fixed issues  | chameleon-system/chameleon-system#108
| License       | MIT

Without fixed heights this looks good/ok:
For the default of 204 images per page the media manager page is rather high. But this is handled by the browser scrollbar.
Other views with fewer images are then only as high as needed.

This also looks ok when selecting an image for some image field.


src/MediaManagerBundle/Resources/public/js/mediaManager.js
```diff
@@ -1036,7 +1034,9 @@
                 newHeight = this.element.height();
             }
 
-            layover.css({'height': newHeight, 'width': '100%'});
+            console.log("Setting height "+newHeight);
```
pixeljunkie: Please remove debug code.

UlrichKu: ok (full block gone)

umulmrum:
Thanks @UlrichKu

PR #109, opened 2018-12-06
ref #190: Replace outdated translation macro usage

UlrichKu:
| Q             | A
| ------------- | ---
| Branch        | master
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Fixed issues  | chameleon-system/chameleon-system#190
| License       | MIT


umulmrum:
The trans tag still needs to be functional - I proposed another PR that should tackle the problem more at its root, see #111.

UlrichKu:
This is however the only place where {% trans %} is used...

PR #112, opened 2018-12-07
ref #19: Deprecate TCMSFieldMediaProperties

UlrichKu:
| Q             | A
| ------------- | ---
| Branch        | master
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | yes
| Tests pass?   | yes
| Fixed issues  | chameleon-system/chameleon-system#19
| License       | MIT

There is only one usage which should be tackled by https://github.com/chameleon-system/chameleon-system/issues/220.

### DIFF
--- a/src/CoreBundle/private/library/classes/TCMSFields/TCMSFieldMediaProperties.class.php
+++ b/src/CoreBundle/private/library/classes/TCMSFields/TCMSFieldMediaProperties.class.php
@@ -11,6 +11,9 @@
 
 use ChameleonSystem\CoreBundle\ServiceLocator;
 
+/**
+ * @deprecated since 6.3.0 - it is a very old mangling of a "number"
+ */
 class TCMSFieldMediaProperties extends TCMSFieldNumber
 {
     /** @var bool */
--- a/src/CoreBundle/private/library/classes/TCMSFields/TCMSFieldMediaProperties.class.php
+++ b/src/CoreBundle/private/library/classes/TCMSFields/TCMSFieldMediaProperties.class.php
@@ -12,7 +12,7 @@
 use ChameleonSystem\CoreBundle\ServiceLocator;
 
 /**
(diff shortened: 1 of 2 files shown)


src/CoreBundle/private/library/classes/TCMSFields/TCMSFieldMediaProperties.class.php
```diff
@@ -11,6 +11,9 @@
 
 use ChameleonSystem\CoreBundle\ServiceLocator;
 
+/**
+ * @deprecated since 6.3.0 - it is a very old mangling of a "number"
```
umulmrum:
```suggestion
 * @deprecated since 6.3.0 - no longer used
```

UlrichKu: I like my version. Maybe we can combine them? :-)

umulmrum: Not quite sure - your version is mostly judging the oddity of the class. There is already a small number of deprecation messages in the code that do also judge, but we should restrict ourselves as far as possible to providing useful information for developers who find a deprecated entity and wonder what to do. In this case here we should simply state that there is no replacement.

umulmrum:
The branch has merge conflicts - could you resolve?

UlrichKu:
> Could you also add a notice in the field type description, so that the type is no longer used for new fields?

You mean in the table "field types"?
There is no type with this class there.
It is only specified as php class for the width field of cms_media.

umulmrum:
> > Could you also add a notice in the field type description, so that the type is no longer used for new fields?
> 
> You mean in the table "field types"?
> There is no type with this class there.
> It is only specified as php class for the width field of cms_media.

Ah OK - that makes it simple.

umulmrum:
Thanks @UlrichKu

PR #115, opened 2018-12-10
ref #219: trim searchterm and change search query to use AND operator…

pixeljunkie:
… for description and path match, search in base language and active language, search only for full ids not parts of the id

| Q             | A
| ------------- | ---
| Branch        | 6.2.x
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed issues  | chameleon-system/chameleon-system#219
| License       | MIT

### DIFF
--- a/src/MediaManagerBundle/Bridge/Chameleon/DataAccess/MediaItemDataAccess.php
+++ b/src/MediaManagerBundle/Bridge/Chameleon/DataAccess/MediaItemDataAccess.php
@@ -227,7 +227,7 @@ class MediaItemDataAccess implements MediaItemDataAccessInterface
         }
 
         $orderBy = $this->getOrderBy($mediaManagerListRequest, $languageId);
-        $query .= $orderBy;
+        $query .= ' '.$orderBy;
 
         return $this->getResultForQueryAndParameters(
             $mediaManagerListRequest,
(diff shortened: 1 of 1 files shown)


src/MediaManagerBundle/Bridge/Chameleon/DataAccess/MediaItemDataAccess.php
```diff
@@ -279,38 +279,62 @@ private function addSearchRestrictions(
...
+        $descriptionSearchQueryParts = [];
         $i = 0;
         foreach ($terms as $term) {
             ++$i;
-            $parts[] = $this->databaseConnection->quoteIdentifier($descriptionFieldName).' LIKE :term'.(string) $i;
-            $parts[] = $this->databaseConnection->quoteIdentifier($pathFieldName).' LIKE :term'.(string) $i;
+
+            $descriptionSearchQueryParts[$descriptionFieldNameTranslated][] = $this->databaseConnection->quoteIdentifier($descriptionFieldNameTranslated).' LIKE :term'.(string) $i;
+
+            if ($descriptionFieldNameTranslated !== $descriptionFieldName) {
```
umulmrum: I wouldn't expect the system to search in other languages as the currently active language. Why would you like to do this?

src/MediaManagerBundle/Bridge/Chameleon/DataAccess/MediaItemDataAccess.php
```diff
@@ -279,38 +279,62 @@ private function addSearchRestrictions(
...
+                $descriptionSearchQueryParts[$pathFieldName][] = $this->databaseConnection->quoteIdentifier($pathFieldName).' LIKE :term'.(string) $i;
+            }
+
             $params['term'.(string) $i] = '%'.$term.'%';
             $paramTypes['term'.(string) $i] = \PDO::PARAM_STR;
         }
 
+        foreach ($descriptionSearchQueryParts as $key => $termItems) {
+            if (count($termItems) > 1) {
+                $parts[] = '('.implode(' AND ', $termItems).')';
```
umulmrum:
```suggestion
                $parts[] = ' ('.implode(' AND ', $termItems).') ';
```

src/MediaManagerBundle/Bridge/Chameleon/DataAccess/MediaItemDataAccess.php
pixeljunkie: It's not some language... it's the active language and the base language, because in 99% of the cases the images where uploaded in the base language and not in a different language... and it would massively disturb users if the uploaded file is gone after they switched the language. The old media manager bahaved like this i'm pretty sure, because there we used the normal list manager and the list manager uses a fallback strategy to the base language.

pixeljunkie: To be clear. The Names shown for the images have a fallback to the base language. So you will see in german the english base language name, but you would not be able to search for them without this change in the active language.

PR #117, opened 2018-12-11
ref #233: Use standard Symfony error rendering

umulmrum:
| Q             | A
| ------------- | ---
| Branch        | master
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed issues  | chameleon-system/chameleon-system#233
| License       | MIT

The fix is to use standard Symfony exception display. The help text from the former template isn't helpful and doesn't justify the maintenance cost and the non-standard error handling.


src/CoreBundle/Resources/views/snippets-cms/exception-debug.html.twig
```diff
@@ -1,31 +1 @@
...
-
-                <div class="alert alert-info" role="alert">Note: you can assign a cms page to this error by creating a system page for the portal with the name error-{{ status_code }}</div>
-            {% endblock %}
-        </div>
-    </div>
-</div>
-</body>
-</html>
\ No newline at end of file
+{% include '@Twig/Exception/exception_full.html.twig' %}
```
UlrichKu: Why is this Twig (@Twig)?

umulmrum: The Symfony standard notation to reference TwigBundle/Resources/views/Exception/exception_full.html.twig without any further Chameleon code in between.

PR #121, opened 2018-12-12
ref #237: Run field translation even if the field already uses the ta…

umulmrum:
…rget setting (BC)

| Q             | A
| ------------- | ---
| Branch        | 6.2.x
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed issues  | chameleon-system/chameleon-system#237
| License       | MIT

### DIFF
--- a/src/CoreBundle/Util/FieldTranslationUtil.php
+++ b/src/CoreBundle/Util/FieldTranslationUtil.php
@@ -219,9 +219,6 @@ class FieldTranslationUtil
         if (false === $fieldConf->sqlData) {
             throw new InvalidArgumentException("Field with ID '$fieldId' not found.");
         }
-        if ($fieldConf->fieldIsTranslatable === $translatable) {
-            return;
-        }
 
         $translatableString = $translatable ? '1' : '0';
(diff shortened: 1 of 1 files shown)


src/CoreBundle/Util/FieldTranslationUtil.php
```diff
@@ -219,9 +219,6 @@ private function changeTranslatableState($fieldId, $translatable)
         if (false === $fieldConf->sqlData) {
             throw new InvalidArgumentException("Field with ID '$fieldId' not found.");
         }
-        if ($fieldConf->fieldIsTranslatable === $translatable) {
```
UlrichKu: That looks rather much for a simple bug?

Why wasn't this a problem until now?

umulmrum: The operations themselves are idempotent - if a field already has the multilanguage flag or the additional language fields, they will be unchanged.

So this change only removes a performance optimization, which isn't that much of a problem during migrations. 

This code changed in 6.2.0, so it seems that simply no updates that rely on the old behavior were executed.

PR #122, opened 2018-12-12
Fix/Unmanaged List Table Handling

ApricumMedia:
| Q             | A
| ------------- | ---
| Branch        | 6.2.x
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| License       | MIT

This request adds required information on table fields in the backend list views to differentiate managed and non-managed tables, introducing a computed property to `TCMSListManager` and its subclasses. If a list manager indicates it is using managed tables, its created tables (like `TGroupTable`) will carry this flag, as will all created fields.

If a field is not part of a managed table, no overlay or inline editing will be allowed. This primarily prevents non-standard tables to be editable in the list view. Request for discussion.

### DIFF
--- a/src/CoreBundle/private/library/classes/TCMSListManager/TCMSListManagerCMSUser.class.php
+++ b/src/CoreBundle/private/library/classes/TCMSListManager/TCMSListManagerCMSUser.class.php
@@ -75,6 +75,13 @@ class TCMSListManagerCMSUser extends TCMSListManagerFullGroupTable
     }
 
     /**
+     * {@inheritDoc}
+     */
+    protected function usesManagedTables(): bool {
+        return false;
+    }
+
+    /**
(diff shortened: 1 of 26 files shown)


src/CoreBundle/private/library/classes/TCMSListManager/TCMSListManagerCMSUser.class.php
```diff
@@ -74,6 +74,13 @@ public function CallBackFunctionBlockDeleteButton($id, $row)
         );
     }
 
+    /**
+     * {@inheritDoc}
+     */
+    protected function usesManagedTables(): bool {
```
pixeljunkie: Please run cs-fixer over your changes. The { should be in the next line.

PR #127, opened 2018-12-13
ref #242: add missing leading slash in relativeSourceUrl to non exact…

pixeljunkie:
… match query condition

Restores the behaviour for non exact match redirects before chameleon-system/chameleon-system#82

| Q             | A
| ------------- | ---
| Branch        | 6.2.x
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed issues  | chameleon-system/chameleon-system#242
| License       | MIT


pSpeck:
@umulmrum no i only fix the exact matches.

umulmrum:
Thanks @pixeljunkie

PR #129, opened 2018-12-13
228 compare correct old field values

UlrichKu:
| Q             | A
| ------------- | ---
| Branch        | 6.2.x
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed issues  | chameleon-system/chameleon-system#228
| License       | MIT

Done: Determine MLT (compare) values correctly. Consider language during field load/comparisson.
This not yet finished (see Todos).

### DIFF
--- a/src/CoreBundle/private/library/classes/dbobjects/TCMSTableConf.class.php
+++ b/src/CoreBundle/private/library/classes/dbobjects/TCMSTableConf.class.php
@@ -154,8 +154,6 @@ class TCMSTableConf extends TCMSRecord
      */
     public function &GetFields(&$oTableRow, $loadDefaults = false, $bDoNotUseAutoObjects = false)
     {
-        $languageService = self::getLanguageService();
-
         $oFieldDefinition = &$this->GetFieldDefinitions(array(), $bDoNotUseAutoObjects);
         $oFields = new TIterator();
         while ($oFieldDef = $oFieldDefinition->next()) {
@@ -193,7 +191,7 @@ class TCMSTableConf extends TCMSRecord
                         // Standard case
                         // Try to find the correct field name for the data in $oTableRow with respect to the language
 
-                        $data = $this->getDataForCurrentLanguage($oFieldDef, $oTableRow->sqlData, $languageService);
+                        $data = $this->getDataForCurrentLanguage($oFieldDef, $oTableRow->sqlData);
 
                         if (null === $data) {
                             $data = $oTableRow->sqlData[$oField->name];
@@ -216,11 +214,10 @@ class TCMSTableConf extends TCMSRecord
     /**
      * @param TdbCmsFieldConf|TCMSFieldDefinition $fieldDefinition
      * @param array                               $sqlData
-     * @param LanguageServiceInterface            $languageService
      *
      * @return mixed
      */
-    private function getDataForCurrentLanguage($fieldDefinition, array $sqlData, LanguageServiceInterface $languageService)
+    private function getDataForCurrentLanguage($fieldDefinition, array $sqlData)
     {
         $languageId = $this->GetLanguage();
 
@@ -228,7 +225,7 @@ class TCMSTableConf extends TCMSRecord
             return null;
         }
 
-        $language = $languageService->getLanguage($languageId);
+        $language = self::getLanguageService()->getLanguage($languageId);
 
         if (null === $language) {
             return null;
--- a/src/CoreBundle/private/library/classes/dbobjects/TCMSTableConf.class.php
+++ b/src/CoreBundle/private/library/classes/dbobjects/TCMSTableConf.class.php
@@ -155,7 +155,6 @@ class TCMSTableConf extends TCMSRecord
     public function &GetFields(&$oTableRow, $loadDefaults = false, $bDoNotUseAutoObjects = false)
     {
         $languageService = self::getLanguageService();
-        $currentLanguage = $languageService->getLanguage($this->GetLanguage());
 
         $oFieldDefinition = &$this->GetFieldDefinitions(array(), $bDoNotUseAutoObjects);
         $oFields = new TIterator();
--- a/src/CoreBundle/private/library/classes/dbobjects/TCMSTableConf.class.php
+++ b/src/CoreBundle/private/library/classes/dbobjects/TCMSTableConf.class.php
@@ -190,15 +190,17 @@ class TCMSTableConf extends TCMSRecord
                     if (array_key_exists($oField->name.'-original', $oTableRow->sqlData)) {
                         $oField->data = $oTableRow->sqlData[$oField->name.'-original'];
                     } else {
+                        // Standard case
+                        // Try to find the correct field name for the data in $oTableRow with respect to the language
+
                         $fieldName = $oField->name;
 
                         if ($language !== null) {
-                            $fieldName = $oFieldDef->GetEditFieldNameForLanguage($language);
-                        }
+                            $fieldNameForLanguage = $oFieldDef->GetEditFieldNameForLanguage($language);
 
-                        if (false === \array_key_exists($fieldName, $oTableRow->sqlData)) {
-                            // Could be a Save() - containing only mono-language data
-                            $fieldName = $oField->name;
+                            if (true === \array_key_exists($fieldNameForLanguage, $oTableRow->sqlData)) {
+                                $fieldName = $fieldNameForLanguage;
+                            } // else could be a Save() - containing only mono-language data
                         }
 
                         $oField->data = $oTableRow->sqlData[$fieldName];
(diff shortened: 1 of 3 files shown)


src/CoreBundle/private/library/classes/dbobjects/TCMSTableConf.class.php
```diff
@@ -152,11 +152,20 @@ public function GetCmsTblListExtension($sExtensionName)
...
     public function &GetFields(&$oTableRow, $loadDefaults = false, $bDoNotUseAutoObjects = false)
     {
+        // TODO this could also be a parameter here?
+        $languageService = self::getLanguageService();
+        $language = $languageService->getActiveEditLanguage();
+
         $oFieldDefinition = &$this->GetFieldDefinitions(array(), $bDoNotUseAutoObjects);
         $oFields = new TIterator();
-        while ($oFieldDef = $oFieldDefinition->Next()) {
+        while ($oFieldDef = $oFieldDefinition->next()) {
```
umulmrum: I saw only now that PHP-CS-Fixer lower-cased that method name, which was not intended (I filed an issue: https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/4218).

Not sure what to do. Method names are case-insensitive in PHP, and the lower-cased name conforms to our coding standards, but I don't like surprise features that much. Leave it for now, but maybe we'll revert to upper-case.

src/CoreBundle/private/library/classes/dbobjects/TCMSTableConf.class.php
```diff
@@ -152,11 +152,20 @@ public function GetCmsTblListExtension($sExtensionName)
      */
     public function &GetFields(&$oTableRow, $loadDefaults = false, $bDoNotUseAutoObjects = false)
     {
+        // TODO this could also be a parameter here?
```
umulmrum: This class derives from TCMSRecord which has a `GetLanguage()` method. I'd suppose that language should be correct.

src/CoreBundle/private/library/classes/dbobjects/TCMSTableConf.class.php
```diff
@@ -181,7 +190,21 @@ public function &GetFields(&$oTableRow, $loadDefaults = false, $bDoNotUseAutoObj
...
                         $oField->data = $oTableRow->sqlData[$oField->name.'-original'];
                     } else {
-                        $oField->data = $oTableRow->sqlData[$oField->name];
+                        // Standard case
+                        // Try to find the correct field name for the data in $oTableRow with respect to the language
+
+                        $fieldName = $oField->name;
+
+                        if ($language !== null) {
+                            // TODO GetEditFieldNameForLanguage() returns wrong if the base translation language === current language
```
umulmrum: "wrong" --> "false"?

Is this comment really needed?

src/CoreBundle/private/library/classes/dbobjects/TCMSTableConf.class.php
UlrichKu: next() and current() are called like that in TIterator (since first cs-fixer run there probably).

So this here seems to be ok (and no bug).

src/CoreBundle/private/library/classes/dbobjects/TCMSTableConf.class.php
UlrichKu: `GetLanguage()` returns the language id - so not directly usable here.

Also `getActiveEditLanguage()` makes it more clear what is meant here.

src/CoreBundle/private/library/classes/dbobjects/TCMSTableConf.class.php
UlrichKu: Removed

Couldn't reproduce a problem here just now.

UlrichKu:
Noticed just now that there also is a `GetField()` method (parallel to `GetFields()`) which also sets the field's data without considering the language.

src/CoreBundle/private/library/classes/dbobjects/TCMSTableConf.class.php
umulmrum: That's what I meant: PHP-CS-Fixer lower-cased the methods in TIterator when we applied it to the code base last year. Your change can stay as is.

src/CoreBundle/private/library/classes/dbobjects/TCMSTableConf.class.php
umulmrum: Then we should use the LanguageService to lookup the language object from that ID. If possible, we should use the Tdb's language setting in general and avoid the dependency to the "global" language. This makes it easier to use Tdbs in other contexts (in cases where we'd like to compute things for other languages than the current one).

UlrichKu: ok

UlrichKu:
> Could you investigate what the consequences are if we also apply the change to GetField()? The methods differ slightly, so it's maybe a bad idea to call GetField() from GetFields(), but maybe we can move the new code to another method and call it from both methods if that makes sense.

Did the latter.

However that method is also widely used. For example in exports or in "edit mode".

src/CoreBundle/private/library/classes/dbobjects/TCMSTableConf.class.php
UlrichKu: GetLanguage() also relies on the/a language being set in the current object.

I'm not sure this is the case in all relevant cases?

UlrichKu: Checking now if a language exists/is set at all.

src/CoreBundle/private/library/classes/dbobjects/TCMSTableConf.class.php
```diff
@@ -152,11 +154,19 @@ public function GetCmsTblListExtension($sExtensionName)
      */
     public function &GetFields(&$oTableRow, $loadDefaults = false, $bDoNotUseAutoObjects = false)
     {
+        $languageService = self::getLanguageService();
+        $currentLanguage = $languageService->getLanguage($this->GetLanguage());
```
umulmrum: LanguageServiceInterface::getLanguage() expects a string argument, but $this->GetLanguage() might return null (although its signature states that it's always a string). We should be strict and handle this case explicitly.

src/CoreBundle/private/library/classes/dbobjects/TCMSTableConf.class.php
```diff
@@ -195,6 +214,36 @@ public function &GetFields(&$oTableRow, $loadDefaults = false, $bDoNotUseAutoObj
         return $oFields;
     }
 
+    /**
+     * @param TdbCmsFieldConf|TCMSFieldDefinition $fieldDefinition
+     * @param array                               $sqlData
+     * @param LanguageServiceInterface            $languageService
+     * @return mixed|null
```
umulmrum:
```suggestion
     *
     * @return mixed
```

src/CoreBundle/private/library/classes/dbobjects/TCMSTableConf.class.php
```diff
@@ -195,6 +214,36 @@ public function &GetFields(&$oTableRow, $loadDefaults = false, $bDoNotUseAutoObj
...
+     */
+    private function getDataForCurrentLanguage($fieldDefinition, array $sqlData, LanguageServiceInterface $languageService)
```
umulmrum: The language is the same for every call, as it only relies on $this. So we should retrieve the language in the calling method and pass it instead of the LanguageService.

UlrichKu: But it might be null - so every caller would need to handle that case.

src/CoreBundle/private/library/classes/dbobjects/TCMSTableConf.class.php
UlrichKu: ok

umulmrum:
There are conflicts - could you have a look?

UlrichKu:
> There are conflicts - could you have a look?

done

src/CoreBundle/private/library/classes/dbobjects/TCMSTableConf.class.php
umulmrum: Then I'd prefer retrieving the LanguageService also within the method. There will be a minimal performance hit, but having the callers retrieve and pass it, looks a bit odd.

UlrichKu: ok

umulmrum:
Thanks @UlrichKu

PR #131, opened 2018-12-19
ref #231: Show other errors than parseerrors after ajax call

UlrichKu:
| Q             | A
| ------------- | ---
| Branch        | 6.2.x
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed issues  | chameleon-system/chameleon-system#231
| License       | MIT

### DIFF
--- a/src/CoreBundle/Resources/public/javascript/cms.js
+++ b/src/CoreBundle/Resources/public/javascript/cms.js
@@ -77,7 +77,7 @@ function AjaxError(XMLHttpRequest, textStatus, errorThrown) {
     }
 
     if (textStatus === 'parsererror') {
-        toasterMessage('Error! Wasn`t able to parse ajax response.', 'ERROR');
+        toasterMessage(CHAMELEON.CORE.i18n.Translate('chameleon_system_core.js.ajax_parse_error'), 'ERROR');
 
         if (XMLHttpRequest.responseText !== '') {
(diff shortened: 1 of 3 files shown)


src/CoreBundle/Resources/public/javascript/cms.js
```diff
@@ -57,31 +57,41 @@ function GetAjaxCallTransparent(url, functionName) {
...
-            var sMessage = CHAMELEON.CORE.i18n.Translate('chameleon_system_core.js.ajax_error', sError);
+        if (responseError.indexOf('<title>') !== -1) {
+            responseError = XMLHttpRequest.statusText;
+        }
 
+        errorMessage = CHAMELEON.CORE.i18n.Translate('chameleon_system_core.js.ajax_error', responseError);
+    }
+
+    if (textStatus === 'parsererror') {
+        toasterMessage('Error! Wasn`t able to parse ajax response.', 'ERROR');
```
umulmrum: Could we take the opportunity to add translation for this error?

And I'd prefer a single quote instead of the backtick in "wasn't" - backticks are technically a bit unexpected and I think we use quotes everywhere.

UlrichKu: ok

umulmrum:
Thanks @UlrichKu

PR #133, opened 2018-12-19
ref #248: fix height of layover container with iframes that have no i…

pixeljunkie:
…nitial height

| Q             | A
| ------------- | ---
| Branch        |6.2.x
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed issues  | chameleon-system/chameleon-system#248
| License       | MIT

### DIFF
--- a/src/MediaManagerBundle/Resources/public/css/mediaManager.css
+++ b/src/MediaManagerBundle/Resources/public/css/mediaManager.css
@@ -106,7 +106,7 @@
 .snippetMediaManagerBackendModuleStandard .edit-container {
     position: relative;
     padding-left: 10px;
-    min-height: 250px;
+    min-height: 450px;
 }
 
 .snippetMediaManagerBackendModuleStandard .snippetMediaManagerListHeader {
--- a/src/MediaManagerBundle/Resources/public/css/mediaManager.css
+++ b/src/MediaManagerBundle/Resources/public/css/mediaManager.css
@@ -106,7 +106,7 @@
 .snippetMediaManagerBackendModuleStandard .edit-container {
     position: relative;
     padding-left: 10px;
-    min-height: 450px;
+    min-height: 500px;
 }
 
 .snippetMediaManagerBackendModuleStandard .snippetMediaManagerListHeader {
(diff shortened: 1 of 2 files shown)


src/MediaManagerBundle/Resources/public/js/mediaManager.js
```diff
@@ -1014,7 +1014,13 @@
             var layover = self.openLayover(title, '');
 
             if (typeof height === 'undefined') {
-                height = layover.height() - layover.find('.title').height() - 50;
+
+                var height = layover.height();
```
UlrichKu: This should be sufficient: `height = $('#edit-container').height() - (layover.find('.title').height() - 50);`

The layover (opened with no html) will never have a useful height here.

src/MediaManagerBundle/Resources/public/css/mediaManager.css
```diff
@@ -106,7 +106,7 @@
 .snippetMediaManagerBackendModuleStandard .edit-container {
     position: relative;
     padding-left: 10px;
-    min-height: 250px;
+    min-height: 450px;
```
UlrichKu: With 500px also the upload bar that appears later on will fit.

src/MediaManagerBundle/Resources/public/js/mediaManager.js
```diff
@@ -1014,7 +1014,7 @@
             var layover = self.openLayover(title, '');
 
             if (typeof height === 'undefined') {
-                height = layover.height() - layover.find('.title').height() - 50;
+                height = $('#edit-container').height() - (layover.find('.title').height() - 50);
```
UlrichKu: This ` - (layover.find('.title').height() - 50)` is now actually a + 24px.

src/MediaManagerBundle/Resources/public/js/mediaManager.js
pixeljunkie: You are right. simplified the code.

src/MediaManagerBundle/Resources/public/css/mediaManager.css
pixeljunkie: thanks for finding out. changed it.

src/MediaManagerBundle/Resources/public/js/mediaManager.js
pixeljunkie: I'm not 100% sure what this calculations reason was anyway? Should we change this again?

PR #136, opened 2019-01-03
WIP: 170 add tag for mappers

UlrichKu:
| Q             | A
| ------------- | ---
| Branch        | master
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed issues  | chameleon-system/chameleon-system#170
| License       | MIT

Upgrade documentation (if needed) is missing.

### DIFF
--- a/src/NewsletterBundle/Resources/config/services.xml
+++ b/src/NewsletterBundle/Resources/config/services.xml
@@ -5,6 +5,10 @@
            xsi:schemaLocation="http://symfony.com/schema/dic/services http://symfony.com/schema/dic/services/services-1.0.xsd">
 
     <services>
+        <instanceof id="\IViewMapper" autoconfigure="false">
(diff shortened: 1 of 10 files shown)


src/NewsletterBundle/Resources/config/services.xml
```diff
@@ -5,6 +5,10 @@
            xsi:schemaLocation="http://symfony.com/schema/dic/services http://symfony.com/schema/dic/services/services-1.0.xsd">
 
     <services>
+        <instanceof id="\IViewMapper" autoconfigure="false">
```
UlrichKu: This is the strangest of the three autoconfiguration variants (here autoconfigure=false).

instanceof and configuration cannot be mixed.

PR #144, opened 2019-01-16
ref #269: Use edit language for old fields

UlrichKu:
| Q             | A
| ------------- | ---
| Branch        | 6.2.x
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed issues  | chameleon-system/chameleon-system#269
| License       | MIT

### DIFF
--- a/src/CoreBundle/private/library/classes/dbobjects/TCMSRecord.class.php
+++ b/src/CoreBundle/private/library/classes/dbobjects/TCMSRecord.class.php
@@ -884,6 +884,11 @@ class TCMSRecord implements IPkgCmsSessionPostWakeupListener
             } else {
                 $this->_oTableConf->LoadFromFieldWithCaching('name', $this->table);
             }
+
+            // TODO alternative 2 - but probably too late ($this->_oTableConf is cached)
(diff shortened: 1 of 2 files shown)


src/CoreBundle/private/library/classes/dbobjects/TCMSRecord.class.php
```diff
@@ -884,6 +884,11 @@ public function &GetTableConf()
             } else {
                 $this->_oTableConf->LoadFromFieldWithCaching('name', $this->table);
             }
+
+            // TODO alternative 2 - but probably too late ($this->_oTableConf is cached)
+            if (null !== $this->iLanguageId) {
```
umulmrum: We can set the language in line 873 (`GetNewInstance(null, $this->iLanguageId)`) - there the null case is also already handled. I think this could replace all three changes in this PR.

UlrichKu: done

umulmrum:
Thanks @UlrichKu

PR #145, opened 2019-01-16
267 servicify resource collection

UlrichKu:
| Q             | A
| ------------- | ---
| Branch        | 6.2.x
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed issues  | chameleon-system/chameleon-system#267
| License       | MIT

### DIFF
--- a/src/CoreBundle/Controller/ChameleonController.php
+++ b/src/CoreBundle/Controller/ChameleonController.php
@@ -138,7 +138,6 @@ abstract class ChameleonController implements ChameleonControllerInterface
         RequestStack $requestStack,
         EventDispatcherInterface $eventDispatcher,
         PortalDomainServiceInterface $portalDomainService,
-        ResourceCollectorInterface $resourceCollector,
         TModuleLoader $moduleLoader,
         IViewPathManager $viewPathManager = null
     ) {
@@ -148,7 +147,6 @@ abstract class ChameleonController implements ChameleonControllerInterface
         $this->viewPathManager = $viewPathManager;
         $this->eventDispatcher = $eventDispatcher;
         $this->portalDomainService = $portalDomainService;
-        $this->resourceCollector = $resourceCollector;
     }
 
     /**
@@ -948,5 +946,10 @@ abstract class ChameleonController implements ChameleonControllerInterface
     public function setInputFilterUtil(InputFilterUtilInterface $inputFilterUtil)
     {
         $this->inputFilterUtil = $inputFilterUtil;
+    }
+
+    public function setResourceCollector(ResourceCollectorInterface $resourceCollector): void
+    {
+        $this->resourceCollector = $resourceCollector;
     }
 }
--- a/src/CoreBundle/Controller/ChameleonController.php
+++ b/src/CoreBundle/Controller/ChameleonController.php
@@ -13,6 +13,7 @@ namespace ChameleonSystem\CoreBundle\Controller;
 
 use ChameleonSystem\CoreBundle\CoreEvents;
 use ChameleonSystem\CoreBundle\Event\HtmlIncludeEvent;
+use ChameleonSystem\CoreBundle\Interfaces\ResourceCollectorInterface;
 use ChameleonSystem\CoreBundle\Security\AuthenticityToken\AuthenticityTokenManagerInterface;
 use ChameleonSystem\CoreBundle\Service\ActivePageServiceInterface;
 use ChameleonSystem\CoreBundle\Service\PortalDomainServiceInterface;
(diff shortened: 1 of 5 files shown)


src/CoreBundle/Resources/config/services.xml
```diff
@@ -789,6 +790,12 @@
             <argument type="service" id="request_stack"/>
         </service>
 
+        <service id="chameleon_system_core.resource_collector" class="TCMSResourceCollection">
```
umulmrum: The class is not stateless, so we should set the service to non-shared. One could argue that the state is overwritten before it is used, but we should be conservative here.

src/CoreBundle/private/library/classes/TCMSResourceCollection.class.php
```diff
@@ -31,20 +32,40 @@ class TCMSResourceCollection
...
+     */
+    private $portalDomainService;
+
+    /**
+     * @var EventDispatcherInterface
+     */
+    private $eventDispatcher;
+
+    public function __construct(
+        IPkgCmsFileManager $cmsFileManager,
```
umulmrum: Changing constructor arguments is a BC break (the class didn't implement an interface before). We can remedy this by making arguments optional, similar to e.g. `TMapper_ViewPortSwitch`.

src/CoreBundle/Interfaces/ResourceCollectorInterface.php
```diff
@@ -0,0 +1,23 @@
+<?php
+
```
umulmrum: Please add the copyright header.

src/CoreBundle/Controller/ChameleonController.php
```diff
@@ -132,6 +138,7 @@ public function __construct(
         RequestStack $requestStack,
         EventDispatcherInterface $eventDispatcher,
         PortalDomainServiceInterface $portalDomainService,
+        ResourceCollectorInterface $resourceCollector,
```
umulmrum: I'd like to avoid having to adjust constructor arguments for the controller, as custom implementations will rely on this class instead of the interface, and will then break. Could you use setter injection as we already do it for some other dependencies?

src/CoreBundle/Resources/config/services.xml
UlrichKu: ok

src/CoreBundle/private/library/classes/TCMSResourceCollection.class.php
UlrichKu: ok

Does that work there? I would expect the type to begin with ? -> ?ActivePageServiceInterface

src/CoreBundle/Interfaces/ResourceCollectorInterface.php
UlrichKu: ok

src/CoreBundle/Controller/ChameleonController.php
UlrichKu: ok

Will such things be changed for example for 6.3?

src/CoreBundle/private/library/classes/TCMSResourceCollection.class.php
umulmrum: `?`is not required, because null as default value automatically allows null values. But it doesn't hurt either, so it's fine as is now.

src/CoreBundle/Controller/ChameleonFrontendController.php
```diff
@@ -36,17 +36,15 @@ class ChameleonFrontendController extends ChameleonController
...
-    /**
-     * @param RequestStack                       $requestStack
-     * @param EventDispatcherInterface           $eventDispatcher
-     * @param PortalDomainServiceInterface       $portalDomainService
-     * @param TModuleLoader                      $moduleLoader
-     * @param IViewPathManager|null              $viewPathManager
-     * @param ContainerInterface                 $container
-     * @param TPkgViewRendererConfigToLessMapper $configToLessMapper
-     */
-    public function __construct(RequestStack $requestStack, EventDispatcherInterface $eventDispatcher, PortalDomainServiceInterface $portalDomainService, TModuleLoader $moduleLoader, $viewPathManager, ContainerInterface $container, TPkgViewRendererConfigToLessMapper $configToLessMapper)
```
umulmrum: Just a note that the Symfony coding standards require to declare all method arguments in a single line (https://symfony.com/doc/current/contributing/code/standards.html under "Structure"). I'm not the biggest fan of that rule, so it's OK for me to wrap lines. Maybe we enforce the rule in the future, though.

umulmrum:
Thanks @UlrichKu

PR #146, opened 2019-01-17
134 running cronjobs

UlrichKu:
| Q             | A
| ------------- | ---
| Branch        | master
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed issues  | chameleon-system/chameleon-system#134
| License       | MIT

### DIFF
--- a/src/CoreBundle/Command/ActiveCronjobCommand.php
+++ b/src/CoreBundle/Command/ActiveCronjobCommand.php
@@ -17,7 +17,7 @@ use Symfony\Component\Console\Command\Command;
 use Symfony\Component\Console\Input\InputInterface;
 use Symfony\Component\Console\Output\OutputInterface;
 
-class DisableCronjobsCommand extends Command
+class ActiveCronjobCommand extends Command
 {
     /**
      * @var CronjobEnablingServiceInterface
@@ -26,7 +26,7 @@ class DisableCronjobsCommand extends Command
 
     public function __construct(CronjobEnablingServiceInterface $cronjobEnablingService)
     {
-        parent::__construct('chameleon_system:cronjobs:disable');
+        parent::__construct('chameleon_system:cronjobs:active_check');
 
         $this->cronjobEnablingService = $cronjobEnablingService;
     }
@@ -37,9 +37,9 @@ class DisableCronjobsCommand extends Command
     protected function configure()
     {
         $this
-            ->setDescription('Disables the cron job execution')
+            ->setDescription('Checks if any cron job is running')
             ->setHelp(<<<EOF
-The <info>%command.name%</info> command disables all cronjob execution.
+The <info>%command.name%</info> command outputs "true" if a cron job is currently running. Otherwise "false".
 EOF
             )
         ;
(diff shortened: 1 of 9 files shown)


UlrichKu:
Though during testing/reviewing I noticed:

- It might better to call it "running" instead of "active" - in the command name for example?
- If it prints "true" it might be nice to know which ones are blocked - at least from the command line. The cronjob table also does not show which ones are locked.
- Potentially this is a cronjob that is locked because of an error during execution and the lock has not yet been reset. Which does - at least on development systems - never go away.

src/CoreBundle/Command/ActiveCronjobCommand.php
```diff
@@ -0,0 +1,67 @@
...
+
+namespace ChameleonSystem\CoreBundle\Command;
+
+use ChameleonSystem\CoreBundle\CronJob\CronjobEnablingServiceInterface;
+use ChameleonSystem\CoreBundle\Exception\CronjobHandlingException;
+use Symfony\Component\Console\Command\Command;
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Output\OutputInterface;
+
+class ActiveCronjobCommand extends Command
```
umulmrum: I agree to your thought - "running" instead of "active" is clearer and unambiguous.

Commands should have a "command-like" imperative name - maybe something like `GetCronjobRunStateCommand`?

src/CoreBundle/CronJob/CronjobEnablingServiceInterface.php
```diff
@@ -26,12 +26,17 @@
...
     /**
-     * @throws CronjobEnableException
+     * @throws CronjobHandlingException
      */
     public function disableCronjobExecution(): void;
+
+    /**
+     * @throws CronjobHandlingException
+     */
+    public function isOneCronjobRunning(): bool;
```
umulmrum: How about `isCronjobRunning()`? We are interested in the information if at least one cronjob is running, not exactly one (in practice this is currently the same on the implementation level, but I wouldn't say that this is by concept).

And I think this method doesn't really fit in the interface. We would ask a service that is responsible for activation and deactivation if there's a cronjob currently running - doesn't really sound right.

src/CoreBundle/Command/ActiveCronjobCommand.php
UlrichKu: Using "GetCronjobsStateCommand"

src/CoreBundle/CronJob/CronjobEnablingServiceInterface.php
UlrichKu: Ok introduced CronjobStateService.

src/CoreBundle/Command/ActiveCronjobCommand.php
umulmrum: Sounds good.

src/CoreBundle/CronJob/CronjobStateServiceInterface.php
```diff
@@ -0,0 +1,13 @@
+<?php
+
```
umulmrum: Please add the copyright header.

src/CoreBundle/CronJob/CronjobStateService.php
```diff
@@ -0,0 +1,33 @@
+<?php
+
```
umulmrum: Please add the copyright header and run PHP-CS-Fixer.

src/CoreBundle/CronJob/CronjobStateServiceInterface.php
UlrichKu: ok

src/CoreBundle/CronJob/CronjobStateService.php
UlrichKu: ok

UlrichKu:
> Regarding your comments:
> 
> I think it's OK for now to print true and false - we can add details later.

It is now "running" and "idle" due to the change of class/command names.

src/CoreBundle/CronJob/CronjobStateServiceInterface.php
umulmrum: Forgot to push? :-)

umulmrum:
Thanks @UlrichKu

PR #150, opened 2019-01-21
Feature/204 add select2 plugin

ckleis:
| Q             | A
| ------------- | ---
| Branch        | features
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no  
| Deprecations? | no 
| Tests pass?   | yes    
| Fixed issues  | chameleon-system/chameleon-system#204
| License       | MIT

select2-plugin added to the table manager search field with record auto-completion and allows you to select and edit a record directly.


src/CoreBundle/private/library/classes/TFullGroupTable/TFullGroupTable.class.php
```diff
@@ -866,10 +921,22 @@ protected function _BuildFilterSection()
 
         if ($this->showSearchBox) {
             $this->somethingToShow = true;
+
+            $filterContent .= '<div class="form-group mr-2">';
+            $filterContent .= '<select id="searchLookup" name="_search_word" data-listname="'. TGlobal::OutHTML($this->listName) .'" class="form-control" data-select2-placeholder="';
```
umulmrum: Could you use sprintf for all lines with nontrivial string concatenations? The concatenations are quite hard to read and it's tempting to omit output escaping because it makes these lines even harder to read.

This line could then be combined with the following 2 lines so that the entire select start tag is contained within one sprint statement.

src/CoreBundle/private/library/classes/TFullGroupTable/TFullGroupTable.class.php
```diff
@@ -866,10 +921,22 @@ protected function _BuildFilterSection()
...
+            $filterContent .= $this->searchFieldText;
+            $filterContent .= '" data-select2-ajax="'. $this->getAjaxRecordsUrl() .'">';
```
umulmrum: Maybe we should escape this value.

src/CoreBundle/private/library/classes/TFullGroupTable/TFullGroupTable.class.php
```diff
@@ -866,10 +921,22 @@ protected function _BuildFilterSection()
...
+            }
+            $filterContent .= '</select></div>';
+
             $filterContent .= '<div class="form-group">';
-            $filterContent .= '<input type="text" class="form-control form-control-sm" name="_search_word" value="'.TGlobal::OutHTML($this->_postData['_search_word'])."\" onKeyDown=\"if (window.event && window.event.keyCode == 13) document.{$this->listName}._startRecord.value=0\" onChange=\"document.{$this->listName}._startRecord.value=0;document.{$this->listName}.submit();\" placeholder=\"".$this->searchFieldText."\">\n";
-            $filterContent .= "<input type=\"button\" class=\"form-control form-control-sm btn-sm btn-primary\" value=\"{$this->searchButtonText}\" onClick=\"document.{$this->listName}._startRecord.value=0;document.{$this->listName}.submit();\" class=\"btn btn-sm btn-primary\">
-            </div>";
+            $filterContent .= '<input type="button" class="form-control form-control-sm btn-sm btn-primary" value="'.$this->searchButtonText.'" onClick="document.'.$this->listName.'._startRecord.value=0;document.'.$this->listName.'.submit();" class="btn btn-sm btn-primary">';
```
umulmrum: `$this->searchButtonText` needs to be escaped.

src/CoreBundle/private/library/classes/TFullGroupTable/TFullGroupTable.class.php
```diff
@@ -415,6 +415,61 @@ private function getInlineFromAttributes(array $attributes)
         return $inline;
     }
 
+    private function getTableEditorPagedef($pagedef): string
```
umulmrum: Please order these new methods so that they appear after the method in which they were called first.

src/CoreBundle/private/library/classes/TFullGroupTable/TFullGroupTable.class.php
```diff
@@ -415,6 +415,61 @@ private function getInlineFromAttributes(array $attributes)
...
+    {
+        $inputFilterUtil = $this->getInputFilterUtil();
+        $customTableEditor = $inputFilterUtil->getFilteredInput('sTableEditorPagdef');
```
umulmrum: We could add the empty string as second argument, so that the comparison in the line below doesn't need to check for null.

src/CoreBundle/private/library/classes/TFullGroupTable/TFullGroupTable.class.php
```diff
@@ -415,6 +415,61 @@ private function getInlineFromAttributes(array $attributes)
...
+
+        return $sAjaxURL;
+    }
+
+    private function switchToSelectedRecord(): string
+    {
+        $pagedef = $this->getTableEditorPagedef('tableeditor');
+        $inputFilterUtil = $this->getInputFilterUtil();
+        $tableId = $inputFilterUtil->getFilteredInput('id');
+        $urlUtil = ServiceLocator::get('chameleon_system_core.util.url');
```
umulmrum: Could you move the ServiceLocator call to a separate method that includes a return type declaration? This improves type safety and keeps the distracting details of service retrieval away.

src/CoreBundle/private/library/classes/TFullGroupTable/TFullGroupTable.class.php
```diff
@@ -415,6 +415,61 @@ private function getInlineFromAttributes(array $attributes)
...
+            '_fnc' => 'getAutocompleteRecordList',
+            'sRestrictionField' => '',
+            'sRestriction' => '',
+            'recordID' => ''
+        ], PATH_CMS_CONTROLLER.'?', '&');
+
+        return $sAjaxURL;
+    }
+
+    private function switchToSelectedRecord(): string
```
umulmrum: Can we somehow move this JS code to a JS file? It's strange to add JS code conditionally - I'd advocate for providing this code always, but block usage in another way where it shouldn't be used.

If we cannot remove it, we should rename it, as it doesn't actually switch but only provides the code needed for switching. How about `GetSwitchToRecordCode()`?

src/CoreBundle/private/library/classes/TFullGroupTable/TFullGroupTable.class.php
```diff
@@ -415,6 +415,61 @@ private function getInlineFromAttributes(array $attributes)
...
+    {
+        $inputFilterUtil = $this->getInputFilterUtil();
+        $customTableEditor = $inputFilterUtil->getFilteredInput('sTableEditorPagdef');
+        if (null !== $customTableEditor && '' !== $customTableEditor) {
+            $pagedef = $customTableEditor;
+        }
+        return $pagedef;
+    }
+
+    private function getAjaxRecordsUrl(): string
```
umulmrum: How about `getRecordAutocompleteUrl()`?

src/CoreBundle/private/library/classes/TFullGroupTable/TFullGroupTable.class.php
pixeljunkie: I'm pretty sure, we had this first, but escaping it here broke the URL for the selec2 plugin. Maybe OutHTML was not the right method here. But we should try this first.

src/CoreBundle/private/library/classes/TFullGroupTable/TFullGroupTable.class.php
ckleis: Done.

src/CoreBundle/private/library/classes/TFullGroupTable/TFullGroupTable.class.php
ckleis: it works with OutHTML, so I used it for $this->getAjaxRecordsUrl()

src/CoreBundle/private/library/classes/TFullGroupTable/TFullGroupTable.class.php
ckleis: Done.

src/CoreBundle/private/library/classes/TFullGroupTable/TFullGroupTable.class.php
ckleis: Done.

src/CoreBundle/private/library/classes/TFullGroupTable/TFullGroupTable.class.php
ckleis: Done.

src/CoreBundle/private/library/classes/TFullGroupTable/TFullGroupTable.class.php
ckleis: The JS code here is not conditional, the problem is that it contains a recordUrl that comes from php.
This function "switchRecord()" is called by the select2 function in table.js, actually to change the record. Hmm, I think the name says what the function does.

src/CoreBundle/private/library/classes/TFullGroupTable/TFullGroupTable.class.php
ckleis: Changed.

src/CoreBundle/private/library/classes/TFullGroupTable/TFullGroupTable.class.php
umulmrum: Not yet done :-)

I meant to move all three methods under `_BuildFilterSection`, as they are called in that method.

ckleis: OK, now under ``_BuildFilterSection`` ;-)

src/CoreBundle/private/library/classes/TFullGroupTable/TFullGroupTable.class.php
```diff
@@ -907,4 +979,9 @@ private function getInputFilterUtil()
     {
         return ServiceLocator::get('chameleon_system_core.util.input_filter');
     }
+
+    private function getUrlUtil()
```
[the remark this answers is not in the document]
umulmrum: No changes yet (please use return type declarations, not docblocks for new methods).

umulmrum:
Thanks @ckleis

PR #156, opened 2019-01-23
ref #278: enable configurable variable(s) for the less compiler

UlrichKu:
| Q             | A
| ------------- | ---
| Branch        | master
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | yes
| Deprecations? | no
| Tests pass?   | yes
| Fixed issues  | chameleon-system/chameleon-system#278
| License       | MIT

The change to an upgrade or changelog file is missing as probably "UPGRADE-6.3.md" might not be the correct one.
(Change/added feature and changed constructor.)


pixeljunkie:
@UlrichKu Could you please fix the merge conflicts here.

UlrichKu:
Also added upgrade notice.

UPGRADE-7.0.md
```diff
@@ -19,6 +19,12 @@ project requires WYSIWYG JavaScript, allow it by setting the configuration key
 
 # Changed Features
 
+## Less Compilation supports STATIC_CONTENT_URL
```
umulmrum:
```suggestion
## Less Compilation Supports STATIC_CONTENT_URL
```

UPGRADE-7.0.md
```diff
@@ -19,6 +19,12 @@ project requires WYSIWYG JavaScript, allow it by setting the configuration key
...
+
+The less compiler TPkgViewRendererLessCompiler now supports an additional argument for variables that are passed to
+less files. Currently this is used to inject STATIC_CONTENT_URL which can be used to load resources from another system
+(ie a CDN). This in turn can be configured with `chameleon_system_view_renderer.less_compiler: static_content_url`.
+
```
umulmrum: Could you move the new sections to a new file `UPGRADE-7.1md`?

UPGRADE-7.0.md
```diff
@@ -19,6 +19,12 @@ project requires WYSIWYG JavaScript, allow it by setting the configuration key
 
 # Changed Features
 
+## Less Compilation supports STATIC_CONTENT_URL
+
+The less compiler TPkgViewRendererLessCompiler now supports an additional argument for variables that are passed to
+less files. Currently this is used to inject STATIC_CONTENT_URL which can be used to load resources from another system
+(ie a CDN). This in turn can be configured with `chameleon_system_view_renderer.less_compiler: static_content_url`.
```
umulmrum:
```suggestion
(e.g. a CDN). This in turn can be configured with `chameleon_system_view_renderer.less_compiler: static_content_url`.
```

src/ViewRendererBundle/DependencyInjection/Configuration.php
```diff
@@ -22,6 +22,9 @@ public function getConfigTreeBuilder(): TreeBuilder
                     ->defaultValue('chameleon/outbox/static/less')
                     ->info('Path used as working and output directory for the less compiler, relative to the web root directory. CAUTION: All files in this directory are deleted on cache clear.')
                 ->end()
+                ->scalarNode('static_content_url')
+                    ->defaultValue('')
+                    ->info('URL for static content (images, fonts, ...) if served from a different location (i. e. a CDN). Can be omitted. Default is empty.')
```
umulmrum:
```suggestion
                    ->info('URL for static content (images, fonts, ...) if served from a different location (e.g. a CDN). Can be omitted. Default is empty.')
```

src/ViewRendererBundle/DependencyInjection/ChameleonSystemViewRendererExtension.php
```diff
@@ -28,5 +28,8 @@ protected function loadInternal(array $mergedConfig, ContainerBuilder $container
 
         $lessCompiler = $container->getDefinition('chameleon_system_view_renderer.less_compiler');
         $lessCompiler->replaceArgument(0, $mergedConfig['css_dir']);
+        $lessCompiler->replaceArgument(2, ['STATIC_CONTENT_URL' => '"'.$mergedConfig['static_content_url'].'"']);
+        // NOTE must be included in .less something like: url("@{STATIC_CONTENT_URL}")
```
umulmrum: Both comment lines are quite unclear to me. Could you try and rephrase?

src/ViewRendererBundle/objects/helper/TPkgViewRendererLessCompiler.php
```diff
@@ -23,15 +23,22 @@ class TPkgViewRendererLessCompiler
...
      */
     private $resourceCollectionRefreshPrefix;
 
-    public function __construct(string $cssDirRelativeToWebRoot, string $resourceCollectionRefreshPrefix)
+    /**
+     * @var array
+     */
+    private $additionalVariables;
+
+    public function __construct(string $cssDirRelativeToWebRoot, string $resourceCollectionRefreshPrefix, array $additionalVariables)
```
umulmrum: We could think about adding a method `addLessVariable(string $key, string $value)` instead. Although constructor injection in general is preferred, we would gain a few advantages:
- In projects or third-party bundles additional variables can be defined through compiler passes.
- Users of the class don't need to learn how the array needs to look like; they can pass key and value as separate method arguments. Also type-safety is improved a bit.
- No BC break, so we could think about adding this feature to 6.2.x (only if "necessary" - as this is a feature, I'd prefer master).

UlrichKu: Now using a method addAdditionalVariables().

UPGRADE-7.1.md
```diff
@@ -0,0 +1,12 @@
+UPGRADE FROM 7.0 TO 7.1
```
pixeljunkie: This file needs to be removed and the text needs to be added to the UPGRADE-7.0.md

UlrichKu: ok

PR #161, opened 2019-01-25
ref #281: Update outdated method signature

UlrichKu:
| Q             | A
| ------------- | ---
| Branch        | 6.2.x
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed issues  | chameleon-system/chameleon-system#281
| License       | MIT


umulmrum:
This change will break Twig before 1.27 (our version constraint in chameleon-base is ~1.13), so it won't resolve the problem completely. In master, where we use Twig 2.x, this is already fixed without drawbacks - is it OK for you to wait for the 6.3.0 release? As far as I can see the problem occurs because the template can't be found; if this is a problem in chameleon-base, a fix would be great :-)

UlrichKu:
(See issue)

PR #162, opened 2019-01-25
ref #277: field help texts without popovers

ckleis:
| Q             | A
| ------------- | ---
| Branch        | master
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Fixed issues  | chameleon-system/chameleon-system#277
| License       | MIT

Here a change was changed back.
For version 6.3 the help texts were displayed with popovers. But this is impractical.
Now it was solved again similar to version 6.2, but with a new icon and an alert box info.

### DIFF
--- a/src/CoreBundle/Resources/public/javascript/tableEditor.js
+++ b/src/CoreBundle/Resources/public/javascript/tableEditor.js
@@ -810,7 +810,7 @@ function SetChangedDataMessage() {
 
 CHAMELEON.CORE.MTTableEditor.initHelpTexts = function () {
     $(".help-text-button").click(function () {
-        var helpTextId = '#tooltip-' + $(this).attr("data-helptextId");
+        var helpTextId = '#helptext-' + $(this).attr("data-helptextId");
         $(helpTextId).toggle();
     });
 };
--- a/src/CoreBundle/Resources/public/javascript/tableEditor.js
+++ b/src/CoreBundle/Resources/public/javascript/tableEditor.js
@@ -712,11 +712,7 @@ CHAMELEON.CORE.MTTableEditor.DeleteRecordWithCustomConfirmMessage = function (sC
     }
 };
 
-$(document).ready(function () {
-    CHAMELEON.CORE.MTTableEditor.initTabs();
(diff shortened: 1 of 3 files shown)


src/CoreBundle/private/modules/MTTableEditor/views/includes/fields.inc.php
```diff
@@ -84,8 +84,12 @@
             </div>
     	    </th>
           <td class="rightTD col-10">';
+                if (!empty($oField->oDefinition->sqlData['049_helptext'])) {
```
umulmrum: Please do never use `empty` or the `!` operator (see https://docs.chameleon-system.de/contributing/best_practices.html#php-quirks) - can be replaced with empty string comparison here.

src/CoreBundle/Resources/public/javascript/tableEditor.js
```diff
@@ -812,7 +808,19 @@ function SetChangedDataMessage() {
     CHAMELEON.CORE.MTTableEditor.initInputChangeObservation();
 }
 
+CHAMELEON.CORE.MTTableEditor.initHelpTexts = function () {
+    $(".help-text-button").click(function () {
+        var helpTextId = '#tooltip-' + $(this).attr("data-helptextId");
```
umulmrum: The term `tooltip` is used multiple times, but I think this is too generic (and the help text isn't even a tooltip). Can we replace `tooltip` with `helptext` or `help-text`?

src/CoreBundle/private/modules/MTTableEditor/views/includes/fields.inc.php
ckleis: oh, I saw this in many lines of the same file.
I changed it.

src/CoreBundle/Resources/public/javascript/tableEditor.js
ckleis: changed in helptext
I have also added this functionality to fields across the entire width.

umulmrum:
Thanks @ckleis

PR #163, opened 2019-01-25
282 use correct remote field

UlrichKu:
| Q             | A
| ------------- | ---
| Branch        | 6.2.x
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed issues  | chameleon-system/chameleon-system#282
| License       | MIT

Based on https://github.com/chameleon-system/chameleon-base/pull/161

### DIFF
--- a/src/CoreBundle/private/library/classes/TCMSFields/TCMSFieldLookup.class.php
+++ b/src/CoreBundle/private/library/classes/TCMSFields/TCMSFieldLookup.class.php
@@ -564,7 +564,7 @@ class TCMSFieldLookup extends TCMSField
             return '';
         }
         $tableName = $this->GetConnectedTableName();
-        $identifyingColumnName = 'name';
+        $identifyingColumnName = $this->oDefinition->_oTableConf->GetNameColumn();;
 
         if (false === TCMSTableConf::FieldExists($tableName, $identifyingColumnName)) {
             $identifyingColumnName = 'id';
@@ -578,7 +578,7 @@ class TCMSFieldLookup extends TCMSField
         } else {
             $idList = $databaseConnection->quote($this->data);
         }
-        $sQuery = "SELECT `id`, $quotedIdentifierColumn FROM $quotedTableName WHERE `id` in ($idList) ORDER BY $quotedIdentifierColumn";
+        $sQuery = "SELECT $quotedIdentifierColumn FROM $quotedTableName WHERE `id` in ($idList) ORDER BY $quotedIdentifierColumn";
 
         $result = $databaseConnection->query($sQuery);
         if (!$result) {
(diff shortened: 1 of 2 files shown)


src/CoreBundle/private/library/classes/TCMSFields/TCMSFieldLookup.class.php
```diff
@@ -564,17 +564,21 @@ public function toString()
             return '';
         }
         $tableName = $this->GetConnectedTableName();
-        $nameColumn = 'name';
+        $identifyingColumnName = 'name';
+
+        if (false === TCMSTableConf::FieldExists($tableName, $identifyingColumnName)) {
+            $identifyingColumnName = 'id';
+        }
```
UlrichKu: else: probably the id should / must be present anyway - otherwise entries cannot be really safely identified

src/CoreBundle/private/library/classes/TCMSFields/TCMSFieldLookup.class.php
```diff
@@ -564,17 +564,21 @@ public function toString()
             return '';
         }
         $tableName = $this->GetConnectedTableName();
-        $nameColumn = 'name';
+        $identifyingColumnName = 'name';
```
umulmrum: Every table should have a name column, it just doesn't suffice to hardcode `name` here. This should work (didn't test it though):

```php
$tableConf = new TCMSTableConf(TTools::GetCMSTableId($tableName));
$nameColumn = $tableConf->GetNameColumn();
```

Then we wouldn't need the fallback on the non-human-readable ID.

src/CoreBundle/private/library/classes/TCMSFields/TCMSFieldLookup.class.php
umulmrum: Do we really need to identify records uniquely here? The method is just for printing a readable representation of the field value.

In the query we select the ID field but don't use it, so I think we should drop it from the query.

src/ViewRendererBundle/Twig/Loader/ChameleonTwigLoader.php
```diff
@@ -39,10 +39,10 @@ private function init()
         $this->initialized = true;
     }
 
-    protected function findTemplate($name)
+    protected function findTemplate($name, $throw = true)
```
umulmrum: Could you rebase the branch onto 6.2.x? I think it's better not to mix this one in.

src/CoreBundle/private/library/classes/TCMSFields/TCMSFieldLookup.class.php
UlrichKu: Check removed.

src/CoreBundle/private/library/classes/TCMSFields/TCMSFieldLookup.class.php
UlrichKu: Using $this->oDefinition->_oTableConf->GetNameColumn() now.

src/ViewRendererBundle/Twig/Loader/ChameleonTwigLoader.php
UlrichKu: These "changes" are now removed - but still show up here...

PR #164, opened 2019-01-25
ref #283: Check for export directory

UlrichKu:
| Q             | A
| ------------- | ---
| Branch        | 6.2.x
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed issues  | chameleon-system/chameleon-system#283
| License       | MIT

### DIFF
--- a/src/GenericTableExportBundle/objects/db/TPkgGenericTableExport.class.php
+++ b/src/GenericTableExportBundle/objects/db/TPkgGenericTableExport.class.php
@@ -47,19 +47,22 @@ class TPkgGenericTableExport extends TPkgGenericTableExportAutoParent
             $sOutput = utf8_decode($sOutput);
         }
 
-        $sFilePath = $this->getExportFilePath($sId);
+        $this->createExportDirectoryIfNeeded();
+
+        $exportFilePath = $this->getExportFilePath($sId);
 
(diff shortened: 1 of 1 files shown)


src/GenericTableExportBundle/objects/db/TPkgGenericTableExport.class.php
```diff
@@ -300,4 +302,22 @@ public static function GetInstanceFromSystemName($sSystemName)
...
+    }
+
+    private function createExportDirectoryIfNeeded(): void
+    {
+        $dir = $this->getExportDirectoryPath();
+
+        if (false === \is_dir($dir)) {
+            if (false === \mkdir($dir, 0777, true) && false === \is_dir($dir)) {
+
+                // TODO what kind of exception?
```
umulmrum: I think we can omit the check if mkdir succeeded. If there's an error, `WriteExportToFile` will trigger a warning anyway (and we can't throw an exception from `WriteExportToFile` because of BC).

UlrichKu: both error handlings / todo removed.

umulmrum:
Thanks @UlrichKu

PR #165, opened 2019-01-28
Feature/99 cleanup

umulmrum:
| Q             | A
| ------------- | ---
| Branch        | master
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| License       | MIT

This PR proposes some corrections and improvements encountered while working on other issues.

### DIFF
--- a/src/CoreBundle/private/modules/CMSUpdateManager/CMSUpdateManager.class.php
+++ b/src/CoreBundle/private/modules/CMSUpdateManager/CMSUpdateManager.class.php
@@ -59,9 +59,16 @@ class CMSUpdateManager extends TModelBase
     {
         $aIncludes = parent::GetHtmlHeadIncludes();
         $aIncludes[] = '<link href="'.TGlobal::GetPathTheme().'/css/modules/updateManager.css" rel="stylesheet" type="text/css" />';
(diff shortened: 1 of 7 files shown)


src/CoreBundle/private/modules/CMSUpdateManager/CMSUpdateManager.class.php
```diff
@@ -59,11 +59,18 @@ public function GetHtmlHeadIncludes()
     {
         $aIncludes = parent::GetHtmlHeadIncludes();
         $aIncludes[] = '<link href="'.TGlobal::GetPathTheme().'/css/modules/updateManager.css" rel="stylesheet" type="text/css" />';
-        $aIncludes[] = '<script src="'.TGlobal::GetStaticURLToWebLib('/javascript/modules/updateManager/updateManager.js').'" type="text/javascript"></script>';
 
         return $aIncludes;
     }
 
+    public function GetHtmlFooterIncludes()
```
UlrichKu: Looks reasonable.

But is there maybe a special reason (here) why this change is needed?

umulmrum: The CHAMELEON variable is now defined only in the footer and is required by the update manager, so the change is needed, yes.

PR #169, opened 2019-01-31
ref #290: Use a sub-select for the portal visibility restriction

UlrichKu:
| Q             | A
| ------------- | ---
| Branch        | 6.2.x
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed issues  | chameleon-system/chameleon-system#290
| License       | MIT

Maybe needed to document the (slightly) different behavior of \TCMSListManagerEndPoint::FilterQuery() now: does not call GetPortalRestrictionJoin() itself anymore?

### DIFF
--- a/src/CoreBundle/private/library/classes/TCMSListManager/TCMSListManagerEndPoint.class.php
+++ b/src/CoreBundle/private/library/classes/TCMSListManager/TCMSListManagerEndPoint.class.php
@@ -129,10 +129,6 @@ class TCMSListManagerEndPoint
         if (is_array($this->oTableConf->sqlData) && array_key_exists('list_query', $this->oTableConf->sqlData)) {
             $recordquery = trim($this->oTableConf->sqlData['list_query']);
             $portalRestriction = $this->GetPortalRestriction();
-            $portalRestrictionJoin = '';
-            if (!empty($portalRestriction)) {
-                $portalRestrictionJoin = $this->GetPortalRestrictionJoin();
-            }
             $userRestriction = $this->GetUserRestriction();
             $userRestrictionJoin = '';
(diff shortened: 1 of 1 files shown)


src/CoreBundle/private/library/classes/TCMSListManager/TCMSListManagerEndPoint.class.php
```diff
@@ -301,7 +302,12 @@ public function GetPortalRestriction()
                 } else {
                     if ($portalMLTFieldExists) {
                         // mlt connection (record may be in many portals)
-                        $restriction .= " ($quotedMltTable.`target_id` IS NULL)";
+
+                        $mltSubSelect = "SELECT DISTINCT $quotedTableName.`id` FROM $quotedTableName";
```
umulmrum: Can we move this code to a separate method (--> to simplify this method, avoid code duplication [these lines are almost identical to 289-294], and explain the purpose by method name)?

UlrichKu: I was thinking about it. But at least not easily:

It has a number of parameters (4?) and the difference between the two blocks would need another for a non-trivial if in that method.

umulmrum: If we omit the `$quotedMltTable.target_id IN ($portalRestriction)` part, constructing $mltSubSelect is identical. So we can at least easily factor this path out, and the method only needs 3 arguments.

UlrichKu: done

umulmrum:
Thanks @UlrichKu

PR #180, opened 2019-02-07
Feature/99 use form container in table editor

pixeljunkie:
| Q             | A
| ------------- | ---
| Branch        | master
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| License       | MIT

### DIFF
--- a/src/CoreBundle/Resources/public/javascript/tableEditor.js
+++ b/src/CoreBundle/Resources/public/javascript/tableEditor.js
@@ -539,11 +539,13 @@ function SaveViaAjax() {
 function SaveViaAjaxCallback(data, statusText) {
     CloseModalIFrameDialog();
 
-    // remove all message-classes from the fields
-    $('*[id^="fieldname_"]').removeClass();
+    // remove all message background classes from field containers.
+    $('*[id^="fieldname_"]').removeClass(function (index, className) {
+        return (className.match (/(^|\s)bg-\S+/g) || []).join(' ');
+    });
 
     var returnVal = false;
-    if (data != false && data != null) {
+    if (data !== false && data != null) {
 
         returnVal = true;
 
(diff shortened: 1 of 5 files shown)


src/CoreBundle/private/modules/MTTableEditor/views/includes/fields.inc.php
```diff
@@ -35,16 +34,18 @@
                     $oField->oDefinition->sqlData['modifier'] = 'readonly';
                 }
 
-                $sTmpFormTabsContent .= '<tr class="row m-0 '.$rowClass.'">
-            <th valign="top" class="leftTD col-2"'.$rowColorStyle.'>
-              <div id="fieldname_'.TGlobal::OutHTML($oField->name).'">
+
+                // todo $rowColorStyle
```
umulmrum: Is this a TODO we can already work on? Ideally we don't have TODOs left in production code.

src/CoreBundle/Resources/public/javascript/tableEditor.js
```diff
@@ -539,11 +539,13 @@ function SaveViaAjax() {
...
-    // remove all message-classes from the fields
-    $('*[id^="fieldname_"]').removeClass();
+    // remove all message background classes from field containers.
+    $('*[id^="fieldname_"]').removeClass(function (index, className) {
+        return (className.match (/(^|\s)bg-\S+/g) || []).join(' ');
+    });
 
     var returnVal = false;
-    if (data != false && data != null) {
+    if (data !== false && data != null) {
```
umulmrum: Are you confident enough that this change won't cause problems? The true/false rules in JS are at least as wonderful and straightforward as in PHP.

src/CoreBundle/private/modules/MTTableEditor/views/includes/fields.inc.php
pixeljunkie: oooh... missed that and found out i missed the handling of the separator field type as well. Now included.

src/CoreBundle/Resources/public/javascript/tableEditor.js
pixeljunkie: tested this. works. So, yes.

umulmrum:
Thanks @pixeljunkie

PR #186, opened 2019-02-07
Feature/99 open cms help in overlay

pixeljunkie:
| Q             | A
| ------------- | ---
| Branch        | master
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| License       | MIT

Includes styling the help module and a handling iframe mode or standalone mode. Reworked the module but did not replace all of the deprecated calls.


src/CoreBundle/private/modules/MTHeader/views/standard.view.php
```diff
@@ -253,13 +251,13 @@ class="nav-link dropdown-toggle <?= $userButtonStyle ?>"
                             </span>
                         </a>
                         <div class="dropdown-menu dropdown-menu-right">
-                            <a class="dropdown-item" href="<?= PATH_CMS_CONTROLLER ?>?pagedef=tableeditor&tableid=<?= $data['iTableIDCMSUser'] ?>&id=<?= $data['oUser']->id ?>&<?= urlencode('module_fnc[contentmodule]') ?>">
+                            <a class="dropdown-item" href="<?= PATH_CMS_CONTROLLER; ?>?pagedef=tableeditor&tableid=<?= $data['iTableIDCMSUser']; ?>&id=<?= $data['oUser']->id; ?>&<?= urlencode('module_fnc[contentmodule]'); ?>">
```
umulmrum: Those semicolons aren't required and are omitted by most projects, and I would omit them, too (in fact I removed some in the last few days).

src/CoreBundle/private/modules/CMSModuleHelp/CMSModuleHelp.class.php
```diff
@@ -25,23 +29,37 @@ class CMSModuleHelp extends TCMSModelBase
...
-    {
-    }
-
     public function &Execute()
     {
         $this->data = parent::Execute();
 
+        $translator = $this->getTranslator();
+
+        $this->data['translator'] = $translator;
```
umulmrum: I'd advocate against passing service dependencies to views as it increases coupling between module and view.

src/CoreBundle/private/modules/MTHeader/views/standard.view.php
pixeljunkie: CS-Fixer did this. Without changing the code style config i don`t get rid of them.

src/CoreBundle/private/modules/CMSModuleHelp/CMSModuleHelp.class.php
pixeljunkie: Right. Now setting the translated title in the module instead.

src/CoreBundle/private/modules/MTHeader/views/standard.view.php
umulmrum: Looks like I did it wrong :-) - OK, I don't like the semicolons in this case, but we stick with the CS-Fixer defaults.

umulmrum:
OK, now only the question "target blank vs. overlay" is open.

pixeljunkie:
Blank vs overlay? I managed to allow both. default = overlay. CTRL click = blank. I think this is the best solution. The links in the help are targeted with top, so the links work inside the overlay.

umulmrum:
OK rechecked - looks good to me.

But there are merge conflicts. Could you resolve?

pixeljunkie:
resolved merge conflict

umulmrum:
Thanks @pixeljunkie

PR #187, opened 2019-02-08
Fix/178 replace glyph icons

pixeljunkie:
| Q             | A
| ------------- | ---
| Branch        | master
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | yes
| Tests pass?   | yes
| Fixed issues  | chameleon-system/chameleon-system#178
| License       | MIT


src/CoreBundle/Resources/public/javascript/pnotify-3.2.0/pnotify.custom.min.js
```diff
@@ -0,0 +1,22 @@
+/*
+PNotify 3.2.0 sciactive.com/pnotify/
+(C) 2015 Hunter Perrin; Google, Inc.
+license Apache-2.0
```
umulmrum: The older version of this lib was licensed under GPL/LGPL/MPL but was by accident listed under MIT in THIRD-PARTY-LICENSE. Could you:
a) move the old entry to the MPL section (we want to use this license) and mark it with e.g. "up to version 2.0.1".
b) add a separate entry in the Apache License section.

pixeljunkie: done

umulmrum:
Thanks @pixeljunkie

PR #191, opened 2019-02-11
Feature/301 navigation tree styling

ckleis:
| Q             | A
| ------------- | ---
| Branch        | master 
| Bug fix?      | yes
| New feature?  | yes
| BC breaks?    | no 
| Deprecations? | no 
| Fixed issues  | chameleon-system/chameleon-system#301   
| License       | MIT

Styling Navigation Tree:
The navigation tree is called in two places. First from the header area and second when a navigation point is assigned to a page. In the second case in an iFrame. This has to be differentiated.


umulmrum:
Thanks @ckleis

PR #193, opened 2019-02-11
Fix/306 TCMSFieldDateTime with seconds

ckleis:
| Q             | A
| ------------- | ---
| Branch        | master
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Fixed issues  | chameleon-system/chameleon-system#306
| License       | MIT

First, the TCMSFieldDateTime field with the datetimepicker() widget did mysterious things because the date field in the database had seconds and datetimepicker() by default doesn't represent seconds. After I found the country-specific date and time setting WITH seconds (format: "L LTS") for the datetimepicker() widget, I had an ugly styling error:

selection_267

Unfortunately this could not be investigated, because it immediately collapsed again when leaving the field. Even with devTools this could not be prevented.

After some searching I found the options keepOpen: true and debug: true for the datetimepicker(). This allowed me to find and overwrite the wrong css setting.

### DIFF
--- a/src/CoreBundle/Resources/public/themes/standard/css/tableeditcontainer.css
+++ b/src/CoreBundle/Resources/public/themes/standard/css/tableeditcontainer.css
@@ -224,8 +224,8 @@ label.switch .switch-slider::after {
     transition: inherit;
 }
 
-/**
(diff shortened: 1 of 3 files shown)


umulmrum:
Thanks @ckleis

PR #195, opened 2019-02-12
ref #308: products: field "is variant of": styling

ckleis:
| Q             | A
| ------------- | ---
| Branch        | master
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no 
| Deprecations? | no
| Fixed issues  | chameleon-system/chameleon-system#308  
| License       | MIT

Products => field "Is a variant of": correction of a small styling error.

### DIFF
--- a/src/CoreBundle/private/library/classes/TCMSFields/TCMSFieldLookupParentID.class.php
+++ b/src/CoreBundle/private/library/classes/TCMSFields/TCMSFieldLookupParentID.class.php
@@ -45,19 +45,21 @@ class TCMSFieldLookupParentID extends TCMSFieldLookup
         $item->Load($this->data);
         $showLinkToParentRecord = $this->oDefinition->GetFieldtypeConfigKey('bShowLinkToParentRecord');
 
-        $html = $this->_GetHiddenField().$item->GetName();
+        $html = $this->_GetHiddenField();
+        $parentName = $item->GetName();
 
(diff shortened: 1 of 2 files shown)


pixeljunkie:
Please rework the styling with an active variant name left to the button.

src/CoreBundle/private/library/classes/TCMSFields/TCMSFieldLookupParentID.class.php
```diff
@@ -45,19 +45,21 @@ protected function getLinkToParentRecordIfSet()
         $item->Load($this->data);
         $showLinkToParentRecord = $this->oDefinition->GetFieldtypeConfigKey('bShowLinkToParentRecord');
 
-        $html = $this->_GetHiddenField().$item->GetName();
+        $html = $this->_GetHiddenField();
+        $parentName = $item->GetName();
```
umulmrum: What do you think of naming the variable `$itemName`? This would be in line with `$item`.

Also I'd move the declaration directly before its usage, so it's easier to relate declaration and usage (and there's also a really slight performance gain when we don't enter the if block).

ckleis: OK, done

umulmrum:
Thanks @ckleis

PR #199, opened 2019-02-12
Fix/245 cleanup

umulmrum:
| Q             | A
| ------------- | ---
| Branch        | master
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed issues  | chameleon-system/chameleon-system#245
| License       | MIT

### DIFF
--- a/src/CoreBundle/Util/FieldTranslationUtil.php
+++ b/src/CoreBundle/Util/FieldTranslationUtil.php
@@ -222,7 +222,9 @@ class FieldTranslationUtil
 
         $translatableString = $translatable ? '1' : '0';
         $tableEditorManager = TTools::GetTableEditorManager('cms_field_conf', $fieldId);
+        $tableEditorManager->AllowEditByAll(true);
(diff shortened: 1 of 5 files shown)


src/CoreBundle/Util/FieldTranslationUtil.php
```diff
@@ -222,7 +222,9 @@ private function changeTranslatableState($fieldId, $translatable)
 
         $translatableString = $translatable ? '1' : '0';
         $tableEditorManager = TTools::GetTableEditorManager('cms_field_conf', $fieldId);
+        $tableEditorManager->AllowEditByAll(true);
```
UlrichKu: This wasn't a problem before?

umulmrum: It was, or at least I can't imagine that it wasn't a problem. So maybe there a systems where fields were not configured translatable.

PR #100, opened 2018-11-28
ref #216: add missing page number value to ajax data

pixeljunkie:
| Q             | A
| ------------- | ---
| Branch        | 6.2.x
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed issues  | chameleon-system/chameleon-system#216
| License       | MIT

Adds the mising pageNumber to ajax call for image list rendering.

PR #200, opened 2019-02-12
ref #100: Display sidebar menu categories as dropdowns

umulmrum:
| Q             | A
| ------------- | ---
| Branch        | master
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| License       | MIT

Displays sidebar menu items as dropdowns so that the minimized view makes more sense. This required adding icons to the menu categories.

When minimized, the items from the lower dropdowns will be cropped by the lower screen border, but I didn't spend that much time in fixing it. We can live with it for some time, and maybe someone else finds a solution :-)

### DIFF
--- a/src/CoreBundle/Bridge/Chameleon/Migration/Script/update-1549281421.inc.php
+++ b/src/CoreBundle/Bridge/Chameleon/Migration/Script/update-1549281421.inc.php
@@ -219,7 +219,52 @@ TCMSLogChange::update(__LINE__, $data);
 $query = 'ALTER TABLE `cms_menu_category` ADD INDEX ( `position` )';
 TCMSLogChange::RunQuery(__LINE__, $query);
 
-// Add field cms_menu_item
+// Add icon_font_css_class field.
+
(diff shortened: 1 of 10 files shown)


src/CoreBundle/Bridge/Chameleon/Migration/Script/update-1549281421.inc.php
```diff
@@ -219,7 +219,52 @@
...
+        'fieldclass_subtype' => '',
+        'class_type' => 'Core',
+        'modifier' => 'none',
+        'field_default_value' => '',
+        'length_set' => '',
+        'fieldtype_config' => '',
+        'restrict_to_groups' => '0',
+        'field_width' => '',
+        'position' => '0',
+        '049_helptext' => '',
```
UlrichKu: Some short explanations? (on when to use for example)

umulmrum: - Added help text.
- Improved help texts for other icon class fields.
- Added German translation where missing.

PR #202, opened 2019-02-14
ref #312: Add welcome screen and set as default pagedef; add configur…

umulmrum:
…ation for main pagedef

| Q             | A
| ------------- | ---
| Branch        | master
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed issues  | chameleon-system/chameleon-system#312
| License       | MIT

This PR handles the technical part of the issue. Adding useful text will be done separately. // cc @pixeljunkie

### DIFF
--- a/UPGRADE-6.3.md
+++ b/UPGRADE-6.3.md
@@ -392,6 +392,12 @@ From a technical point no changes are required if the project is a shop system. 
 should consider removing menu categories that are nevertheless created during migration. Note that there will be some
 error messages during migration that complain about missing shop tables and modules in this case, which can be ignored. 
 
+## Home Page Changes
+
+As the main menu is now displayed in the sidebar, the classic main menu was replaced by a welcome screen (which we plan
+to replace by a dashboard in a future Chameleon release). If users would like to keep the classic main menu yet, it can
+be restored by setting the config value `chameleon_system_core: backend: main_pagedef: 'main'` in `config.yml`.
+
 # Deprecated Code Entities
 
 It is recommended that all references to the classes, interfaces, properties, constants, methods and services in the
--- a/UPGRADE-6.3.md
+++ b/UPGRADE-6.3.md
@@ -396,7 +396,7 @@ error messages during migration that complain about missing shop tables and modu
 
 As the main menu is now displayed in the sidebar, the classic main menu was replaced by a welcome screen (which we plan
 to replace by a dashboard in a future Chameleon release). If users would like to keep the classic main menu yet, it can
-be restored by setting the config value `chameleon_system_core: backend: main_pagedef: 'main'` in `config.yml`.
+be restored by setting the config value `chameleon_system_core: backend: home_pagedef: 'main'` in `config.yml`.
 
 # Deprecated Code Entities
 
--- a/src/CoreBundle/Bridge/Chameleon/Migration/Script/update-1550132341.inc.php
+++ b/src/CoreBundle/Bridge/Chameleon/Migration/Script/update-1550132341.inc.php
@@ -1,0 +1,38 @@
+<h1>Build #1550132341</h1>
+<h2>Date: 2019-02-14</h2>
+<div class="changelog">
+    - Add StaticView module
+</div>
+<?php
+
+$moduleId = TCMSLogChange::createUnusedRecordId('cms_tpl_module');
+
+$data = TCMSLogChange::createMigrationQueryData('cms_tpl_module', 'en')
+  ->setFields([
+      'description' => 'This module can be used to display static information in the backend.',
+      'icon_list' => 'application.png',
+      'view_mapper_config' => 'standard=StaticView/standard.html.twig',
+      'mapper_chain' => '',
+      'view_mapping' => '',
+      'revision_management_active' => '0',
+      'is_copy_allowed' => '0',
+      'show_in_template_engine' => '0',
+      'position' => '',
+      'is_restricted' => '0',
+      'name' => 'Static view',
+      'classname' => 'chameleon_system_core.module.static_view_module',
+      'id' => $moduleId,
+  ])
+;
+TCMSLogChange::insert(__LINE__, $data);
+
+$data = TCMSLogChange::createMigrationQueryData('cms_tpl_module', 'de')
+    ->setFields([
+        'description' => 'Mit diesem Modul können im Backend statische Informationen angezeigt werden.',
+        'name' => 'Statische Ansicht',
+    ])
+    ->setWhereEquals([
+        'id' => $moduleId,
+    ])
+;
+TCMSLogChange::update(__LINE__, $data);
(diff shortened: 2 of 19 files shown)


UPGRADE-6.3.md
```diff
@@ -392,6 +392,12 @@ From a technical point no changes are required if the project is a shop system.
 should consider removing menu categories that are nevertheless created during migration. Note that there will be some
 error messages during migration that complain about missing shop tables and modules in this case, which can be ignored. 
 
+## Home Page Changes
+
+As the main menu is now displayed in the sidebar, the classic main menu was replaced by a welcome screen (which we plan
+to replace by a dashboard in a future Chameleon release). If users would like to keep the classic main menu yet, it can
+be restored by setting the config value `chameleon_system_core: backend: main_pagedef: 'main'` in `config.yml`.
```
UlrichKu: Is "main_page" and "main" sufficiently clear here?

At the moment I don't know what this is (or where).

src/CoreBundle/DependencyInjection/Configuration.php
```diff
@@ -154,7 +155,24 @@ private function getModuleExecutionConfig()
...
+     */
+    private function getBackendConfig()
+    {
+        $tree = new TreeBuilder();
+        $subTree = $tree->root('backend');
+        $subTree->addDefaultsIfNotSet();
+        $subTree->children()
+            ->scalarNode('main_pagedef')
+                ->defaultValue('welcome')
+                ->info('The pagedef that is displayed after login and when clicking any "to main" button.')
```
UlrichKu: (for my understanding: this is then what is normally called "home page"?)

src/CoreBundle/Bridge/Chameleon/Migration/Script/update-1550132341.inc.php
```diff
@@ -0,0 +1,38 @@
...
+      'is_copy_allowed' => '0',
+      'show_in_template_engine' => '0',
+      'position' => '',
+      'is_restricted' => '0',
+      'name' => 'Static view',
+      'classname' => 'chameleon_system_core.module.static_view_module',
+      'id' => $moduleId,
+  ])
+;
+TCMSLogChange::insert(__LINE__, $data);
```
UlrichKu: Is this module used here?

Or it will be used for the "frontend / text" part?

UPGRADE-6.3.md
umulmrum: I suppose it should be clear enough once the reader knows how pagedefs work in the backend (which should also be documented). Otherwise they should just follow this instruction - I think in this place we can't add meaningful explanation without also explaining the whole pagedef concept. Or do you have something in mind?

src/CoreBundle/DependencyInjection/Configuration.php
umulmrum: Yes. I renamed it to home_pagedef to use the term "home" consistently.

src/CoreBundle/Bridge/Chameleon/Migration/Script/update-1550132341.inc.php
umulmrum: Not sure if I understand correctly. It is used statically in welcome.pagedef.php.

umulmrum:
Also added the StaticViewModule I forgot before :-)

src/CoreBundle/DependencyInjection/Configuration.php
UlrichKu: home_page_def?

src/CoreBundle/Bridge/Chameleon/Module/StaticView/StaticViewModule.php
```diff
@@ -0,0 +1,42 @@
...
+    {
+        parent::__construct();
+        $this->languageService = $languageService;
+    }
+
+    public function Accept(\IMapperVisitorRestricted $visitor, $cachingEnabled, \IMapperCacheTriggerRestricted $cacheTriggerManager)
+    {
+        $targetView = $this->aModuleConfig['targetView'];
+        $targetLanguage = $this->languageService->getActiveLanguage()->fieldIso6391;
+        $targetView = \str_replace('[{language}]', $targetLanguage, $targetView);
```
UlrichKu: That is necessary due to legacy reasons?

Or is this a current technique for the backend?

UPGRADE-6.3.md
UlrichKu: Parts of it I understood now: the "home page" is the main area which used to show the 3-column menu.

What is the default here (the other possibilities instead of "main")?
main has here the meaning of "classic menu"?

This is actually a pagedef?

If I specify "main" here what will happen to the side bar menu?

umulmrum: Yes, "main" is the pagedef that contains the classic main menu. Default value is the new "welcome" pagedef; this can be viewed with `app/console config:dump-reference ChameleonSystemCoreBundle`, so I'm against duplicating this piece of documentation. In principle any backend pagedef can be used, so there isn't a complete list (some pagedefs wouldn't work, as they require GET parameters, e.g. the table editor needs a table ID; I'm quite sure the user doesn't want to configure any arbitrary pagedef but is aware if a pagedef can be used without further configuration, so I think it's OK as is).

The sidebar is also visible on the `main` page, so there are two menus (which is the reason for this PR). The sidebar should be visible on any page, especially as it will be the only menu from a future Chameleon version on and users should get used to it.

src/CoreBundle/DependencyInjection/Configuration.php
umulmrum: The `pagedef` notation is used throughout the code base, I wouldn't write it in another way here.

src/CoreBundle/Bridge/Chameleon/Module/StaticView/StaticViewModule.php
umulmrum: Do you mean the bracket notation? It is the established syntax for replacing placeholders, so I wouldn't introduce a change here (the only other candidate is Twig's `%...%` notation, but this isn't a template).

UlrichKu: No: that there might be multiple file version according to language.

And not for example a translation mechanism that prints the relevant translations - like with frontend twig (trans filter).

UPGRADE-6.3.md
UlrichKu: ok.

Then only two details still:
Shouldn't the view be named "classic_menu" or similar?

And one could document that any view can be specified here besides "main" if desired.

umulmrum: I'd leave it unchanged as it is "deprecated" and as some users will even be more familiar with `main`. Most users won't need this anymore, others only need to perform this one instruction - even if it isn't perfectly clear why it's called `main`, it will just work and doesn't need further thoughts by users.

src/CoreBundle/Bridge/Chameleon/Module/StaticView/StaticViewModule.php
umulmrum: There isn't any mechanism like this yet. The welcome texts will be complete views and therefore have multiple lines and also contain HTML, so this was really cumbersome if we tried to handle it via translation.

PR #203, opened 2019-02-14
Minor cleanup

umulmrum:
| Q             | A
| ------------- | ---
| Branch        | master
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed issues  | -
| License       | MIT

### DIFF
--- a/src/CoreBundle/private/library/classes/MenuItems/TCMSMenuItem_Module.class.php
+++ b/src/CoreBundle/private/library/classes/MenuItems/TCMSMenuItem_Module.class.php
@@ -29,9 +29,6 @@ class TCMSMenuItem_Module extends TCMSMenuItem
         if ('' !== $pagedefType) {
             $url .= '&_pagedefType='.$pagedefType;
         }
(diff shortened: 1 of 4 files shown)


src/CoreBundle/private/library/classes/MenuItems/TCMSMenuItem_Module.class.php
```diff
@@ -29,9 +29,6 @@ public function GetLink()
         if ('' !== $pagedefType) {
             $url .= '&_pagedefType='.$pagedefType;
         }
-        if ('1' == $this->data['show_as_popup']) {
```
UlrichKu: This has been doing what?

pixeljunkie: Did open modules in an overlay. Now all modules are opened as full page, with the default browser option to open in a new tab while holding CTRL key.

PR #210, opened 2019-02-18
ref #100: Fix cache clear URL (did no longer work because of changes …

umulmrum:
…in breadcrumb; dependency on breadcrumb isn't useful)

| Q             | A
| ------------- | ---
| Branch        | master
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| License       | MIT

### DIFF
--- a/src/CoreBundle/private/modules/MTHeader/MTHeader.class.php
+++ b/src/CoreBundle/private/modules/MTHeader/MTHeader.class.php
@@ -118,11 +118,7 @@ class MTHeader extends TCMSModelBase
             if (TGlobal::CMSUserDefined()) {
                 $this->data['breadcrumb'] = $this->global->GetURLHistory()->GetBreadcrumb(true);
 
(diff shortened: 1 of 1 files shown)


src/CoreBundle/private/modules/MTHeader/MTHeader.class.php
```diff
@@ -118,11 +118,7 @@ public function &Execute()
             if (TGlobal::CMSUserDefined()) {
                 $this->data['breadcrumb'] = $this->global->GetURLHistory()->GetBreadcrumb(true);
 
-                $lastHistNode = end($this->data['breadcrumb']);
```
UlrichKu: the "last hist node" is obsolete anyway?

umulmrum: It was only used for the following - now replaced - lines.

PR #217, opened 2019-02-25
Feature/100 sidebar save active category

umulmrum:
| Q             | A
| ------------- | ---
| Branch        | master
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| License       | MIT

Saving and restoring the active category as suggested in #100. To have an "active" category, we now only open one category at a time and close others.


src/CoreBundle/Bridge/Chameleon/Module/Sidebar/SidebarBackendModule.php
```diff
@@ -127,13 +146,37 @@ private function getMenuItems(): array
...
     }
 
+    private function getActiveCategory(): ?string
+    {
+        $session = $this->requestStack->getCurrentRequest()->getSession();
+
+        return $session->get(self::ACTIVE_CATEGORY_SESSION_KEY);
+    }
+
+    public function saveActiveCategory(): void
```
UlrichKu: Would protected be sufficient here?

src/CoreBundle/Resources/public/javascript/modules/sidebar/sidebar.js
```diff
@@ -42,6 +43,21 @@
...
+
+            let categoryId = null;
+            // The following condition is inverted, as this handler will be executed before the actual class change.
+            if (false === $category.hasClass('open')) {
+                categoryId = $category.data('categoryid');
+            }
+
+            const url = this.$baseElement.data('save-active-category-notification-url');
+            $.post(url, {
+                categoryId: categoryId
```
UlrichKu: errors are not handled here?

src/CoreBundle/Bridge/Chameleon/Module/Sidebar/SidebarBackendModule.php
umulmrum: Ah yes, changed to `protected`.

src/CoreBundle/Resources/public/javascript/modules/sidebar/sidebar.js
umulmrum: We can't really do anything in case an error occurs. The sidebar will still work on failure (only the state won't be saved), so it's better not to bother the user,

UlrichKu: You could log an error to the console.

umulmrum: Logging to console in production code is considered bad practice. I'd just leave it as is: In case of an error the sidebar will still mostly work, and in development the error can easily be seen as failed request in the debug toolbar.

PR #220, opened 2019-02-26
WIP Feature/332 coreui sources

umulmrum:
| Q             | A
| ------------- | ---
| Branch        | master
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed issues  | chameleon-system/chameleon-system#332
| License       | MIT

Add CoreUI sources and a build guide; also update JS to 2.1.6 (we updated CSS to 2.1.6 previously).


src/CoreBundle/Resources/doc/backend-assets.rst
```diff
@@ -0,0 +1,20 @@
...
+CoreUI Theme
+------------
+
+The backend uses the CoreUI theme. To make changes e.g. to SCSS variables, proceed as follows:
+
+- Install Node.js and NPM on your machine.
+- Sources are located under `src/CoreBundle/Resources/asset-source/coreui-2.1.6`. Make changes there.
+- In a terminal, go to `src/CoreBundle/Resources/asset-source/coreui-2.1.6`.
+- Run `npm install` (ignore dependency warnings - we manage dependencies separately).
+- Run `npm run dist`.
```
UlrichKu: (Maybe here or for future dependencies: I have provided a shell script for similar installation steps which does these steps: call run (= compile), copy all relevant files to relevant directories.)

umulmrum: I thought about adding a build script but decided against it, for two reasons:
1. I didn't want to make assumptions on the developer's OS, and so didn't want to just build a bash script.
2. Also I didn't want to write a simplistic script that contains just the commands without proper guidance and error handling (leading to a cascade of errors when the first command fails).

Doing it "right" would cost too much time now - it should be quite a rare case to change the sources, so I think it's OK to do this manually.

umulmrum:
@pixeljunkie had concerns and wanted to think about requiring the sources instead of copying - please wait until we found common ground.

umulmrum:
There doesn't seem to be a built-in way to achieve variable overrides without changing the CoreUI sources (or I just didn't find it). Maybe we can create a custom package.json file to build sources?

pixeljunkie:
A custom package.json should do the job.

pixeljunkie:
There is now a build process for the coreUI sources. So this PR will be closed.

PR #222, opened 2019-02-26
ref #328: add check in handleFormAndLinkTargetsInModals method to che…

pixeljunkie:
…ck if the current page is inside an iframe, search dialog in top document and not inside the iframe itself

| Q             | A
| ------------- | ---
| Branch        | 6.2.x
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed issues  | chameleon-system/chameleon-system#328
| License       | MIT


umulmrum:
Thanks @pixeljunkie

PR #228, opened 2019-02-28
ref #342: Add permission for all languages and portals

umulmrum:
| Q             | A
| ------------- | ---
| Branch        | 6.2.x
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed issues  | chameleon-system/chameleon-system#342
| License       | MIT

### DIFF
--- a/src/DistributionBundle/Bootstrap/InitialBackendUserCreator.php
+++ b/src/DistributionBundle/Bootstrap/InitialBackendUserCreator.php
@@ -159,47 +159,34 @@ class InitialBackendUserCreator
             'show_as_rights_template' => '1',
         ]);
 
-        $allLanguages = $this->getAllLanguages();
+        $this->addLanguageLinks($userId);
+        $this->addPortalLinks($userId);
+    }
+
+    private function addLanguageLinks(string $userId): void
+    {
+        $languageList = \TdbCmsLanguageList::GetList();
         $position = 0;
-        foreach ($allLanguages as $languageId) {
+        while (false !== $language = $languageList->Next()) {
             $this->databaseConnection->insert('cms_user_cms_language_mlt', [
                 'source_id' => $userId,
-                'target_id' => $languageId,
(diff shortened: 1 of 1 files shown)


src/DistributionBundle/Bootstrap/InitialBackendUserCreator.php
```diff
@@ -159,11 +159,47 @@ private function addUser(string $userId, string $name, string $password): void
...
+                'entry_sort' => ++$position,
+            ]);
+        }
+    }
+
+    private function getAllLanguages(): array
+    {
+        $languages = [];
+        $languageList = \TdbCmsLanguageList::GetList();
+        while (false !== $languageObject = $languageList->Next()) {
```
UlrichKu: Why not loop these above directly?

(As these methods also do not get all languages but all ids.)

umulmrum: Yes, that was overly complex - changed.

src/DistributionBundle/Bootstrap/InitialBackendUserCreator.php
```diff
@@ -159,11 +159,34 @@ private function addUser(string $userId, string $name, string $password): void
             'show_as_rights_template' => '1',
         ]);
 
-        $this->databaseConnection->insert('cms_user_cms_language_mlt', [
-            'source_id' => $userId,
-            'target_id' => $languageEn->id,
-            'entry_sort' => 1,
-        ]);
+        $this->addLanguageLinks($userId);
```
UlrichKu: connectUserWithPortals? (or similar) "Links" is a bit arbitrary

umulmrum: If it makes you happy .. ;-)

PR #233, opened 2019-02-28
ref #204: correct date format in JS after format has changed to "L" or "L LTS"

ckleis:
| Q             | A
| ------------- | ---
| Branch        | master
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no 
| Deprecations? | no
| Fixed issues  | chameleon-system/chameleon-system#204
| License       | MIT

datetimepicker plugin:
With commit "ref #306: TCMSFieldDateTime with seconds" the format has changed from "" to "L LTS". This was not considered in the corresponding tableEditor.js yet and led to the error.


umulmrum:
Thanks @ckleis

PR #234, opened 2019-03-01
Fix/344 delete button not working

pixeljunkie:
| Q             | A
| ------------- | ---
| Branch        | master
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed issues  | chameleon-system/chameleon-system#344
| License       | MIT

### DIFF
--- a/src/CoreBundle/Resources/public/javascript/tableEditor.js
+++ b/src/CoreBundle/Resources/public/javascript/tableEditor.js
@@ -498,14 +498,17 @@ function SaveViaAjaxCustomCallback(customCallbackFunction, closeAfterSave) {
     document.cmseditform._fnc.value = 'AjaxSave';
 
     PostAjaxForm('cmseditform', eval(customCallbackFunction));
-    if (closeAfterSave == true && typeof parent != 'undefined') parent.setTimeout("parent.CloseModalIFrameDialog()", 3000);
+
+    if (closeAfterSave == true && typeof parent != 'undefined') {
+        parent.setTimeout("parent.CloseModalIFrameDialog()", 3000);
+    }
 }
 
 /*
  * tableEditor: save table single field via ajax
  */
 function SaveFieldViaAjaxCustomCallback(customCallbackFunction) {
-    if (customCallbackFunction == 'undefined') {
+    if ('undefined' === customCallbackFunction) {
         customCallbackFunction = SaveViaAjaxCallback;
     }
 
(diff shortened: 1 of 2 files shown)


src/CoreBundle/Resources/public/javascript/tableEditor.js
```diff
@@ -705,11 +704,10 @@ function DeleteRecord() {
 }
 
 CHAMELEON.CORE.MTTableEditor.DeleteRecordWithCustomConfirmMessage = function (sConfirmText) {
-    if (sCurrentRecordName != '') {
-        sConfirmText += "\n\n " + CHAMELEON.CORE.i18n.Translate('chameleon_system_core.js.record') + ": \"" + sCurrentRecordName + '"';
-    } else {
-        sConfirmText += "\n\n " + CHAMELEON.CORE.i18n.Translate('chameleon_system_core.js.record') + ": \"" + CHAMELEON.CORE.i18n.Translate('chameleon_system_core.js.unnamed') + '"';
-    }
+    var currentRecordName = $('#tableEditorContainer .navbar-brand').text();
```
UlrichKu: There is still a "sCurrentRecordName" here?

UlrichKu:
What is the actual fix here?

Or what was the problem

src/CoreBundle/Resources/public/javascript/tableEditor.js
pixeljunkie: sCurrentRecordName does not exist anymore. It was used to set the record name only for the delete method.
This was not necessary because we could get the name from the title of the record.
I added to update the title on save so this fixes a very old bug, two, where a changed name never updated the title and breadcrumb item without reload.

pixeljunkie:
The frontend openign problem occured, because there were 2 elements removed from the breadcrumb which lead to a "false" return value for the base url. The double delete was removed and to be sure i added a fallback to base cms url

src/CoreBundle/Resources/public/javascript/tableEditor.js
UlrichKu: sCurrentRecordName is set above (line 586) but not defined. So I guess it exists as an instance/global variable somewheren.

pixeljunkie: Oh, this was an earlier fix attempt. Now we don't need it anymore. Removed it.

PR #243, opened 2019-03-08
ref #349: handle empty dummy element and hide it, highlight active re…

pixeljunkie:
…cord in results via css

| Q             | A
| ------------- | ---
| Branch        | master
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed issues  | chameleon-system/chameleon-system#349
| License       | MIT

### DIFF
--- a/src/CoreBundle/private/modules/MTTableManager/MTTableManager.class.php
+++ b/src/CoreBundle/private/modules/MTTableManager/MTTableManager.class.php
@@ -441,17 +441,19 @@ class MTTableManager extends TCMSModelBase
         $recordList = call_user_func(array($autoClassName.'List', 'GetList'), $this->getAutocompleteListQuery());
 
         $returnVal = [];
-        $returnVal[] = ['id' => ' ', 'text' => ' '];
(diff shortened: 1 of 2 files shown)


src/CoreBundle/private/modules/MTTableManager/MTTableManager.class.php
```diff
@@ -441,17 +441,19 @@ public function getAutocompleteRecordList()
         $recordList = call_user_func(array($autoClassName.'List', 'GetList'), $this->getAutocompleteListQuery());
 
         $returnVal = [];
-        $returnVal[] = ['id' => ' ', 'text' => ' '];
+        $returnVal[] = ['id' => ' ', 'text' => ' ', 'html' => ' ', 'class' => 'd-none'];
```
UlrichKu: Better name it "css-class" (or similiar)?

At least here (in .php) it would be confusable with normal php classes.

pixeljunkie: Changed to "cssClass" (without "-" so it is accessible in javascript as property)

PR #244, opened 2019-03-11
Feature/348 shop related menu items

umulmrum:
| Q             | A
| ------------- | ---
| Branch        | master
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed issues  | chameleon-system/chameleon-system#348
| License       | MIT

The previous solution didn't take positioning into account. The solution proposed here does this while also maintaining all menu categories and menu items in a single place so the updates can be understood more easily.

### DIFF
--- a/UPGRADE-6.3.md
+++ b/UPGRADE-6.3.md
@@ -416,10 +416,6 @@ should be quite straightforward.
 Also some menu items that were located in the top bar were now moved to the sidebar. Finally the backend modules that
 were called in a popup window, like navigation, product search index generation and sanity check, now open inline.
 
-From a technical point no changes are required if the project is a shop system. Projects that are pure CMS systems
-should consider removing menu categories that are nevertheless created during migration. Note that there will be some
-error messages during migration that complain about missing shop tables and modules in this case, which can be ignored. 
-
 ## Home Page Changes
 
 As the main menu is now displayed in the sidebar, the classic main menu was replaced by a welcome screen (which we plan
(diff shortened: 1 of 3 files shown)


UPGRADE-6.3.md
```diff
@@ -416,10 +416,6 @@ should be quite straightforward.
 Also some menu items that were located in the top bar were now moved to the sidebar. Finally the backend modules that
 were called in a popup window, like navigation, product search index generation and sanity check, now open inline.
 
-From a technical point no changes are required if the project is a shop system. Projects that are pure CMS systems
-should consider removing menu categories that are nevertheless created during migration. Note that there will be some
```
pixeljunkie: Why are shop categories created in non shop environments? That should not be the case.

umulmrum: It is no longer the case which is why those lines are removed :-)

pixeljunkie: stuuuupid me! *fetching some coffee

src/CoreBundle/Bridge/Chameleon/Migration/Script/update-1549299689.inc.php
```diff
@@ -6,26 +6,38 @@
...
 $position = 0;
 
+$isShopSystem = \class_exists(TShop::class);
+
 foreach ($categoryLines as $categoryLine) {
     $id = TCMSLogChange::createUnusedRecordId('cms_menu_category');
-    [$systemName, $nameEn, $nameDe, $iconFontCssClass] = \preg_split('/\s+#\s+/', $categoryLine);
+    [$systemType, $systemName, $nameEn, $nameDe, $iconFontCssClass] = \preg_split('/\s+#\s+/', $categoryLine);
+
+    if (false === $isShopSystem && 'shop' === $systemType) {
```
UlrichKu: It is no shop system but the system type is "shop"?

Maybe name it entryType?

umulmrum: It is a type that refers to what kind of "system" is run. So I'd say the name fits - we check if the current system is a shop against the type of system that is required by the entry.

PR #248, opened 2019-03-12
ref #352: add event to catch arrow down key and trigger select2 pulld…

pixeljunkie:
…own open

| Q             | A
| ------------- | ---
| Branch        | master
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed issues  | chameleon-system/chameleon-system#352
| License       | MIT

### DIFF
--- a/src/CoreBundle/Resources/public/javascript/tableEditor.js
+++ b/src/CoreBundle/Resources/public/javascript/tableEditor.js
@@ -869,11 +869,10 @@ CHAMELEON.CORE.MTTableEditor.initSelectBoxes = function () {
         });
     });
 
+    // catch arrow key keypress on select2 containers to open the select2 pulldown.
     $('.select2.select2-container').on('keydown', function(e) {
         if (40 === e.keyCode) {
-
-            var $selectBox = $(e.target).parent().parent().siblings('select')[0];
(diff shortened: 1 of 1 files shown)


src/CoreBundle/Resources/public/javascript/tableEditor.js
```diff
@@ -868,6 +868,14 @@ CHAMELEON.CORE.MTTableEditor.initSelectBoxes = function () {
             });
         });
     });
+
+    $('.select2.select2-container').on('keydown', function(e) {
+        if (40 === e.keyCode) {
+
+            var $selectBox = $(e.target).parent().parent().siblings('select')[0];
```
umulmrum: This lookup seems a bit fragile. Can we jump to a named parent element and then down to the select element?

Also, can we avoid dereferencing the first element here, as we wrap it in a jQuery element in the next line anyway?

pixeljunkie: closes() with a selector was the solution here.

umulmrum:
Thanks @pixeljunkie

PR #253, opened 2019-03-13
ref #92: Remove note on fingers-crossed handler

umulmrum:
| Q             | A
| ------------- | ---
| Branch        | master
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed issues  | chameleon-system/chameleon-system#92
| License       | MIT

### DIFF
--- a/UPGRADE-6.3.md
+++ b/UPGRADE-6.3.md
@@ -209,8 +209,6 @@ monolog:
 ```
 
 Note that the `doctrine` channel MUST be excluded from database logging.
-Also note that the standard logging channel is not configured as "fingerscrossed" anymore. All messages for this channel
-will simply be logged on every request.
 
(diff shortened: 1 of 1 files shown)


UPGRADE-6.3.md
```diff
@@ -209,8 +209,6 @@ monolog:
 ```
 
 Note that the `doctrine` channel MUST be excluded from database logging.
-Also note that the standard logging channel is not configured as "fingerscrossed" anymore. All messages for this channel
```
UlrichKu: Not sure if that is not still the case: there is also the deprecated standard logger (?)

umulmrum: The old standard logger uses the "new" standard logger which is now configured as fingers-crossed (when doing it like suggested in chameleon-system/chameleon-system). So should be the same as before, or don't you think?

PR #256, opened 2019-03-14
ref #359: Use standard page path generation so that special cases are…

umulmrum:
… handled; fix path lowercasing

| Q             | A
| ------------- | ---
| Branch        | 6.2.x
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed issues  | chameleon-system/chameleon-system#359
| License       | MIT

The most important part of the fix is to use the PageServiceUtil to generate the page path. Otherwise special cases like the ".html" suffix is not taken into account correctly.

I'm not sure why we removed the trailing slash before though. Adding it instead seems to to be what we want here.

Fixed URL lowercasing in addition.

### DIFF
--- a/src/CoreBundle/Service/ActivePageService.php
+++ b/src/CoreBundle/Service/ActivePageService.php
@@ -15,6 +15,7 @@ use ChameleonSystem\CoreBundle\CoreEvents;
 use ChameleonSystem\CoreBundle\Event\ChangeActivePageEvent;
 use ChameleonSystem\CoreBundle\Routing\PortalAndLanguageAwareRouterInterface;
 use ChameleonSystem\CoreBundle\Service\Initializer\ActivePageServiceInitializerInterface;
+use ChameleonSystem\CoreBundle\Util\PageServiceUtilInterface;
 use ChameleonSystem\CoreBundle\Util\RoutingUtilInterface;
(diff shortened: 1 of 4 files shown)


src/CoreBundle/Service/ActivePageService.php
```diff
@@ -252,8 +258,11 @@ private function modifyRouteParameters(array &$parameters, TdbCmsLanguage $langu
             if ($activePage->fieldPrimaryTreeIdHidden === $activePortal->fieldHomeNodeId) {
                 unset($parameters['pagePath']);
             } else {
-                $allPageRoutes = $this->routingUtil->getAllPageRoutes($activePortal, $language);
-                $parameters['pagePath'] = rtrim($allPageRoutes[$activePage->id]->getPrimaryPath(), '/');
+                $pagePath = $this->pageServiceUtil->getPagePath($activePage, $language);
+                if (true === $activePortal->fieldUseSlashInSeoUrls) {
```
UlrichKu: Shouldn't do PageServicUtil this?

umulmrum: In theory yes, but I think PageServiceUtil needs to work exactly as is (trailing slashes in URLs are an astonishingly deep lava lake).

PR #260, opened 2019-03-15
ref #356: remove internal breadrumb history index, get it on the fly …

pixeljunkie:
…instead

| Q             | A
| ------------- | ---
| Branch        | master
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | yes
| Tests pass?   | yes
| Fixed issues  | chameleon-system/chameleon-system#356
| License       | MIT

### DIFF
--- a/src/CoreBundle/private/library/classes/TCMSURLHistory.class.php
+++ b/src/CoreBundle/private/library/classes/TCMSURLHistory.class.php
@@ -203,9 +203,6 @@ class TCMSURLHistory
                 unset($this->aHistory[$i]);
             }
         }
-
-        // reset the keys to prevent key gaps
-        $this->aHistory = array_values($this->aHistory);
     }
 
     /**
@@ -247,11 +244,6 @@ class TCMSURLHistory
     {
         if (0 === $this->getHistoryCount()) {
             return true;
-        }
-
-        if (false === isset($this->aHistory[0])) {
-            // reset the keys to prevent key gaps
-            $this->aHistory = array_values($this->aHistory);
         }
 
         return array_key_exists('params', $this->aHistory[0]);
--- a/src/CoreBundle/private/library/classes/TCMSURLHistory.class.php
+++ b/src/CoreBundle/private/library/classes/TCMSURLHistory.class.php
@@ -42,8 +42,15 @@ class TCMSURLHistory
     public function __set($name, $val)
     {
         if ('index' === $name) {
-            throw new \LogicException('Invalid property: '.$name);
+            @trigger_error('The property TCMSURLHistory::$index is deprecated.', E_USER_DEPRECATED);
         }
+
+        $trace = debug_backtrace();
+        trigger_error(sprintf('Undefined property via __set(): %s in %s on line %s',
+            $name,
+            $trace[0]['file'],
+            $trace[0]['line']),
+            E_USER_NOTICE);
     }
 
     public function __isset($name)
--- a/src/CoreBundle/private/library/classes/TCMSURLHistory.class.php
+++ b/src/CoreBundle/private/library/classes/TCMSURLHistory.class.php
@@ -21,7 +21,17 @@ class TCMSURLHistory
      */
     public $aHistory = array();
 
-    public $index = 0;
+    /**
+     * @deprecated since 6.3.0 - use getHistoryIndex() instead
+     */
+    // public $index = 0;
+
+    public function __get($parameterName)
+    {
+        if ('index' === $parameterName) {
+            return $this->getHistoryIndex();
+        }
+    }
 
     /**
      * adds item to the breadcrumb array.
@@ -36,9 +46,6 @@ class TCMSURLHistory
         if (null !== $foundHistoryElementIndex) {
             // element found, so remove it and add the new one at the end.
             $this->removeHistoryElementByIndex($foundHistoryElementIndex);
-        } else {
-            // the history has an internal item count, which needs to be decreased if we don't replace an item.
-            $this->index++;
         }
 
         // add the new item
@@ -78,13 +85,18 @@ class TCMSURLHistory
      */
     public function GetURL()
     {
-        if ($this->index > 0) {
-            $url = $this->aHistory[($this->index - 1)]['url'];
+        $historyIndex = $this->getHistoryIndex();
 
-            return $url;
+        if ($historyIndex > 0 && isset($this->aHistory[$historyIndex - 1])) {
+            return $this->aHistory[$historyIndex - 1]['url'];
         } else {
             return false;
         }
+    }
+
+    public function getHistoryIndex()
+    {
+        return count($this->aHistory);
     }
 
     /**
@@ -94,10 +106,14 @@ class TCMSURLHistory
      */
     public function PopURL()
     {
-        if ($this->index > 0) {
+        $historyIndex = $this->getHistoryIndex();
+
+        if ($historyIndex > 0) {
             $url = $this->GetURL();
-            unset($this->aHistory[($this->index - 1)]);
-            --$this->index;
+            unset($this->aHistory[$historyIndex - 1]);
+
+            // reset the keys to prevent key gaps
+            $this->aHistory = array_values($this->aHistory);
 
             return $url;
         } else {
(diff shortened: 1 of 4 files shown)


src/CoreBundle/private/library/classes/TCMSURLHistory.class.php
```diff
@@ -21,7 +21,17 @@ class TCMSURLHistory
      */
     public $aHistory = array();
 
-    public $index = 0;
+    /**
+     * @deprecated since 6.3.0 - use getHistoryIndex() instead
```
umulmrum: A docblock must belong to a code entity, so we can't use it here. Please trigger a deprecation warning instead (see `TModelBase::__get` for an example).

src/CoreBundle/private/library/classes/TCMSURLHistory.class.php
```diff
@@ -21,7 +21,17 @@ class TCMSURLHistory
...
+     */
+    // public $index = 0;
+
+    public function __get($parameterName)
```
umulmrum: We should also add a `__set` method that reacts on setting `index` as it was a public property before this change and clients could still set it. Maybe triggering a warning is the right thing to do there.

src/CoreBundle/private/library/classes/TCMSURLHistory.class.php
```diff
@@ -21,7 +21,17 @@ class TCMSURLHistory
...
+    {
+        if ('index' === $parameterName) {
+            return $this->getHistoryIndex();
+        }
```
umulmrum: We need to handle other arguments, too - see `TModelBase::__get` for an example on what to do.

src/CoreBundle/private/library/classes/TCMSURLHistory.class.php
```diff
@@ -78,26 +85,35 @@ public function getSimilarHistoryElementIndex(array $newElementParameters): ?int
...
 
-            return $url;
+        if ($historyIndex > 0 && isset($this->aHistory[$historyIndex - 1])) {
+            return $this->aHistory[$historyIndex - 1]['url'];
         } else {
             return false;
         }
     }
 
+    public function getHistoryIndex()
```
umulmrum: We don't return an index but the amount of history items - how about `getHistoryCount`?

There are also multiple occasions in this class where the history elements are counted (old code). Could you call this method instead to be consistent?

src/CoreBundle/private/library/classes/TCMSURLHistory.class.php
```diff
@@ -78,26 +85,35 @@ public function getSimilarHistoryElementIndex(array $newElementParameters): ?int
...
+        $historyIndex = $this->getHistoryIndex();
+
+        if ($historyIndex > 0) {
             $url = $this->GetURL();
-            unset($this->aHistory[($this->index - 1)]);
-            --$this->index;
+            unset($this->aHistory[$historyIndex - 1]);
+
+            // reset the keys to prevent key gaps
+            $this->aHistory = array_values($this->aHistory);
```
umulmrum: Is this really necessary? We remove the last element after all, so no indices should change.

I think we can replace the complete method body with a call to `array_pop` (we only need to return false if array_pop returns null).

src/CoreBundle/private/library/classes/TCMSURLHistory.class.php
```diff
@@ -168,21 +184,22 @@ public function EncodeParameters($aParameters, $returnWithFragment = true)
...
-        $endpoint = count($this->aHistory)-1;
+        $endpoint = count($this->aHistory) - 1;
         if ($endpoint > 0) {
             for ($i = $endpoint; $i > $id; --$i) {
                 unset($this->aHistory[$i]);
             }
         }
+
+        // reset the keys to prevent key gaps
+        $this->aHistory = array_values($this->aHistory);
```
umulmrum: I'd suspect we can also remove this one as we only remove items from the end.

src/CoreBundle/private/library/classes/TCMSURLHistory.class.php
```diff
@@ -220,7 +237,8 @@ public function paramsParameterExists()
         }
 
         if (false === isset($this->aHistory[0])) {
-            $this->aHistory = array_values($this->aHistory); // rearrange elements
+            // reset the keys to prevent key gaps
+            $this->aHistory = array_values($this->aHistory);
```
umulmrum: Not sure why we ever needed this (added it myself a few years ago because the indices were wrong, but how can this happen?).

src/CoreBundle/private/library/classes/TCMSURLHistory.class.php
pixeljunkie: done

src/CoreBundle/private/library/classes/TCMSURLHistory.class.php
pixeljunkie: done, also added __isset

src/CoreBundle/private/library/classes/TCMSURLHistory.class.php
pixeljunkie: done

src/CoreBundle/private/library/classes/TCMSURLHistory.class.php
pixeljunkie: The old property was an index, so it started with zero and count would start with 1. This change would mean, that all external code that worked with the index now counts wrong, so i decided not to work with the count here.

pixeljunkie: OH... just checked the code.. the method really returns the count and not the index. i'm puzzled... will check it and change the method name...
Need more coffee

pixeljunkie: Looks like the "index" was never an index, so i renamed the method.

src/CoreBundle/private/library/classes/TCMSURLHistory.class.php
pixeljunkie: done

src/CoreBundle/private/library/classes/TCMSURLHistory.class.php
pixeljunkie: Makes me nervous, but i removed it. I had issues with a wrong index, but i have no idea, how this happened.

src/CoreBundle/private/library/classes/TCMSURLHistory.class.php
pixeljunkie: Code says "session may contain old history element". I cannot imagine how the index gets these gaps, but i really had situations where this happened, but during the dev process. I tested the current version and could not find any situation where this happened again. So maybe it is fixed.

src/CoreBundle/private/library/classes/TCMSURLHistory.class.php
```diff
@@ -21,7 +21,35 @@ class TCMSURLHistory
...
+            $trace[0]['line']),
+            E_USER_NOTICE);
+
+        return null;
+    }
+
+    public function __set($name, $val)
+    {
+        if ('index' === $name) {
+            throw new \LogicException('Invalid property: '.$name);
```
umulmrum: Can we trigger a deprecation notice here? And a "plain" notice for everything else? I think this is closest to PHP default behavior for nonexisting properties.

src/CoreBundle/private/library/classes/TCMSURLHistory.class.php
umulmrum: Almost - now we return the whole history entry, not just the URL part as before.

src/CoreBundle/private/library/classes/TCMSURLHistory.class.php
umulmrum: Ah right. OK let's be bold and only re-add the fix if we find out it's still needed.

src/CoreBundle/private/library/classes/TCMSURLHistory.class.php
pixeljunkie: gnarf... fixed it

src/CoreBundle/private/library/classes/TCMSURLHistory.class.php
pixeljunkie: done

umulmrum:
Thanks @pixeljunkie

PR #271, opened 2019-03-19
Feature/294 media manager iframe with flex content height

ckleis:
| Q             | A
| ------------- | ---
| Branch        | master
| Bug fix?      | yes
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Fixed issues  | chameleon-system/chameleon-system#294   
| License       | MIT

The iframe for the cutout (inside the mediaManager) always had the height of its container, but the image was usually larger and the iframe had scrollbars.
Now the iframe gets its own height.

In addition, there was a styling adjustment for the save button.


umulmrum:
Thanks @ckleis

PR #279, opened 2019-03-20
ref #320: Remove Flash support

umulmrum:
| Q             | A
| ------------- | ---
| Branch        | master
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | yes
| Deprecations? | no
| Tests pass?   | yes
| Fixed issues  | chameleon-system/chameleon-system#320
| License       | MIT


src/CoreBundle/Bridge/Chameleon/Migration/Script/update-1553067741.inc.php
```diff
@@ -0,0 +1,28 @@
...
+;
+TCMSLogChange::delete(__LINE__, $data);
+
+$data = TCMSLogChange::createMigrationQueryData('cms_smart_url_handler', 'en')
+  ->setWhereEquals([
+      'id' => $flashUrlHandlerId,
+  ])
+;
+TCMSLogChange::delete(__LINE__, $data);
+
```
pixeljunkie: I think we need to handle the now deprecated SWF files in the media library.
I would not delete them, but list them to the admin, so they may be archived if needed and then deleted manually.

src/CoreBundle/private/library/classes/dbobjects/TCMSImageEndpoint.class.php
```diff
@@ -797,7 +728,7 @@ public function CheckAllowThumbnailing()
                     }
 
                     $fileType = $this->GetImageType();
-                    if ('jpg' == $fileType || 'gif' == $fileType || 'png' == $fileType || 'swf' == $fileType) {
+                    if ('jpg' == $fileType || 'gif' == $fileType || 'png' == $fileType) {
                         $allowThumbing = true;
```
UlrichKu: -> $thumbsUp ?

src/CoreBundle/private/library/functions/callback_functions/gcf_ShowImage.fun.php
```diff
@@ -27,8 +27,8 @@ function gcf_ShowImage($field, $row, $fieldName)
             $oImage->useUnsharpMask = false;
             $oImage->Load($imageId);
 
-            if ($oImage->IsFlashMovie() || $oImage->IsExternalMovie()) {
-                $imageTag = $oImage->GetThumbnailTag(100, 75, null, null, '');
+            if ($oImage->IsExternalMovie()) {
+                $imageTag = $oImage->renderImage(100, 75);
```
UlrichKu: renderImage() is equivalent to GetThumbnailTag()?

src/CoreBundle/Bridge/Chameleon/Migration/Script/update-1553067741.inc.php
umulmrum: I'm not sure what happens when the system tries to display yet-existing Flash files in the backend. And they could still be linked in text fields. So yes, we should check this.

src/CoreBundle/private/library/classes/TCMSListManager/TCMSListManagerWYSIWYGImage.class.php
```diff
@@ -106,7 +106,7 @@ public function GetCustomRestriction()
     }
 
     /**
-     *  Add media filter for flash videos and images that be got from URL.
+     *  Add media filter for images that be got from URL.
```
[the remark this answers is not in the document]
umulmrum: Not sure what the text is trying to say, but changed it.

src/CoreBundle/private/library/functions/callback_functions/gcf_ShowImage.fun.php
umulmrum: Yes.

src/CoreBundle/Bridge/Chameleon/Migration/Script/update-1553067741.inc.php
umulmrum: Rechecked what happens after migration. In the media manager the files can still be viewed, just without the preview image. Flash files used in text fields are still displayed in the frontend, as the object tag is rendered on save in the table editor. So nothing explodes and the default behavior is OK I think.

Therefore I just added a note in the upgrade guide, so that developers/administrators remove usages manually (in the rare case that there are massive amounts of Flash usages developers need to be creative).

UlrichKu:
There would be conflicts.

umulmrum:
Solved.

PR #284, opened 2019-03-21
Feature/323 remove deprecations

umulmrum:
| Q             | A
| ------------- | ---
| Branch        | master
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | yes
| Deprecations? | no
| Tests pass?   | yes
| License       | MIT

This is a first PR to remove misc deprecated code. Although there are ~5k lines removed, there are still many deprecations left.


src/CoreBundle/private/library/classes/dbobjects/TCMSPage.class.php
```diff
@@ -446,6 +433,50 @@ public function GetRealURL($aAdditionalParameters = array(), $aExcludeParameters
...
+    /**
+     * returns all POST and GET parameters as url.
+     *
+     * @param array $excludeArray
+     *
+     * @return string
+     */
+    private function OutputDataAsURL($excludeArray = array())
+    {
+        $aData = TGlobal::instance()->GetUserData();
```
UlrichKu: This is also deprecated (?)

src/CoreBundle/private/modules/CMSModuleChooser/CMSModuleChooser.class.php
```diff
@@ -1035,9 +1035,8 @@ protected function TriggerWorkflowLockForPage($sPageId = false)
         if (false === $sPageId) {
             $sPageId = $this->global->GetUserData('pagedef');
         }
-        $oCmsTplPage = TCMSActivePage::GetInstance();
```
UlrichKu: TCMSActivePage could now be deprecated completely?

UPGRADE-7.0.md
```diff
@@ -0,0 +1,207 @@
...
+- ChameleonSystem\CoreBundle\Controller\ChameleonController::$requestInfoService
+- MTPageMetaCoreEndPoint::$oActivePortal
+- TCMSMemcache::$sUsedMemcacheClass
+- TGlobal::$oURLHistory
+- TGlobalBase::$aUnitTestMockedObjects
+- TModelBase::$isExportCall
+
+## Methods
+
+- ChameleonSystem\CoreBundle\Controller\ChameleonController::GetExecutionTime()
```
UlrichKu: The other PR has no () here...

src/CommentBundle/objects/db/TPkgComment.class.php
```diff
@@ -304,14 +306,14 @@ protected function GetCommentURLData($iCommentPage = false)
     public function GetChildComments($iLanguageId = null)
     {
         $oList = $this->GetFromInternalCache('CommentChildList');
-        if (is_null($oList)) {
+        if ($oList === null) {
```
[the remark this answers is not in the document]
umulmrum: Changed.

src/CoreBundle/private/library/classes/dbobjects/TCMSPage.class.php
umulmrum: The method was just moved from TGlobalBase without changes. A handful of extra usages won't hurt and can be changed later (touching hundreds of files, so I'd like to avoid effort and errors for now).

src/CoreBundle/private/modules/CMSModuleChooser/CMSModuleChooser.class.php
umulmrum: There are still methods in TCMSActivePage that aren't deprecated yet. I'd like to do this in a future version.

UPGRADE-7.0.md
umulmrum: I prefer the variant with brackets, but we can still change near the end of the release.

src/CoreBundle/private/library/classes/dbobjects/TCMSPage.class.php
UlrichKu: You want to revisit this with another PR (because I think the issue will be the same)?

umulmrum: We will keep some deprecated code in 7.0 as it's too much to remove everything at once. TGlobalBase::Instance() is definitely one of those methods that will stay, as there are currently 327 usages in chameleon-base and chameleon-shop alone. We will need to reduce usage of those methods in an ongoing effort so that it can then be removed in 8.0.

The ultimate goal with TGlobal and TGlobalBase (and TTools) should be to remove them completely as they are global util classes that should never have existed. So there will be more deprecations in future minor releases and less and less need for the instance() method, so some of the work should be done "automatically".

PR #285, opened 2019-03-22
#372: Check for existence of field data in "post data"

UlrichKu:
| Q             | A
| ------------- | ---
| Branch        | 6.3.x
| Bug fix?      | yes/
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed issues  | chameleon-system/chameleon-system#372
| License       | MIT

### DIFF
--- a/src/CoreBundle/private/library/classes/TCMSTableEditor/TCMSTableFieldWriter.class.php
+++ b/src/CoreBundle/private/library/classes/TCMSTableEditor/TCMSTableFieldWriter.class.php
@@ -152,18 +152,24 @@ class TCMSTableFieldWriter extends TCMSTableEditor
 
         $this->_oField = $this->adaptFieldAndRelatedTables($oPostTable->sqlData);
 
-        $this->getAutoclassesCacheWarmer()->updateTableById($oPostTable->sqlData['cms_tbl_conf_id']);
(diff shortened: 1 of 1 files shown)


src/CoreBundle/private/library/classes/TCMSTableEditor/TCMSTableFieldWriter.class.php
```diff
@@ -152,18 +152,24 @@ protected function PostSaveHook(&$oFields, &$oPostTable)
...
     /**
      * @param array $postData
      *
-     * @return TCMSField - the new definition after changes
+     * @return TCMSField|null - the new definition after changes; or null for small changes (without type)
      *
      * @throws DBALException
      */
-    protected function adaptFieldAndRelatedTables(array $postData): TCMSField
+    protected function adaptFieldAndRelatedTables(array $postData): ?TCMSField
```
umulmrum: This is a BC break that is not allowed in a fix release. Can we instead only call this method if the requirements are met?

UlrichKu: ok

umulmrum:
Thanks @UlrichKu

PR #287, opened 2019-03-22
ref #373: delete CMSFieldMLT - not used anywhere

pixeljunkie:
| Q             | A
| ------------- | ---
| Branch        | master
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | yes
| Deprecations? | no
| Tests pass?   | yes
| Fixed issues  | chameleon-system/chameleon-system#373
| License       | MIT


pixeljunkie:
The class is not used since 2009 says the internal GIT and it was marked for removal with a deleteme txt file. I know this is not the right practice, but this were other times. And the file is useless because the database structure changed. Would it be o.k. for you if we mark it deprecated in 6.3.1. I would not want to wait another year to get rid of this orphan.

umulmrum:
OK let's do this. But there is a merge conflict - could you have a look?

pixeljunkie:
merge conflict fixed.

umulmrum:
Thanks @pixeljunkie

PR #291, opened 2019-03-26
Feature/323 cleanup

umulmrum:
| Q             | A
| ------------- | ---
| Branch        | master
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| License       | MIT

### DIFF
--- a/src/CoreBundle/private/library/classes/TCMSLogChange.class.php
+++ b/src/CoreBundle/private/library/classes/TCMSLogChange.class.php
@@ -1780,6 +1780,51 @@ class TCMSLogChange
     }
 
     /**
+     * This method lets old updates insert shop system pages. The table "shop_system_page" has been removed in shop-165.inc.php and
+     * old updates would fail when triggered afterwards.
+     *
+     * You should not use this method in new packages, which already know about the change.
+     *
+     * @param $shop_id
+     * @param $name_internal
+     * @param $name
+     * @param string $cms_tree_id
+     *
+     * @deprecated since 6.2.0 - no longer used.
+     */
+    public static function addShopSystemPage($shop_id, $name_internal, $name, $cms_tree_id = '', $id = null)
+    {
+        if (self::TableExists('shop_system_page')) {
+            $query = "INSERT INTO `shop_system_page`
+                      SET `shop_id` = '".$shop_id."',
+                          `name_internal` = '".$name_internal."',
+                          `name` = '".$name."',
+                          `cms_tree_id` = '".$cms_tree_id."'";
+            if (null !== $id) {
+                $query .= ", `id`='".MySqlLegacySupport::getInstance()->real_escape_string($id)."'";
+            }
+            self::_RunQuery($query, __LINE__);
+        } else {
+            $shop = TdbShop::GetNewInstance();
+            if ($shop->Load($shop_id)) {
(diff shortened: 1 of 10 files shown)


src/CoreBundle/private/library/classes/TCMSLogChange.class.php
```diff
@@ -1779,6 +1779,51 @@ public static function escapeJSOutput($str)
...
+     * @param $shop_id
+     * @param $name_internal
+     * @param $name
+     * @param string $cms_tree_id
+     *
+     * @deprecated since 6.2.0 - no longer used.
+     */
+    public static function addShopSystemPage($shop_id, $name_internal, $name, $cms_tree_id = '', $id = null)
+    {
+        if (self::TableExists('shop_system_page')) {
```
UlrichKu: true ===

also below

src/CoreBundle/private/library/classes/TCMSLogChange.class.php
```diff
@@ -1779,6 +1779,51 @@ public static function escapeJSOutput($str)
...
+     * You should not use this method in new packages, which already know about the change.
+     *
+     * @param $shop_id
+     * @param $name_internal
+     * @param $name
+     * @param string $cms_tree_id
+     *
+     * @deprecated since 6.2.0 - no longer used.
+     */
+    public static function addShopSystemPage($shop_id, $name_internal, $name, $cms_tree_id = '', $id = null)
```
UlrichKu: Where is this/will this be used?

src/CoreBundle/private/library/classes/TCMSLogChange.class.php
```diff
@@ -1779,6 +1779,51 @@ public static function escapeJSOutput($str)
...
+                $query .= ", `id`='".MySqlLegacySupport::getInstance()->real_escape_string($id)."'";
+            }
+            self::_RunQuery($query, __LINE__);
+        } else {
+            $shop = TdbShop::GetNewInstance();
+            if ($shop->Load($shop_id)) {
+                $portals = $shop->GetFieldCmsPortalIdList();
+                foreach ($portals as $portalid) {
+                    $query = "INSERT INTO `cms_portal_system_page` SET
+                                `cms_portal_id` =   '".MySqlLegacySupport::getInstance()->real_escape_string($portalid)."',
```
UlrichKu: Is this copied from somewhere?

Otherwise on could use "database_connection" for the escaping.

src/CoreBundle/private/library/classes/TCMSLogChange.class.php
umulmrum: The method was prematurely removed and is only restored in its former state without any improvements - as it is subject to removal, I'd like to keep it as is.

src/CoreBundle/private/library/classes/TCMSLogChange.class.php
umulmrum: Is it used by dependent code outside of the Chameleon core system.

src/CoreBundle/private/library/classes/TCMSLogChange.class.php
umulmrum: See above.

PR #292, opened 2019-03-27
Feature/323-cleanup

umulmrum:
| Q             | A
| ------------- | ---
| Branch        | master
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | yes
| Deprecations? | no
| Tests pass?   | yes
| License       | MIT

The last batch of removed deprecated code. I also added a second commit that cleans some unused code.


src/CoreBundle/private/library/classes/TCMSListManager/TCMSListManagerMediaSelector.class.php
```diff
@@ -205,12 +205,6 @@ protected function GetMaxHeight()
 
         if ($this->oGlobal->UserDataExists('imageHeight')) {
             $iHeight = $this->oGlobal->GetUserData('imageHeight');
-        } else {
-            if (isset($this->oDefaultImage->aData) && is_array($this->oDefaultImage->aData)) {
-                if (array_key_exists('height', $this->oDefaultImage->aData) && $this->oDefaultImage->aData['height'] > 0) {
-                    $iWidth = $this->oDefaultImage->aData['height'];
```
UlrichKu: Why is this removed?

The wrong variable name ($iWidth instead of $iHeight) shows that this was not used?

umulmrum: Ah right, I didn't see that the wrong name led to the unused-message. This has been unchanged for almost 9 years now, so I'm inclined to just keep it removed, as no one seems to complain.

UlrichKu: The documentation should be adapted then.

umulmrum: Done.

PR #295, opened 2019-03-28
ref #25: Remove old media manager

umulmrum:
| Q             | A
| ------------- | ---
| Branch        | master
| Bug fix?      | no
| New feature?  |no
| BC breaks?    | yes
| Deprecations? | no
| Tests pass?   | yes
| Fixed issues  | chameleon-system/chameleon-system#25
| License       | MIT

There were some translations and even classes that were not marked as deprecated before, assumably because they simply forgotten. If we did 7.0.0 parallel to 6.3.0, we could mark as deprecated in 6.3.0, but now we can only choose between keeping or removing code (should keep this in mind for the next major release). I chose to remove this code, as it doesn't make sense without the old media manager. Let me know if you dissent.


src/CoreBundle/Bridge/Chameleon/Migration/Script/update-1553768043.inc.php
```diff
@@ -0,0 +1,29 @@
+<h1>Build #1553768043</h1>
+<h2>Date: 2019-03-28</h2>
+<div class="changelog">
+    - Reset field class and fix translations for cms_media.width
+</div>
+<?php
+
+$fieldId = TCMSLogChange::GetTableFieldId(TCMSLogChange::GetTableId('cms_media'), 'width');
```
UlrichKu: Doesn't this field (still) have the wrong type? Properties if I remember correctly. There should be a ticket / PR for this somewhere.

src/MediaManagerBundle/DependencyInjection/Compiler/ReplaceUrlGeneratorPass.php
```diff
@@ -1,59 +0,0 @@
...
-use ChameleonSystem\MediaManagerBundle\DependencyInjection\Configuration;
-use LogicException;
-use Symfony\Component\Config\Definition\Processor;
-use Symfony\Component\DependencyInjection\Compiler\CompilerPassInterface;
-use Symfony\Component\DependencyInjection\ContainerBuilder;
-use Symfony\Component\DependencyInjection\Exception\InvalidArgumentException;
-use Symfony\Component\DependencyInjection\Exception\ServiceNotFoundException;
-use Symfony\Component\DependencyInjection\Reference;
-
-class ReplaceUrlGeneratorPass implements CompilerPassInterface
```
UlrichKu: What was this doing?

(It's not deprecated?)

src/CoreBundle/Bridge/Chameleon/Migration/Script/update-1553768043.inc.php
umulmrum: The type is correct, it's just the field class that needs to be removed so that the default class is used - this is done below in line 13.

src/MediaManagerBundle/DependencyInjection/Compiler/ReplaceUrlGeneratorPass.php
umulmrum: The pass replaced a class in a service as soon as the bundle was installed. This is no longer required, as the class from this bundle is now the default (defined as a service in services.xml, configured in the bundle extension class).

As written above, there were some classes for which the @deprecated annotation was forgotten - we need to decide if we remove those (I'd say it's OK).

PR #297, opened 2019-03-28
ref #351: Remove pagedef from GET parameters

umulmrum:
| Q             | A
| ------------- | ---
| Branch        | master
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | yes
| Deprecations? | no
| Tests pass?   | yes
| Fixed issues  | chameleon-system/chameleon-system#351
| License       | MIT

### DIFF
--- a/UPGRADE-7.0.md
+++ b/UPGRADE-7.0.md
@@ -1,6 +1,52 @@
 UPGRADE FROM 6.3 TO 7.0
 =======================
 
+# Essentials
+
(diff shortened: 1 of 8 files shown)


umulmrum:
@pixeljunkie You had concerns if this change leads to lots of breaking projects. I think this isn't the case, see the UPGRADE guide. But please have a look at the PR and tell if you still find it problematic.

PR #311, opened 2019-04-05
ref #381: backport from 7.0: reset field class and fix translations f…

pixeljunkie:
…or cms_media.width

| Q             | A
| ------------- | ---
| Branch        | 6.3.x
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed issues  | chameleon-system/chameleon-system#381
| License       | MIT

### DIFF
--- a/src/CoreBundle/Bridge/Chameleon/Migration/Script/update-1553768043.inc.php
+++ b/src/CoreBundle/Bridge/Chameleon/Migration/Script/update-1553768043.inc.php
@@ -27,3 +27,5 @@ $data = TCMSLogChange::createMigrationQueryData('cms_field_conf', 'de')
     ])
 ;
 TCMSLogChange::update(__LINE__, $data);
+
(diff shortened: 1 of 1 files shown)


src/CoreBundle/Bridge/Chameleon/Migration/Script/update-1553768043.inc.php
```diff
@@ -0,0 +1,29 @@
+<h1>Build #1553768043</h1>
+<h2>Date: 2019-03-28</h2>
+<div class="changelog">
+    - Reset field class and fix translations for cms_media.width
+</div>
+<?php
+
+$fieldId = TCMSLogChange::GetTableFieldId(TCMSLogChange::GetTableId('cms_media'), 'width');
```
UlrichKu: @ Backwards incompatibility:
You could add an update message (INFO) here that the "width" field was restored.

And if somebody really needs the current state they can revert this update.

That would make it easier for 99% and give the potential 1% using the old media manager the chance to know and correct.

pixeljunkie: done

PR #313, opened 2019-04-08
ref #323: Remove CMS search

umulmrum:
| Q             | A
| ------------- | ---
| Branch        | master
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | yes
| Deprecations? | no
| Tests pass?   | yes
| License       | MIT


UPGRADE-7.0.md
```diff
@@ -55,6 +55,12 @@ Support for Adobe Flash was removed. We recommend to search the media manager fo
 extensions "flv", "f4v" and "swf") and remove them.
 The media manager will also display where these files are still used; these usages should also be removed.
 
+## Remove module_cms_search Table
+
+The CMS search module was removed. If the project still needs this module, restore it from an earlier Chameleon release.
+Otherwise remove the associated database table by calling `TCMSLogChange::deleteTable('module_cms_search')` in an update
```
pixeljunkie: What about the template module record?

umulmrum: Added an update that removes the module :-)

UPGRADE-7.0.md
```diff
@@ -55,6 +55,12 @@ Support for Adobe Flash was removed. We recommend to search the media manager fo
 extensions "flv", "f4v" and "swf") and remove them.
 The media manager will also display where these files are still used; these usages should also be removed.
 
+## Remove module_cms_search Table
+
+The CMS search module was removed. If the project still needs this module, restore it from an earlier Chameleon release.
```
UlrichKu: "Restoring" means to restore all the below listed classes, translations, ...?

umulmrum: Yes. All of these can be added to project code without relying on Chameleon. I expect that this case is *really* rare, so we shouldn't invest time to simplify this - if developers have difficulties in doing this, we can still help.

PR #319, opened 2019-04-09
ref #393: Avoid going back in history

umulmrum:
| Q             | A
| ------------- | ---
| Branch        | 6.3.x
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed issues  | chameleon-system/chameleon-system#393
| License       | MIT

I'm not sure why we needed this "redirect" in the past, but think that it is no longer required because of backend changes in 6.3.x. Not too familiar with this region of the code, maybe @pixeljunkie can tell if this is a good idea.

### DIFF
--- a/src/CoreBundle/Resources/public/javascript/tableEditor.js
+++ b/src/CoreBundle/Resources/public/javascript/tableEditor.js
@@ -585,10 +585,6 @@ function SaveViaAjaxCallback(data, statusText) {
 
         $('#tableEditorContainer .navbar-brand').html(data.name);
         $('#cmsbreadcrumb .breadcrumb-item:last').html(data.name);
-
-        if (self !== top) {
-            window.history.go(-1);
-        }
     } else {
         toasterMessage(CHAMELEON.CORE.i18n.Translate('chameleon_system_core.js.error_save'), 'ERROR');
     }


umulmrum:
This is also the cause of chameleon-system/chameleon-system#395 - so we should think about merging  into 6.2.x instead.

pixeljunkie:
I think the intention of this was to return to the previous list page in an iframe, because we don't have a back button here. The problem has something to do with calling save in sub frames like this is the case on products with the product stock record. I will look into this.

src/CoreBundle/Resources/public/javascript/tableEditor.js
```diff
@@ -585,10 +585,6 @@ function SaveViaAjaxCallback(data, statusText) {
 
         $('#tableEditorContainer .navbar-brand').html(data.name);
         $('#cmsbreadcrumb .breadcrumb-item:last').html(data.name);
-
-        if (self !== top) {
```
UlrichKu: This only matches/targets the describe problem in the media manager?

Or ar there other "popup-like things"?

pixeljunkie: It is not media manager related. It is related to all table editors in frames (we don't have such a testcase in the default shop).

pixeljunkie:
After we found other cases where this creates strange behaviors (returning to an ajax call in the history), this should be removed asap.

UlrichKu:
Also cherry-picked for 6.2.x.

PR #324, opened 2019-04-10
ref #402: Prohibit page call if the page is not part of a navigation

umulmrum:
| Q             | A
| ------------- | ---
| Branch        | 6.2.x
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | see below
| Deprecations? | no
| Tests pass?   | yes
| Fixed issues  | chameleon-system/chameleon-system#402-->
| License       | MIT

We should be careful when merging this fix. While it fixes a BC break, this break has been around for some time, so people might rely on the current behavior. I included an update message to warn developers though.

### DIFF
--- a/src/CoreBundle/Bridge/Chameleon/Migration/Script/update-1554901333.inc.php
+++ b/src/CoreBundle/Bridge/Chameleon/Migration/Script/update-1554901333.inc.php
@@ -6,5 +6,5 @@
 
 TCMSLogChange::addInfoMessage('Pages are now only available if they are part of a navigation (see navigation
 options in the portal settings). This restores the behavior of a previous Chameleon release; it was never intended that
-any pages are accessible in the frontend. Check if all pages that should be available in the frontend are part of a
+every page is accessible in the frontend. Ensure that all pages that should be available in the frontend are part of a
 navigation.', TCMSLogChange::INFO_MESSAGE_LEVEL_TODO);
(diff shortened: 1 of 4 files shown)


src/CoreBundle/Bridge/Chameleon/Migration/Script/update-1554901333.inc.php
```diff
@@ -0,0 +1,10 @@
+<h1>Build #1551195244</h1>
+<h2>Date: 2019-02-26</h2>
+<div class="changelog">
+</div>
+<?php
+
+TCMSLogChange::addInfoMessage('Pages are now only available if they are part of a navigation (see navigation
```
UlrichKu: How big is the probability (risk) that this is actually the case?

src/CoreBundle/Bridge/Chameleon/Migration/Script/update-1554901333.inc.php
```diff
@@ -0,0 +1,10 @@
...
+options in the portal settings). This restores the behavior of a previous Chameleon release; it was never intended that
+any pages are accessible in the frontend. Check if all pages that should be available in the frontend are part of a
```
UlrichKu: "any"? not "all"?

any would mean "more than zero" here.

umulmrum: Improved wording.

src/CoreBundle/Bridge/Chameleon/Migration/Script/update-1554901333.inc.php
umulmrum: I'm afraid I have no idea. Some developers might not be aware that this was a BC break, so I think there could be some systems with pages that will no longer be available. But that's why update messages are added in cases we think are important enough to justify behavior changes even in fix releases.

PR #326, opened 2019-04-12
ref #276: Add delete confirmation dialog

UlrichKu:
| Q             | A
| ------------- | ---
| Branch        | 6.2.x
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed issues  | chameleon-system/chameleon-system#276
| License       | MIT

### DIFF
--- a/src/MediaManagerBundle/Resources/public/js/mediaManager.js
+++ b/src/MediaManagerBundle/Resources/public/js/mediaManager.js
@@ -140,6 +140,11 @@
                                 label: CHAMELEON.CORE.i18n.Translate('chameleon_system_media_manager.tree_context_menu.delete_folder'),
                                 icon: "glyphicon glyphicon-trash",
                                 action: function (data) {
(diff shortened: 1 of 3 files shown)


pixeljunkie:
I would add the missing blocking screen using the "PleaseWait()" method which is listed in #276 because it is no big deal to add and the other issue which moved to follow-up issue will be more complicated and maybe not part of the next release.

UlrichKu:
PleaseWait() added.

src/MediaManagerBundle/Resources/public/js/mediaManager.js
```diff
@@ -289,10 +295,12 @@
                     'id': encodeURIComponent(id)
                 },
                 error: function (responseData) {
+                    $.unblockUI();
```
pixeljunkie: Just remembered $.unblockUI() is deprecated in 6.3. As of 6.3 we need to call: CHAMELEON.CORE.showProcessingModal() and CHAMELEON.CORE.hideProcessingModal(); So either we need seperate branches, or after the upmerge a change in 6.3/7.0.

UlrichKu: Added #408 for this.

PR #330, opened 2019-04-23
Fix/403 extended lookup multi table relations delete

pixeljunkie:
| Q             | A
| ------------- | ---
| Branch        | 6.2.x
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed issues  | chameleon-system/chameleon-system#403
| License       | MIT

### DIFF
--- a/src/CoreBundle/private/library/classes/TCMSTableEditor/TCMSTableEditorEndPoint.class.php
+++ b/src/CoreBundle/private/library/classes/TCMSTableEditor/TCMSTableEditorEndPoint.class.php
@@ -1023,13 +1023,6 @@ class TCMSTableEditorEndPoint
         /** @var TCMSMLTField $oField */
         $oField = $this->oTableConf->GetField($sFieldName, $this->oTable);
 
-        $sTargetTable = $oField->GetConnectedTableName();
-        $sTargetTableName = TCMSTableToClass::GetClassName('Tdb', $sTargetTable);
-        if (class_exists($sTargetTableName)) {
-            $oTargetTable = call_user_func(array($sTargetTableName, 'GetNewInstance'));
-            $oTargetTable->Load($sConnectedID);
-        }
-
         $this->RemoveMLTConnectionExecute($oField, $sConnectedID);
     }
 
--- a/src/CoreBundle/private/library/classes/TCMSTableEditor/TCMSTableEditorEndPoint.class.php
+++ b/src/CoreBundle/private/library/classes/TCMSTableEditor/TCMSTableEditorEndPoint.class.php
@@ -2094,22 +2094,23 @@ class TCMSTableEditorEndPoint
     {
         $databaseConnection = $this->getDatabaseConnection();
 
-        $query = 'SELECT `id` FROM `cms_field_type` WHERE `constname` = '.$databaseConnection->quote(TCMSFieldExtendedLookupMultiTable::FIELD_SYSTEM_NAME);
-        $fieldTypeId = $databaseConnection->fetchColumn($query);
+        $fieldTypeId = $this->getFieldTypeIdBySystemName(TCMSFieldExtendedLookupMultiTable::FIELD_SYSTEM_NAME);
 
-        if (false === $fieldTypeId) {
+        if (null === $fieldTypeId) {
             return;
         }
 
-        $query = 'SELECT 
+        $fieldConfigQuery = 'SELECT 
                          `cms_field_conf`.`name` AS fieldName,
                          `cms_tbl_conf`.`name` AS tableName
                     FROM `cms_field_conf` 
                LEFT JOIN `cms_tbl_conf` ON `cms_tbl_conf`.`id` = `cms_field_conf`.`cms_tbl_conf_id` 
                    WHERE `cms_field_conf`.`cms_field_type_id` = :fieldTypeId';
-        $result = $databaseConnection->fetchAll($query, ['fieldTypeId' => $fieldTypeId]);
-        foreach ($result as $row) {
-            $query = 'UPDATE 
+        $fieldConfigResult = $databaseConnection->fetchAll($fieldConfigQuery, ['fieldTypeId' => $fieldTypeId]);
+
+        $recordedQueries = [];
+        foreach ($fieldConfigResult as $row) {
+            $updateQuery = 'UPDATE 
                          '.$databaseConnection->quoteIdentifier($row['tableName']).'
                      SET 
                          '.$databaseConnection->quoteIdentifier($row['fieldName'].TCMSFieldExtendedLookupMultiTable::TABLE_NAME_FIELD_SUFFIX)." = '',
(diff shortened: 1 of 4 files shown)


src/CoreBundle/private/library/classes/TCMSTableEditor/TCMSTableEditorEndPoint.class.php
```diff
@@ -1023,13 +1023,6 @@ public function RemoveMLTConnection($sFieldName, $sConnectedID = false)
         /** @var TCMSMLTField $oField */
         $oField = $this->oTableConf->GetField($sFieldName, $this->oTable);
 
-        $sTargetTable = $oField->GetConnectedTableName();
```
pixeljunkie: Note: During development i stumbled several times on this method and these part has no functionality. It loads a record, sets it to $oTargetTable but is not used anywhere. This looks like an orphan from the workflow feature or something. So i removed it, so other people will not scratch their heads like i did.

src/CoreBundle/private/library/classes/dbobjects/TCMSTableConf.class.php
```diff
@@ -281,13 +281,13 @@ public function &GetField($sFieldName, &$oTableRow, $loadDefaults = false)
...
     /**
-     * returns an iterator of field definitions that match the types given in aFieldTypes
+     * Returns a record list of field definitions that match the types given in aFieldTypes
      * (aFieldTypes = `constname` in cms_field_type).
      *
      * @param array $aFieldTypes
      * @param bool  $bDoNotUseAutoObjects - set to true if you want to prevent the class from using auto objects
      *
-     * @return TIterator
+     * @return TCMSRecordList|TdbCmsFieldConfList
```
pixeljunkie: TIterator is not wrong, but totally a useless info here, because it is always a recordList and autocomplete is not working with TIterator, so i fixed this.

src/CoreBundle/private/library/classes/TCMSTableEditor/TCMSTableEditorEndPoint.class.php
```diff
@@ -2096,6 +2090,44 @@ public function DeleteIdConnectedRecordReferences()
         }
     }
 
+    protected function deleteMultiTableRecordReferences(string $tableName, string $id = ''): void
+    {
+        $databaseConnection = $this->getDatabaseConnection();
+
+        $query = 'SELECT `id` FROM `cms_field_type` WHERE `constname` = '.$databaseConnection->quote(TCMSFieldExtendedLookupMultiTable::FIELD_SYSTEM_NAME);
+        $fieldTypeId = $databaseConnection->fetchColumn($query);
```
UlrichKu: There are three queries in this method (all with name "query")?

One early return, one get all configured connections and the delete all data?

Maybe name them accordingly or even move to appropriately named method?

src/CoreBundle/private/library/classes/TCMSTableEditor/TCMSTableEditorEndPoint.class.php
```diff
@@ -2096,6 +2090,44 @@ public function DeleteIdConnectedRecordReferences()
...
+                         '.$databaseConnection->quoteIdentifier($row['fieldName'].TCMSFieldExtendedLookupMultiTable::TABLE_NAME_FIELD_SUFFIX)." = '',
+                         ".$databaseConnection->quoteIdentifier($row['fieldName'])." = ''   
+                   WHERE 
+                         ".$databaseConnection->quoteIdentifier($row['fieldName'].TCMSFieldExtendedLookupMultiTable::TABLE_NAME_FIELD_SUFFIX).' = '.$databaseConnection->quote($tableName);
+
+            if ('' !== $id) {
+                $query .= ' AND '.$databaseConnection->quoteIdentifier($row['fieldName']).' = '.$databaseConnection->quote($id);
+            }
+
+            $databaseConnection->query($query);
```
UlrichKu: How many entries/updates can this be?

src/CoreBundle/private/library/classes/TCMSTableEditor/TCMSTableWriter.class.php
```diff
@@ -354,6 +354,18 @@ public function Delete($sId = null)
...
+    /**
+     * {@inheritdoc}
+     */
+    public function DeleteConnectedRecordReferences()
+    {
+        $this->DeleteIdConnectedRecordReferences();
+        $this->DeleteMltConnectedRecordReferences();
+
+        // delete all related records targeting the deleted table
+        $this->deleteMultiTableRecordReferences($this->oTableConf->sqlData['name']);
```
UlrichKu: Name the method (for example) `deleteRecordReferencesTargeting($table)` and get rid of the comment?

src/CoreBundle/private/library/classes/TCMSTableEditor/TCMSTableEditorEndPoint.class.php
```diff
@@ -2096,6 +2090,44 @@ public function DeleteIdConnectedRecordReferences()
...
+
+            $queries[] = new LogChangeDataModel($query, LogChangeDataModel::TYPE_CUSTOM_QUERY);
```
UlrichKu: Is this collected somehow?

Because it is also written directly in the loop.

src/CoreBundle/private/library/classes/TCMSTableEditor/TCMSTableEditorEndPoint.class.php
pixeljunkie: This depends on the usage of this fieldtype and the related tables. The query is a per table update, but only updates the records, that have the table name or table name and id set, so i would not call an update per record. This would be slower i'm sure.

src/CoreBundle/private/library/classes/TCMSTableEditor/TCMSTableWriter.class.php
pixeljunkie: The method needs info about what records are deleted (MultiTable references). The difference of this call is only that i only set the table. This handles the deletion of the whole table where there is no deletion record by record first. So i  introduced a new method called: deleteMultiTableRecordReferencesForDeletedTable

pixeljunkie: And the method can call the parent here i found out.

src/CoreBundle/private/library/classes/TCMSTableEditor/TCMSTableEditorEndPoint.class.php
pixeljunkie: This was on error. Of course it should collect the queries and at them at once to the log recorder.

src/CoreBundle/private/library/classes/TCMSTableEditor/TCMSTableEditorEndPoint.class.php
pixeljunkie: moved one part out and renamed the other two.

src/CoreBundle/private/library/classes/TCMSTableEditor/TCMSTableEditorEndPoint.class.php
UlrichKu: For earlier versions this could replace the $sConectedId (something to do with translations).

PR #333, opened 2019-04-25
Fix/403 extended lookup multi table relations delete

pixeljunkie:
| Q             | A
| ------------- | ---
| Branch        | 6.2.x
| Bug fix?      | yes
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed issues  | chameleon-system/chameleon-system#403
| License       | MIT

Introduces methods for easier Index creation and deletion in TCMSFields.
Fixes issues with the extendedLookupMultiTable field and adds a missing index.


src/CoreBundle/private/library/classes/TCMSFields/TCMSField.class.php
```diff
@@ -659,28 +680,51 @@ public function CreateFieldIndex($returnDDL = false)
...
+
+        if ('INDEX' !== $indexType && 'UNIQUE' !== $indexType) {
             return $returnDDL ? '' : null;
         }
 
-        if ('index' === $oFieldType->sqlData['indextype']) {
-            $indexType = 'INDEX';
-        } elseif ('unique' === $oFieldType->sqlData['indextype']) {
-            $indexType = 'UNIQUE';
+        if (true === $returnDDL) {
```
UlrichKu: Could be moved only to the return. The `getIndexQuery` is done in any case. Or it should be?

src/CoreBundle/private/library/classes/TCMSFields/TCMSFieldExtendedLookupMultiTable.class.php
```diff
@@ -527,33 +526,33 @@ public function RemoveFieldIndex()
     {
         parent::RemoveFieldIndex();
 
-        $query = 'ALTER TABLE `'.MySqlLegacySupport::getInstance()->real_escape_string($this->sTableName).'` DROP INDEX `'.MySqlLegacySupport::getInstance()->real_escape_string($this->name).'_combined`';
-
-        MySqlLegacySupport::getInstance()->query($query);
-        $aQuery = array(new LogChangeDataModel($query));
-        TCMSLogChange::WriteTransaction($aQuery);
+        $this->dropIndexByName($this->name.self::TABLE_NAME_FIELD_SUFFIX);
+        $this->dropIndexByName($this->name.'_combined');
```
UlrichKu: Was this missing? I only see one drop so far?

src/CoreBundle/private/library/classes/TCMSTableEditor/TCMSTableEditorEndPoint.class.php
```diff
@@ -2093,41 +2093,53 @@ public function DeleteIdConnectedRecordReferences()
     protected function deleteMultiTableRecordReferences(string $tableName, string $id = ''): void
     {
         $databaseConnection = $this->getDatabaseConnection();
+        $editLanguage = $this->getLanguageService()->getActiveEditLanguage();
```
UlrichKu: Both initializations are not necessary before the early return.

src/CoreBundle/private/library/classes/TCMSTableEditor/TCMSTableEditorEndPoint.class.php
```diff
@@ -2144,6 +2156,39 @@ protected function getFieldTypeIdBySystemName(string $systemName): ?string
...
+
+        return $databaseConnection->fetchAll($fieldConfigQuery, ['fieldTypeId' => $fieldTypeId]);
+    }
+
+    protected function hasMultiTableRecordReferences(string $tableName, string $fieldName, string $deletedTableName, string $deletedRecordId = ''): bool
+    {
+        $databaseConnection = $this->getDatabaseConnection();
+
+        $relatedRecordsQuery = 'SELECT `id`
+                     FROM
```
UlrichKu: Write these lines in one line? Also above.

(SELECT
   FROM x
  WHERE y ... all aligned)

src/CoreBundle/private/library/classes/TCMSTableEditor/TCMSTableWriter.class.php
```diff
@@ -361,7 +361,7 @@ public function DeleteConnectedRecordReferences()
     {
         parent::DeleteConnectedRecordReferences();
 
-        $this->deleteMultiTableRecordReferencesForDeletedTable($this->oTableConf->sqlData['name']);
+        $this->deleteMultiTableRecordReferencesForDeletedTable($this->_sqlTableName);
```
UlrichKu: This is now different from "cms_tbl_conf"?

src/CoreBundle/private/library/classes/TCMSFields/TCMSField.class.php
pixeljunkie: Not really. getIndexQuery is also called in createIndex, so we don't want to call it twice. This database description export feature is in our way here. Or do i miss something?

src/CoreBundle/private/library/classes/TCMSFields/TCMSFieldExtendedLookupMultiTable.class.php
pixeljunkie: Yes. The index with the _table_name suffix was missing, which in combination with the new introduced related record deletion would lead to slow deletions everywhere. Now that i write this... i think we need a fix update, that ads this index. I will do so. :-(

src/CoreBundle/private/library/classes/TCMSTableEditor/TCMSTableEditorEndPoint.class.php
pixeljunkie: Right. moved.

src/CoreBundle/private/library/classes/TCMSTableEditor/TCMSTableEditorEndPoint.class.php
pixeljunkie: reformatted the queries.

src/CoreBundle/private/library/classes/TCMSTableEditor/TCMSTableWriter.class.php
pixeljunkie: Yes. This is the table, that is deleted at this moment.

src/CoreBundle/private/library/classes/TCMSFields/TCMSFieldExtendedLookupMultiTable.class.php
pixeljunkie: Added an update to add the missing index.

src/CoreBundle/Bridge/Chameleon/Migration/Script/update-1556200994.inc.php
```diff
@@ -0,0 +1,40 @@
...
+                   WHERE `cms_field_conf`.`cms_field_type_id` = :fieldTypeId';
+
+$fieldConfigResult = $databaseConnection->fetchAll($fieldConfigQuery, ['fieldTypeId' => $fieldTypeId]);
+
+foreach ($fieldConfigResult as $row) {
+    $quotedTableName = $databaseConnection->quoteIdentifier($row['tableName']);
+
+    $indexName = $row['fieldName'].TCMSFieldExtendedLookupMultiTable::TABLE_NAME_FIELD_SUFFIX;
+    $indexExistsQuery = 'SHOW INDEX FROM '.$quotedTableName.' WHERE KEY_NAME = '.$databaseConnection->quote($indexName);
+    $indexExistsResult = $databaseConnection->query($indexExistsQuery);
```
UlrichKu: Normally the connection should be able to handle any quoting - also for only ->query(). (?)

This is also a consideration for other (new) locations in this change.

src/CoreBundle/private/library/classes/TCMSFields/TCMSField.class.php
UlrichKu: I am mainly wondering if true === $returnDDL means that no change should be done (only the query returned).

src/CoreBundle/private/library/classes/TCMSTableEditor/TCMSTableEditorEndPoint.class.php
UlrichKu: Doesn't look like it? I meant SELECT x on one line.

src/CoreBundle/private/library/classes/TCMSFields/TCMSField.class.php
pixeljunkie: This is the intension here. Do nothing, just return the SQL that would run.

src/CoreBundle/private/library/classes/TCMSTableEditor/TCMSTableEditorEndPoint.class.php
pixeljunkie: I don't get it. Maybe you can show me? The query is perfectly aligned in the code.

src/CoreBundle/Bridge/Chameleon/Migration/Script/update-1556200994.inc.php
pixeljunkie: Removed the update. Found out, that the combined index is enough to get the fields with the table name without the id fast, because it is the first part of the index.

src/CoreBundle/private/library/classes/TCMSTableEditor/TCMSTableEditorEndPoint.class.php
pixeljunkie: done

PR #341, opened 2019-05-16
332 add core ui based theme with build process

gerritious:
| Q             | A
| ------------- | ---
| Branch        | master
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? |no
| Tests pass?   | unknown
| Fixed issues  | chameleon-system/chameleon-system/332
| License       | MIT

By this pull request a new theme named "coreui" will be added to the CoreBundle. It contains a copy of Chameleon's css and everything necessary to...
1. install CoreUI as a dependency
2. build CourUI using CoreUI's own build process
3. files to store scss variables to override CoreUI's styles
4. add custom scssas imports
5. scripts (`package.json`) to build CoreUI with all overrides and additional scss files

Please make sure to add a new theme named "CoreUI" in The CMS Settings of your Chameleon. Also unhide the `directory` SQL field from the table settings of "CMS Themes" and add "coreui" as a field value in the settings of your newly created theme.


pixeljunkie:
We don't want a new theme... it should be "standard" and we don't want to delete the theme at all. We only wanted a way to build the theme styles with target "/themes/standard/coreui". Now way do we want users to build the backend on setup using npm. This is only intended to update the theme for core developers.
We may allow custom themes later. But this is a different issue.

PR #347, opened 2019-05-20
ref #436: Do not allow html in regular alert messages

UlrichKu:
| Q             | A
| ------------- | ---
| Branch        | 6.2.x
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | maybe - everything (external) that currently relies on alert() showing html
| Deprecations? | no
| Tests pass?   | yes
| Fixed issues  | chameleon-system/chameleon-system#436
| License       | MIT

### DIFF
--- a/src/CoreBundle/Resources/public/javascript/cms.js
+++ b/src/CoreBundle/Resources/public/javascript/cms.js
@@ -87,7 +87,7 @@ function AjaxError(XMLHttpRequest, textStatus, errorThrown) {
                     top.document.location.href = window.location.pathname;
                 }
             } else {
-                alert(errorMessage);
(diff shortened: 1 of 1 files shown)


UlrichKu:
One of the (main?) users here is AjaxError() in cms.js.

But as that also shows things with toasterMessage() it is potentially intentional to use alert() in that certain case.

pixeljunkie:
Looks like the use of alert in AjaxError is a mistake. It should call toasterMessage here, too.
The overwritten alert was only intended to catch system internal alert calls or from third party plugins to show them in a more styled way. So we should change the call in AjaxError to toasterMessage

PR #355, opened 2019-06-24
ref #427: Make getDefaultPortal() more reliable

UlrichKu:
| Q             | A
| ------------- | ---
| Branch        | 6.2.x
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no - slightly: method changes from "first portal" to "primary portal"
| Deprecations? | no
| Tests pass?   | yes
| Fixed issues  | chameleon-system/chameleon-system#427
| License       | MIT

### DIFF
--- a/src/CoreBundle/Service/PortalDomainService.php
+++ b/src/CoreBundle/Service/PortalDomainService.php
@@ -40,10 +40,6 @@ class PortalDomainService implements PortalDomainServiceInterface
      */
     private $domain;
     /**
-     * @var TdbCmsPortal
-     */
-    private $defaultPortal;
-    /**
      * @var array|null
      */
     private $portalDomainNames;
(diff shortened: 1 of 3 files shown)


UlrichKu:
Note this is a slight (and undocumented) interface change: the actual behavior of the method changes.

src/CoreBundle/Service/PortalDomainService.php
```diff
@@ -121,18 +117,7 @@ private function initialize()
...
-
-        $portalList = TdbCmsPortalList::GetList();
-        $portalList->GoToStart();
-        $this->defaultPortal = $portalList->Current();
-        if (false === $this->defaultPortal) {
-            return null;
-        }
-
-        return $this->defaultPortal;
+        return \TdbCmsConfig::GetInstance()->GetPrimaryPortal();
```
rnuernberg: please add an update that checks if this will result in a chance for the customer (that is, if the default portal is now different from the first portal).

src/CoreBundle/Service/PortalDomainServiceInterface.php
```diff
@@ -35,7 +35,7 @@ public function getActiveDomain();
     public function getActivePortal();
 
     /**
-     * Returns the first defined portal.
+     * Returns the primary portal (cms config).
```
rnuernberg: There is a document in which suche changes should be added, is there not?

src/CoreBundle/Service/PortalDomainService.php
UlrichKu: Added an appropriate db update with warning message.

src/CoreBundle/Service/PortalDomainServiceInterface.php
UlrichKu: Not for 6.2...

PR #356, opened 2019-06-24
ref #428: Use Less_Parser directly for prod environment

UlrichKu:
| Q             | A
| ------------- | ---
| Branch        | 6.2.x
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed issues  | chameleon-system/chameleon-system#428
| License       | MIT

### DIFF
--- a/src/ViewRendererBundle/objects/helper/TPkgViewRendererLessCompiler.php
+++ b/src/ViewRendererBundle/objects/helper/TPkgViewRendererLessCompiler.php
@@ -154,7 +154,7 @@ class TPkgViewRendererLessCompiler
      */
     public function getGeneratedCssForPortal($portal, $minifyCss = null)
     {
-        if (false === class_exists('Less_Cache') && false === class_exists('lessc')) {
+        if (false === class_exists('Less_Parser') && false === class_exists('lessc')) {
             throw new ViewRenderException(
                 'You need to install oyejorge/less.php or leafo/lessphp in an appropriate version. See composer.json in chameleon-system/pkgviewrenderer in the suggest section.'
             );
@@ -166,7 +166,7 @@ class TPkgViewRendererLessCompiler
 
         $lessFiles = $this->getLessFilesFromSnippetsAndTheme($portal);
 
-        if (true === class_exists('Less_Cache')) {
+        if (true === class_exists('Less_Parser')) {
             // Workaround (#37218): Remove trailing slash from document root; it will confuse less.php's "path interpretation"
             $originalDocumentRoot = $_SERVER['DOCUMENT_ROOT'];
             $_SERVER['DOCUMENT_ROOT'] = rtrim($_SERVER['DOCUMENT_ROOT'], '/');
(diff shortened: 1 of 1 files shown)


src/ViewRendererBundle/objects/helper/TPkgViewRendererLessCompiler.php
```diff
@@ -191,25 +191,42 @@ public function getGeneratedCssForPortal($portal, $minifyCss = null)
...
+
+                $css = file_get_contents($absoluteCssFilepath);
+            } else {
+                try {
+                    $parser = new \Less_Parser($options);
+                    foreach ($filesForLessParsing as $file => $root) {
+                        $parser->parseFile($file, $root);
+                    }
+
+                    $css = $parser->getCss();
```
rnuernberg: Just to be sure: we only ever get here when the generated css does not exist, correct?

src/ViewRendererBundle/objects/helper/TPkgViewRendererLessCompiler.php
```diff
@@ -191,25 +191,42 @@ public function getGeneratedCssForPortal($portal, $minifyCss = null)
...
-            try {
-                $cssFile = \Less_Cache::Get($filesForLessParsing, $options);
-            } catch (Exception $exc) {
-                if (false !== strpos($exc->getMessage(), 'stat failed')) {
-                    // Consider this as a 'File removed! Halp!' and clear the cache and try again
-                    array_map('unlink', glob($cachedLessDir.'/*'));
+            $css = '';
 
+            if (_DEVELOPMENT_MODE) {
+                \Less_Cache::SetCacheDir($cachedLessDir);
```
rnuernberg: If that is not already the case, it would make a lot of sense to me if the `cachedLessDir` is within the `/app/cache` folder somewhere.

src/ViewRendererBundle/objects/helper/TPkgViewRendererLessCompiler.php
```diff
@@ -166,7 +166,7 @@ public function getGeneratedCssForPortal($portal, $minifyCss = null)
 
         $lessFiles = $this->getLessFilesFromSnippetsAndTheme($portal);
 
-        if (true === class_exists('Less_Cache')) {
+        if (true === class_exists('Less_Parser')) {
```
rnuernberg: Since we already checked if `Less_Parser` exists in line 157, we could skip the check hier. 

If there is reason to believe, that `Less_Cache` does not exist, even though `Less_Parser` does, then we should add the check to line 196.
`if (_DEVELOPMENT_MODE && true === class_exists('Less_Cache')) {`

src/ViewRendererBundle/objects/helper/TPkgViewRendererLessCompiler.php
UlrichKu: If `chameleon_system_core.cache.cache_less_files` is true.

src/ViewRendererBundle/objects/helper/TPkgViewRendererLessCompiler.php
UlrichKu: There is a follow-up issue for this: https://github.com/chameleon-system/chameleon-system/issues/442

src/ViewRendererBundle/objects/helper/TPkgViewRendererLessCompiler.php
UlrichKu: That is actually an old "if-else": if Less_Parser does not exist then "lessc" is used (last line calls getCssFromLessC()).

PR #363, opened 2019-07-17
ref #468: fixes handling of data uris in image urls in CSS/LESS style…

pixeljunkie:
…s to be handled as relative URLs

| Q             | A
| ------------- | ---
| Branch        | 6.2.x
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed issues  | chameleon-system/chameleon-system#468
| License       | MIT

### DIFF
--- a/src/CoreBundle/private/library/classes/TCMSResourceCollection.class.php
+++ b/src/CoreBundle/private/library/classes/TCMSResourceCollection.class.php
@@ -522,7 +522,7 @@ class TCMSResourceCollection implements ResourceCollectorInterface
      */
     public function ReplaceRealPathCallback($aMatch)
     {
-        if (true === $this->isRelativeUrl($aMatch[1])) {
+        if (isset($aMatch[1]) && true === $this->isRelativeUrl($aMatch[1])) {
             $sNewPath = TGlobal::GetStaticURL($this->sCurrentCSSPath.DIRECTORY_SEPARATOR.$aMatch[1]);
             if ('http://' == substr($sNewPath, 0, 7)) {
                 $sNewPath = substr($sNewPath, 5);
--- a/src/CoreBundle/private/library/classes/TCMSResourceCollection.class.php
+++ b/src/CoreBundle/private/library/classes/TCMSResourceCollection.class.php
@@ -522,7 +522,7 @@ class TCMSResourceCollection implements ResourceCollectorInterface
      */
     public function ReplaceRealPathCallback($aMatch)
     {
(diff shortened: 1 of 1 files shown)


src/CoreBundle/private/library/classes/TCMSResourceCollection.class.php
```diff
@@ -522,7 +522,7 @@ protected function ReplaceRelativePath($sAbsoluteCSSFileURL, $sContent)
      */
     public function ReplaceRealPathCallback($aMatch)
     {
-        if (isset($aMatch[1]) && 'http://' != substr($aMatch[1], 0, 7) && 'https://' != substr($aMatch[1], 0, 8) && 0 !== strpos($aMatch[1], '/')) {
+        if (true === $this->isRelativeUrl($aMatch[1])) {
```
UlrichKu: Doesn't check an "isset" anymore?

src/CoreBundle/private/library/classes/TCMSResourceCollection.class.php
```diff
@@ -536,6 +536,23 @@ public function ReplaceRealPathCallback($aMatch)
...
+    {
+        if ('/' === substr($url, 0, 1)) {
+            return false;
+        }
+
+        if ('http://' === substr($url, 0, 7) || 'https://' === substr($url, 0, 8)) {
+            return false;
+        }
+
+        if ('data:' === substr($url, 0, 5)) {
```
UlrichKu: Shouldn't this be filtered earlier?
ie in \TCMSResourceCollection::ReplaceRelativePath() in the regex?

Because "data:" is no url to start with.

UlrichKu:
When looking a the code/regex in ReplaceRelativePath(): would this recognize a "url (..." with a space?

src/CoreBundle/private/library/classes/TCMSResourceCollection.class.php
pixeljunkie: Changing the regex may have side effects. I'm not a regex guru. But yes, in theory you could handle all of this in the regex already. If you know how exactly and be sure this doesn't have side effects.

src/CoreBundle/private/library/classes/TCMSResourceCollection.class.php
pixeljunkie: You are right. Added it again.

pixeljunkie:
We need a UnitTest here. Will add one.

UlrichKu:
> We need a UnitTest here. Will add one.

Yes.
But the TCMS will probaly not work well with a unit test...

pixeljunkie:
Yes, impossible without refactoring.

PR #365, opened 2019-07-22
ref #465: Remove doubled html escape from file name output

UlrichKu:
| Q             | A
| ------------- | ---
| Branch        | 6.3.x
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed issues  | chameleon-system/chameleon-system#465
| License       | MIT


UlrichKu:
A larger change (not using |raw and not quoting it already in code) would need to wait for the next minor version.

PR #370, opened 2019-08-26
ref #464: Place "GoToStart()" correctly

UlrichKu:
| Q             | A
| ------------- | ---
| Branch        | 6.3.x
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed issues  | chameleon-system/chameleon-system#464
| License       | MIT


bestform:
Just for the record: Please do never merge your own code, even if time is pressing. Set the phone line on fire if it must be reviewed yesterday, but never press the merge button on your own code. If it would be enough for a PR to pass that the developer opening it thinks it is ok we would not need PRs in the first place. ;)

And what's up with all those references to this issue in the commits by @ckleis?

PR #371, opened 2019-08-27
476/update and modify select2

gerritious:
| Q             | A
| ------------- | ---
| Branch        | master
| Bug fix?      | yes
| New feature?  | no 
| BC breaks?    | no     6.2.12 / 6.3.5 / 7.0.4
| Deprecations? | no 
| Tests pass?   | yes    
| Fixed issues  | chameleon-system/chameleon-system#476   
| License       | MIT

**Also solves**: https://github.com/chameleon-system/chameleon-system/issues/378

1. Update select2
2. Create sym link to current select2 version
3. Change include paths of select2 sources to version agnostic ones
4. Modify select2 event handling


UlrichKu:
There are now two directories (4.0.6 and 4.0.9)?

I would simply replace the files in select2.v4.

The new version is still "v4" so that is ok.

src/CoreBundle/Resources/public/javascript/table.js
```diff
@@ -181,52 +181,96 @@ $(document).ready(function () {
...
-        // input as `newOption`.
-
-        // Known drawback: The search field associated to the current select2 widget can not be directly determined as
-        // select2 does not expose any way to retrieve it. Therefore the DOM will be queried for the
-        // search field's selector.
-        // The query will return the first search field found in the DOM which might not be the desired one.
-        // Yet as select2 creates a new search field on activating the widget and removes it when unnecessary there
-        // should be only one search field present in the DOM.
+    }).on('select2:closing', function(e) {
+         /* There is no handler attached to the `select` event. Instead the `closing` handler will also manage the
```
UlrichKu: Better describe what should be achieved here (and how). Like:

Only monitoring closing here because this simplifies X

src/CoreBundle/Resources/public/javascript/table.js
```diff
@@ -181,52 +181,96 @@ $(document).ready(function () {
...
+        const data = originalSelect2EventData.data;
+
+        if ('undefined' === typeof data) {
+            return;
+        }
+
+        /* The following condition validates to `true` when no selection has been made from select2's suggestions.
+        *  In this case a hidden form field with the search term as value will be added to the form.
+        * */
+        if (true === data.newOption) {
```
UlrichKu: Or better yet: maybe use a helper method here with an understandable name and then this and the comment above can be removed.

-> isSimpleSearch() or similar

Also below for the block with createSearchWord...

-> searchForText() ..

src/CoreBundle/Resources/public/javascript/table.js
```diff
@@ -181,52 +181,96 @@ $(document).ready(function () {
...
+        select2SearchField.value = text;
+        select2SearchField.select();
+
+        $(select2SearchField).off('input');
 
-        if (null === select2SearchField || '' === select2SearchField.value ) {
+        if ('1' === this.dataset.keyUpListenerAttached) {
             return;
         }
+        select2SearchField.addEventListener('keyup', (e) => {
```
UlrichKu: keyup is rather special, did you mean keypress?

src/CoreBundle/Resources/public/javascript/table.js
```diff
@@ -181,52 +181,96 @@ $(document).ready(function () {
...
+            selectSelectionRendered.title = select2SearchField.value;
+            selectSelectionRendered.textContent = select2SearchField.value;
+        });
 
-        createSearchWordInputWithValue(select2SearchField.value);
+        this.dataset.keyUpListenerAttached = '1';
     });
 
-    function switchRecord(id) {
+    function getSibling(node, selector) {
```
UlrichKu: Doesn't jquery do this with sibling() or similar?

src/CoreBundle/Resources/public/javascript/table.js
```diff
@@ -181,52 +181,96 @@ $(document).ready(function () {
...
+        return getSibling(sibling, selector);
+    }
+
+    function switchRecord(id, searchTerm = '') {
         if ('' !== id) {
-            var url = searchLookup.data('record-url') + '&id=' + id;
-            top.document.location.href = url;
+            const searchTermPart = '' !== searchTerm ? `&_search_word=${searchTerm}` : '';
+            /* Also submit the search term (the selected suggestion of select2)
+             * so it can be added to the session as last search term.
```
UlrichKu: When is it added to the session?

gerritious:
> There are now two directories (4.0.6 and 4.0.9)?
> 
> I would simply replace the files in select2.v4.
> 
> The new version is still "v4" so that is ok.

I would rather like go version agnostic, meaning without a version suffix in include paths. But I agree that sym linking to a specific version is non standard in this part of the application (or anywhere in Chameleon).
Should I revert the changes I've made to the include paths? You decide. :)

The older version can be deleted. I've just left it there to make reverting back to it (if necessary) easier.

src/CoreBundle/Resources/public/javascript/table.js
gerritious: Gefällt mir. Done.

UlrichKu:
> Should I revert the changes I've made to the include paths? You decide. :)

Yes, please.
Also mainly to declutter this PR. Otherwise the important changes are not easily recognizable/findable.

For reverting to older versions (4.0.6) we have the git history.

src/CoreBundle/Resources/public/javascript/table.js
gerritious: Done.

src/CoreBundle/Resources/public/javascript/table.js
gerritious: Would you please point me to what makes 'keyup' special as an event?
Or do you mean 'specific' rather than 'special'? I think it is advisable to be as specific as possible.

src/CoreBundle/Resources/public/javascript/table.js
gerritious: It does. Thanks.

gerritious:
There has been a question regarding the session, which I can not find here in the discussion (sorry). Therefore I would like to answer it here:
The JavaScript code does not write anything to the session. It merely allows to do just that by submitting the slected option as a search term. I have edited the comment regarding that bit of code.I hope it is clearer now.

src/CoreBundle/Resources/public/javascript/table.js
gerritious: `keypress` is indeed better here, because it is treiggered less often than `keyup`. For example `keyup` will be triggered when pressing the tab key whereas `keypress` won't.
In short `kepress` is triggered when character like keys are being pressed (with some exceptions).

src/CoreBundle/Resources/public/javascript/table.js
```diff
@@ -181,52 +183,150 @@ $(document).ready(function () {
...
+    }
+
+    function switchRecord(id, searchTerm = '') {
         if ('' !== id) {
-            var url = searchLookup.data('record-url') + '&id=' + id;
-            top.document.location.href = url;
+            const searchTermPart = '' !== searchTerm ? `&_search_word=${searchTerm}` : '';
+            /* Here the URL will be set to make the browser load the given detail page's URL. The search form will not
+             * be submitted. Therefore the search term stored in the session will not be changed.
+             * In some cases it may be desirable to change the session stored search term to the selection value of the
```
UlrichKu: Where is this then decided (if the session search term is changed)?

src/CoreBundle/Resources/public/javascript/table.js
```diff
@@ -181,52 +183,150 @@ $(document).ready(function () {
...
         }
-    }).on('select2:select', function (e) {
-        if (e.params.data.newOption) {
-            var listname = searchLookup.data('listname');
-            document[listname]._startRecord.value = 0;
-            document[listname].submit();
+    }).on('select2:clear', function(e) {
+        const form = createSearchWordInputWithValue();
+
+        uiEnabled = false;
```
UlrichKu: uiEnabled (only) means "clear requested"?

src/CoreBundle/Resources/public/javascript/table.js
gerritious: It cannot be decided by JavaScript code. Whatever goes to the session must be desided by the controller logic.

Some background on why I think it advantageous to submit the selectes option here:
There a two ways a User can interact with the select2 widget:

1. The user enters a search term and presses enter (to submit)
2. The user enters a search term and selects one of the options suggested by the select2 widget

Regarding 1: A search form will be submitted by JS and the search term can be stored in the session
Regarding 2: The browsers URL will be changed to that of the detail page of the selected option. In this case the search form will not be submitted. The session stored search term will remain as is.

The Problem:
Say on the product list page a user types "bracelet' into the search input field and presses the enter key. The search term "bracelet" will be stored in the session (and a result list will be displayed).
Next the user types "ocean" into the search input field and selects one of the suggested options. The user's browser will load the detail page of the selected option. The search term will remain "bracelet".
Next the user navigates back to the product list page. The search term "bracelet" will be prefilled in the search input field (from the session).
I think the expected behaviour would be to display either "ocean" (which is what the user last typed) or the complete title of the sleected option (i.e. "Armband Ocean - 17 cm"). 
To make this possible the JavaScrip code adds the complete title of the selected option as a GET parameter to the detail page's URL with the key `_search_word`. subsequently, the controller that handles the resulting request can decide whether or not to store the parameters value to the session. 
As yet there is no function triggered by submitting a `_search_word` parameter by JavaScript. I would like to suggest to cahnge the detail page controller accordingly. I will open a new ticket to propose that.

src/CoreBundle/Resources/public/javascript/table.js
gerritious: Yes, kind of (as long as blocking th ui isn't necessary aywhere else). Do you think it better to change the variable name?

src/CoreBundle/Resources/public/javascript/table.js
UlrichKu: If this is not used I would (only) do this code change within the new ticket.

src/CoreBundle/Resources/public/javascript/table.js
UlrichKu: Yes, please.

gerritious: Done.

src/CoreBundle/Resources/public/javascript/table.js
gerritious: Done.

src/CoreBundle/Resources/public/javascript/table.js
```diff
@@ -149,6 +149,8 @@ $(document).ready(function () {
 
     var searchLookup = $('select#searchLookup');
 
+    let searchInputFieldCleared = true;
```
UlrichKu: Is "input field cleared" the default / main usage case?

I'd expect the values to be swapped: set it to true in select2:clear below.

gerritious: Yes, you're right. Changed.

PR #373, opened 2019-08-28
ref #441: align content with label

ckleis:
| Q             | A
| ------------- | ---
| Branch        | features / 6.3.x
| Bug fix?      | yes
| New feature?  | yes/no 
| BC breaks?    | no  
| Deprecations? | no 
| Tests pass?   | yes    
| Fixed issues  | chameleon-system/chameleon-system#441
| License       | MIT

### DIFF
--- a/src/CoreBundle/private/library/classes/TCMSFields/TCMSFieldLookupDirectory.class.php
+++ b/src/CoreBundle/private/library/classes/TCMSFields/TCMSFieldLookupDirectory.class.php
@@ -116,7 +116,7 @@ class TCMSFieldLookupDirectory extends TCMSField
     {
         $this->GetOptions();
         if (array_key_exists($this->data, $this->options)) {
-            return $this->_GetHiddenField().TGlobal::OutHTML($this->options[$this->data]);
+            return $this->_GetHiddenField().'<div class="col-form-label">'.$this->options[$this->data].'</div>';
         } else {
             return $this->_GetHiddenField();
         }
--- a/src/CoreBundle/private/library/classes/TCMSFields/TCMSFieldLookupDirectory.class.php
+++ b/src/CoreBundle/private/library/classes/TCMSFields/TCMSFieldLookupDirectory.class.php
@@ -116,7 +116,7 @@ class TCMSFieldLookupDirectory extends TCMSField
     {
         $this->GetOptions();
         if (array_key_exists($this->data, $this->options)) {
-            return $this->_GetHiddenField().'<div class="form-content-simple">'.$this->options[$this->data].'</div>';
+            return $this->_GetHiddenField().'<div class="form-content-simple">'.TGlobal::OutHTML($this->options[$this->data]).'</div>';
         } else {
             return $this->_GetHiddenField();
         }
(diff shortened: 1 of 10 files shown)


UlrichKu:
Looks a bit strange/redundant but this is ok as the problem of "aligning the first line of text of arbitrary content" is not really solvable.

Small changes:

I would match the label (col-form-label) to the content on the right: if there is padding (ie for "p-1") this is normally ".25rem".

So: 
`.tableeditcontainer .col-form-label {
    /* matches most content views on the right */

    padding-top: .25rem;
    padding-bottom: .25rem;
}`

And for the contents I would use another class like "form-content-simple" and give that the same padding.

src/CoreBundle/private/library/classes/TCMSFields/TCMSFieldLookupDirectory.class.php
```diff
@@ -116,7 +116,7 @@ public function GetReadOnly()
     {
         $this->GetOptions();
         if (array_key_exists($this->data, $this->options)) {
-            return $this->_GetHiddenField().TGlobal::OutHTML($this->options[$this->data]);
```
UlrichKu: The OutHTML should be kept (also below).

src/CoreBundle/private/modules/MTTableEditor/views/includes/fields.inc.php
```diff
@@ -41,7 +41,7 @@
 
                 $sTmpFormTabsContent .= '
                 <div class="form-group table-editor-form-group row" id="fieldname_'.TGlobal::OutHTML($oField->name).'">
-                    <div class="col-sm-2 col-form-label">
```
UlrichKu: This padding is the only thing that "col-form-label" does?

src/CoreBundle/private/library/classes/TCMSFields/TCMSFieldLookupDirectory.class.php
ckleis: I don't know how it happened, but of course I added it again.

src/CoreBundle/private/modules/MTTableEditor/views/includes/fields.inc.php
ckleis: I used the bootstrap class "col-form-lable" again and overwrote the paddings so it aligns with the standard-fields.

src/CoreBundle/private/library/classes/TCMSFields/TCMSFieldOption.class.php
```diff
@@ -94,7 +94,7 @@ public function GetReadOnly()
     {
         $this->GetOptions();
         if (array_key_exists($this->data, $this->options)) {
-            return $this->_GetHiddenField().TGlobal::OutHTML($this->options[$this->data]);
+            return $this->_GetHiddenField().'<div class="form-content-simple">'.$this->options[$this->data].'</div>';
```
UlrichKu: OutHTML here still missing?

ckleis: also added here

PR #381, opened 2019-09-19
488 use typeahead

UlrichKu:
| Q             | A
| ------------- | ---
| Branch        | 6.3.x
| Bug fix?      | yes
| New feature?  | yes - downsizing one
| BC breaks?    | no
| Deprecations? | yes
| Tests pass?   | yes
| Fixed issues  | chameleon-system/chameleon-system#488
| License       | MIT


pixeljunkie:
Tested the branch and found 2 issues:
- In the template engine the typeahead script include is missing
- I get an error in the console with the first xhr call: Unchecked runtime.lastError: Could not establish connection. Receiving end does not exist.

src/CoreBundle/Resources/public/javascript/cms.js
```diff
@@ -517,3 +517,47 @@ CHAMELEON.CORE.handleFormAndLinkTargetsInModals = function () {
...
+                if ("" !== $element.data('record-url')) {
+                    top.document.location.href = $element.data('record-url') + '&id=' + item.id;
+                }
+            }
+        },
+        autoSelect: false,
+        minLength: 0,
+        showHintOnFocus: true,
+        items: $(window).height() > 600 ? 16 : 8,
+        menu: '<ul class="typeahead dropdown-menu dropdown-menu-right" role="listbox"></ul>X',
```
pixeljunkie: There is an X behind the UL? Is this needed?

pixeljunkie: I would set the dropdown to the left.

src/CoreBundle/private/modules/MTTableManager/MTTableManager.class.php
```diff
@@ -530,6 +542,7 @@ public function GetHtmlHeadIncludes()
         $includeLines = parent::GetHtmlHeadIncludes();
         $includeLines[] = '<script src="'.TGlobal::GetStaticURLToWebLib('/javascript/jquery/flash/flash.js').'" type="text/javascript"></script>';
         $includeLines[] = '<link href="'.TGlobal::GetPathTheme().'/css/table.css" rel="stylesheet" type="text/css" />';
+        $includeLines[] = '<script src="'.TGlobal::GetStaticURLToWebLib('/components/bootstrap3-typeahead/bootstrap3-typeahead.min.js').'" type="text/javascript"></script>';
```
pixeljunkie: This is missing in the template engine includes.

src/CoreBundle/Resources/public/javascript/cms.js
UlrichKu: X removed

left: Normally the input field is on the right of the page. And positioning the dropdown right means that it will open to the left - otherwise it will often be outside of the screen on the right.

pixeljunkie: ok.

src/CoreBundle/private/modules/MTTableManager/MTTableManager.class.php
UlrichKu: Moved it to the footer include file.

pixeljunkie:
Looks like the error: "Unchecked runtime.lastError: Could not establish connection. Receiving end does not exist." has it's cause in a chrome plugin. So ignore this here.

PR #382, opened 2019-10-07
ref #340: add debug output flag to cache parameters of page meta module

pixeljunkie:
| Q             | A
| ------------- | ---
| Branch        | 6.3.x
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed issues  | chameleon-system/chameleon-system#340
| License       | MIT

### DIFF
--- a/src/CoreBundle/private/library/classes/WebModules/MTPageMetaCore/MTPageMetaCoreEndPoint.class.php
+++ b/src/CoreBundle/private/library/classes/WebModules/MTPageMetaCore/MTPageMetaCoreEndPoint.class.php
@@ -503,11 +503,6 @@ class MTPageMetaCoreEndPoint extends TUserModelBase
      */
     public function _AllowCache()
     {
-        // if user is logged in to CMS the cache will be disabled to allow output of CMS only content
-        if (TGlobal::CMSUserDefined()) {
-            return false;
-        }
-
         return null !== $this->getPortalDomainService()->getActivePortal();
     }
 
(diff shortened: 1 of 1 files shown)


src/CoreBundle/private/library/classes/WebModules/MTPageMetaCore/MTPageMetaCoreEndPoint.class.php
```diff
@@ -516,13 +511,17 @@ public function _AllowCache()
      */
     public function _GetCacheParameters()
     {
-        $aParameters = parent::_GetCacheParameters();
+        $cacheParameters = parent::_GetCacheParameters();
+
+        if (_DEVELOPMENT_MODE && TCMSUser::CMSUserDefined()) {
+            $cacheParameters['debugOutputActive'] = true;
```
UlrichKu: The output here is (still) different between backend logged in and normal?

(So this parameter is necessary?)

pixeljunkie: Yes. We don't want the page ID in the frontend. This would be an infosec issue. Not that this is a full secret, but it is more secure to not show this info anyone... but... we could remove the dev mode flag here. Would mean, that any backend user would be able to get the page id from the source code and we have 2 cache elements. Would mean doubled cache elements even in live mode, but as long as you don't click through the full website causing masses of cache elements this should not be a problem at all. And it is not user specific, so the cache element is for alle backend users the same. Better than like now, not having any cache in dev environments and not having the info in the live system at all. So i will remove the check for dev mode here.

UlrichKu:
Superceeded by other PR (385)

PR #400, opened 2019-11-25
504 search mysql table name

UlrichKu:
| Q             | A
| ------------- | ---
| Branch        | 6.3.x
| Bug fix?      | no - refactoring only
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed issues  | chameleon-system/chameleon-system#...
| License       | MIT

This is only a refactoring - it shows that this issue is not easily fixable.

### DIFF
--- a/src/CoreBundle/private/modules/MTTableManager/MTTableManager.class.php
+++ b/src/CoreBundle/private/modules/MTTableManager/MTTableManager.class.php
@@ -11,6 +11,7 @@
 
 use ChameleonSystem\CoreBundle\Service\BackendBreadcrumbServiceInterface;
 use ChameleonSystem\CoreBundle\Service\LanguageServiceInterface;
+use ChameleonSystem\CoreBundle\Util\FieldTranslationUtil;
 use Doctrine\DBAL\Connection;
 use ChameleonSystem\CoreBundle\Interfaces\FlashMessageServiceInterface;
 use ChameleonSystem\CoreBundle\ServiceLocator;
@@ -477,8 +478,7 @@ class MTTableManager extends TCMSModelBase
 
     private function getAutocompleteListQuery(): string
     {
-        $inputFilterUtil = $this->getInputFilterUtil();
-        $searchKey = $inputFilterUtil->getFilteredInput('term');
+        $databaseConnection = $this->getDatabaseConnection();
 
         $listClass = null;
         // allow custom list class overwriting (defined in pagedef)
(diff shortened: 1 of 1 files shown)


src/CoreBundle/private/modules/MTTableManager/MTTableManager.class.php
```diff
@@ -477,56 +478,67 @@ protected function getAutocompleteRecords(): array
...
-        $databaseConnection = $this->getDatabaseConnection();
-        $quotedNameColumn = $databaseConnection->quoteIdentifier($nameColumn);
+        $searchTerm = '%'.$this->getInputFilterUtil()->getFilteredInput('term').'%';
+        $quotedSearchTerm = $databaseConnection->quote($searchTerm);
 
-        $quoteCharacter = $databaseConnection->getDatabasePlatform()->getIdentifierQuoteCharacter();
-        if (false === strpos($parentTableAlias, $quoteCharacter)) {
-            $quotedParentTableAlias = $databaseConnection->quoteIdentifier($parentTableAlias);
+        if (false !== stripos($query, 'WHERE')) {
+            // TODO / NOTE if $query contains an OR (without brackets) then this here will make the query wrong.
```
pixeljunkie: So, this is an open issue that needs to be addressed? Any idea, how?

src/CoreBundle/private/modules/MTTableManager/MTTableManager.class.php
```diff
@@ -477,56 +478,67 @@ protected function getAutocompleteRecords(): array
...
+        $databaseConnection = $this->getDatabaseConnection();
 
         $listClass = null;
         // allow custom list class overwriting (defined in pagedef)
         if (array_key_exists('listClass', $this->aModuleConfig)) {
             $listClass = $this->aModuleConfig['listClass'];
         }
 
-        $oTableList = &$this->oTableConf->GetListObject($listClass);
+        $listManager = &$this->oTableConf->GetListObject($listClass);
```
pixeljunkie: The & is not necessary anymore.

src/CoreBundle/private/modules/MTTableManager/MTTableManager.class.php
pixeljunkie: No i see it. It was always a problem here. Because it is only a refactoring here we don`t need to address this issue at this time.

PR #406, opened 2019-12-06
ref #520: Check for an onclick handler on the relevant table row/cell

UlrichKu:
| Q             | A
| ------------- | ---
| Branch        | 6.3.x
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed issues  | chameleon-system/chameleon-system#520
| License       | MIT

### DIFF
--- a/src/CoreBundle/Resources/public/javascript/cms.js
+++ b/src/CoreBundle/Resources/public/javascript/cms.js
@@ -527,7 +527,19 @@ CHAMELEON.CORE.initializeEntryAutocomplete = function($element) {
         },
         afterSelect: function(item) {
             if (typeof item.id !== "undefined") {
-                if ("" !== $element.data('record-url')) {
+                var handled = false;
+                if ("" !== item.id) {
+                    // Look for a table row for this entry and its click handler:
+
+                    var $row = $("tr[data-record-id='"+item.id+"']");
+                    var $onClicks = $row.find("td[onClick]");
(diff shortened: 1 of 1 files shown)


src/CoreBundle/Resources/public/javascript/cms.js
```diff
@@ -527,7 +527,19 @@ CHAMELEON.CORE.initializeEntryAutocomplete = function($element) {
...
+
+                    var $row = $("tr[data-record-id='"+item.id+"']");
+                    var $onClicks = $row.find("td[onClick]");
+                    if ($row.length > 0 && $onClicks.length > 0) {
+                        $($onClicks[0]).click();
+                        handled = true;
+                    }
+                }
+                if (false === handled && "" !== $element.data('record-url')) {
+                    // TODO could take the href of the above row td or simply click there?
```
pixeljunkie: I would use the url so other events are not triggered here. Or should they? I'm not sure. We don't have other events on cells at the moment, but what about inline editing later maybe?

UlrichKu: Or on-clicks being not defined as "onclick" (but with a proper listener)?

UlrichKu: Just removed the TODO for now:

The existing "record-url" function of an autocomplete field is preserved.

All other functions can be added/changed when needed.

PR #411, opened 2019-12-12
514 enable page delete

UlrichKu:
| Q             | A
| ------------- | ---
| Branch        | 6.3.x
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed issues  | chameleon-system/chameleon-system#514
| License       | MIT

### DIFF
--- a/src/CoreBundle/private/modules/CMSTemplateEngine/CMSTemplateEngine.class.php
+++ b/src/CoreBundle/private/modules/CMSTemplateEngine/CMSTemplateEngine.class.php
@@ -111,28 +111,22 @@ class CMSTemplateEngine extends TCMSModelBase
         parent::Init();
 
         $this->sPageId = $this->getPageIdFromRequest();
+        $this->oPage = TCMSPagedef::GetCachedInstance($this->sPageId);
 
-        $page = $this->getPageService()->getById($this->sPageId);
-
-        if (null === $page) {
+        if (false === $this->oPage->sqlData) {
             throw new NotFoundHttpException(sprintf('A page with the id %s cannot be found.', $this->sPageId));
         }
 
         $this->oTableManager = new TCMSTableEditorManager();
         $this->sTableID = TTools::GetCMSTableId('cms_tpl_page');
         $this->oTableManager->Init($this->sTableID, $this->sPageId);
-
-        //$this->oPage = $this->getPageDefAccess()->get($this->sPageId);
-        // TODO above is wrong (type) and $page cannot be used here either?
-        $this->oPage = TCMSPagedef::GetCachedInstance($this->sPageId);
-
         if ($this->global->UserDataExists('_mode')) {
             $this->sMode = $this->global->GetUserData('_mode');
         }
         $this->bPageDefinitionAssigned = (!empty($this->oPage->iMasterPageDefId));
         $this->AddURLHistory();
 
-        $this->oPortal = $page->GetPortal();
+        $this->oPortal = $this->oPage->GetPortal();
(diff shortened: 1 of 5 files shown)


src/CoreBundle/private/modules/CMSTemplateEngine/CMSTemplateEngine.class.php
```diff
@@ -108,18 +111,29 @@ public function Init()
         parent::Init();
 
         $this->sPageId = $this->getPageIdFromRequest();
+
+        // TODO what about language?
+        $page = $this->getPageService()->getById($this->sPageId);
+
+        if (null === $page) {
+            throw new NotFoundHttpException(sprintf('A page with the id %s cannot be found.', $this->sPageId));
```
UlrichKu: As there is (probably) no notion of "404" in the backend this is also problematic.

src/CoreBundle/private/modules/CMSTemplateEngine/CMSTemplateEngine.class.php
```diff
@@ -108,18 +111,29 @@ public function Init()
         parent::Init();
 
         $this->sPageId = $this->getPageIdFromRequest();
+
+        // TODO what about language?
```
pixeljunkie: language is not relevant here for loading because we do not use the objects data, default language is o.k.

src/CoreBundle/private/modules/CMSTemplateEngine/CMSTemplateEngine.class.php
pixeljunkie: 404 is o.k. here, but we need a new issue about adding 404 handling to the backend controller.

src/CoreBundle/private/modules/CMSTemplateEngine/CMSTemplateEngine.class.php
```diff
@@ -108,18 +111,29 @@ public function Init()
...
+
+        if (null === $page) {
+            throw new NotFoundHttpException(sprintf('A page with the id %s cannot be found.', $this->sPageId));
+        }
+
         $this->oTableManager = new TCMSTableEditorManager();
         $this->sTableID = TTools::GetCMSTableId('cms_tpl_page');
         $this->oTableManager->Init($this->sTableID, $this->sPageId);
+
+        //$this->oPage = $this->getPageDefAccess()->get($this->sPageId);
```
pixeljunkie: Would not change this here. If so, it needs to be tested in deep. I'm not into the details of the difference between this new service and the GetCachedInstance method. But i would like to see this legacy call deprecated. So feel free to check if the new service is right here.

src/CoreBundle/private/modules/CMSTemplateEngine/CMSTemplateEngine.class.php
UlrichKu: Entere #527 for 404.

src/CoreBundle/private/modules/CMSTemplateEngine/CMSTemplateEngine.class.php
UlrichKu: ok

src/CoreBundle/private/modules/CMSTemplateEngine/CMSTemplateEngine.class.php
UlrichKu: Changed: only using $this->oPage now for everything.

PR #412, opened 2019-12-12
516 correct table editor iframes

UlrichKu:
| Q             | A
| ------------- | ---
| Branch        | 6.3.x
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no - maybe: there is a risk of side-effects
| Deprecations? | no
| Tests pass?   | yes
| Fixed issues  | chameleon-system/chameleon-system#516
| License       | MIT

### DIFF
--- a/src/CoreBundle/Resources/public/themes/standard/css/global.css
+++ b/src/CoreBundle/Resources/public/themes/standard/css/global.css
@@ -13,6 +13,30 @@ body.app {
     min-height: 100%;
 }
 
+#cmscontainer .main {
+    display: flex;
+    flex-direction: column;
+}
+
+#cmscontainer .main .content {
+    flex: 1;
+    display: flex;
+    flex-direction: column;
+}
+
+#cmscontainer .main .content #tableEditorContainer {
+    flex: 1;
+}
+
+#cmsbreadcrumb .breadcrumb {
+    border-radius: 0;
+    margin-bottom: 0; /* overwrite bootstrap */
+}
+
+#cmsbreadcrumb .breadcrumb i {
+    color: #098dba;
+}
+
 a:hover {
     text-decoration: underline;
 }
@@ -95,7 +119,7 @@ legend {
 }
 
 .button-element .button-item {
-    padding: 0 3px 6px 3px;
+    padding: 6px 3px;
 }
 
 .typeahead-relative {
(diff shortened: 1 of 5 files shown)


src/CoreBundle/Resources/public/themes/standard/css/global.css
```diff
@@ -5,22 +5,68 @@ html {
...
+#cmsbreadcrumb .breadcrumb i {
+    color: #098dba;
+}
+
 a:hover {
     text-decoration: underline;
 }
 
 iframe {
+    width: 100%;
```
pixeljunkie: I would not set every iframe to the exact same size. This would lead to sideeffects with custom modules.
Could you set this to an extra CSS class or define it more precise like only iframes inside of an table editor container?

src/CoreBundle/Resources/public/themes/standard/css/global.css
```diff
@@ -5,22 +5,68 @@ html {
...
+    height: 470px; /* */
```
pixeljunkie: empty comment?

src/CoreBundle/Resources/public/themes/standard/css/global.css
```diff
@@ -5,22 +5,68 @@ html {
 }
 
 body {
-    height: 100%;
+    overflow-x: hidden;
```
UlrichKu: That would be _some_ solution for surplus horizontal scrollbars.

But will be removed here.

src/CoreBundle/Resources/public/themes/standard/css/global.css
UlrichKu: This only means that it will fill the containing element.

(And can/would be overwritten by more specific css if desired.)

pixeljunkie: The 100% is o.k. it is more the height of 470px that makes me nervous.

src/CoreBundle/Resources/public/themes/standard/css/global.css
UlrichKu: Removed

src/CoreBundle/Resources/public/themes/standard/css/global.css
pixeljunkie: We will try if this height makes any problems.

PR #420, opened 2020-01-23
Fix/539 improve no portal handling

rnuernberg:
| Q             | A
| ------------- | ---
| Branch        | 7.0.x
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed issues  | chameleon-system/chameleon-system#539
| License       | MIT

When no matching domain is found for a request (because no domain exists, or the domain is disabled) a notice about accessing a property of a none object will be thrown.

This makes it more difficult to finde the actual error - and can be easily corrected.


UlrichKu:
Not necessary with https://github.com/chameleon-system/chameleon-base/pull/421 (will be merged up).

PR #428, opened 2020-02-05
552 highlight current entry

UlrichKu:
| Q             | A
| ------------- | ---
| Branch        | 6.3.x
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed issues  | chameleon-system/chameleon-system#552
| License       | MIT

### DIFF
--- a/src/CoreBundle/Resources/public/javascript/modules/sidebar/sidebar.js
+++ b/src/CoreBundle/Resources/public/javascript/modules/sidebar/sidebar.js
@@ -45,18 +45,23 @@
         },
         markSelected: function() {
             var currentTableId = this.extractTableId(document.location.href);
-
-            if (null === currentTableId) {
-                return;
-            }
+            var documentPathAndSearch = document.location.pathname + document.location.search;
 
             var outer = this;
 
             this.$navItems.not(".nav-dropdown").each(function() {
                 var link = $(this).find("a").attr("href");
+                var linkPathAndSearch = outer.getPathAndSearch(link);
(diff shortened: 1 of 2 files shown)


src/CoreBundle/Resources/public/javascript/modules/sidebar/sidebar.js
```diff
@@ -201,6 +220,32 @@
             $.post(url, {
                 clickedMenuId: clickedMenuId
             });
+        },
+        extractTableId: function(url) {
+            // TODO use some proper url splitting? ie https://gomakethings.com/getting-all-query-string-values-from-a-url-with-vanilla-js/
```
pixeljunkie: There is a lot easier way in all browsers except IE11 (which would be o.k.): https://caniuse.com/#search=URLSearchParams
https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams
https://www.sitepoint.com/get-url-parameters-with-javascript/

We should iterate over all parameters and check for keys not only with the exact pagedef name, but allow tablemanager* and tableeditor* because there may be some extensions in customer projects or even in the core using a different tablemanager.

UlrichKu: Ok, using URLSearchParams. The substring search for parameters is already present.

However there are still other cases not covered. A direct comparisson of links should get them though.

PR #452, opened 2020-04-21
Symfony44 upgrade

bestform:
| Q             | A
| ------------- | ---
| Branch        | master
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | yes
| Deprecations? | no 
| Tests pass?   | yes  
| Fixed issues  | -
| License       | MIT

Upgrade to Symfony 4.4

Projects that update to this version might need to make those services explicitly public that are being fetched from the container at runtime via the service locator.


src/CoreBundle/Session/ChameleonSessionManager.php
```diff
@@ -125,10 +125,6 @@ public function boot()
         // add decorator to enforce write order (old data may not overwrite new data)
         $sessionHandler = new TPkgCmsSessionHandler_Decorator_EnforceWriteSequence($sessionStorage->getSaveHandler());
 
-        if ($this->metaDataTimeout > 0) {
-            $sessionHandler = new WriteCheckSessionHandler($sessionHandler);
```
UlrichKu, line 129: What was this doing (why can it be removed)?

composer.json
```diff
@@ -29,13 +29,11 @@
...
-        "sensio/distribution-bundle": "^5.0.19",
-        "symfony/monolog-bundle": "^3.1.0",
-        "symfony/polyfill-apcu": "^1.0",
-        "symfony/symfony": "~3.4.35",
-        "symfony-cmf/routing-bundle": "~2.1.0",
-        "tedivm/jshrink": "~1.0",
-        "twig/twig": "~2.8"
+        "symfony/monolog-bundle": "3.5.0",
+        "symfony/polyfill-apcu": "1.13.1",
+        "symfony/symfony": "4.4.4",
```
UlrichKu, line 34: Should this version be fixed?

src/NewsletterBundle/Resources/config/services.xml
```diff
@@ -21,7 +21,7 @@
             <argument type="service" id="chameleon_system_newsletter.post_processor_collector" />
         </service>
 
-        <service id="chameleon_system_newsletter.user_data_factory" class="ChameleonSystem\NewsletterBundle\PostProcessing\Bridge\NewsletterUserDataFactory">
+        <service id="chameleon_system_newsletter.user_data_factory" class="ChameleonSystem\NewsletterBundle\PostProcessing\Bridge\NewsletterUserDataFactory" public="true">
         </service>
 
         <service id="chameleon_system_newsletter.mapper.signout_config" class="TPkgNewsletterMapper_PkgNewsletterModuleSingOutConfig">
```
UlrichKu, line 27: These 4 also need to be public.

UlrichKu: Comment will be obsolete with my pull request https://github.com/bestform/chameleon-base/pull/1

src/CoreBundle/Session/ChameleonSessionManager.php
bestform: It avoided writing the session in case it wasn't changed. It now uses a core functionality in PHP. See https://symfony.com/blog/new-in-symfony-3-4-session-improvements#safer-and-lazier-sessions

composer.json
bestform: This was done intentionally because at least on my machine I couldn't get composer to resolve the massive dependency graph using less strict versions. We should address this in a future update, but it is not necessarily bad practice to define exact versions - if you take care of them from time to time.

UlrichKu: Ok, should be done with a "broader" version.

But working / testing with this is easier with this merged...

PR #454, opened 2020-05-20
ref #51: Use other minifier to avoid broken css

UlrichKu:
| Q             | A
| ------------- | ---
| Branch        | 6.3.x
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed issues  | chameleon-system/chameleon-system#51
| License       | MIT


src/CoreBundle/private/library/classes/TCMSResourceCollection.class.php
```diff
@@ -297,7 +298,12 @@ protected function CreateCSSResourceCollectionFile($sFileMD5, $aCSS)
 
                             $minify = ServiceLocator::getParameter('chameleon_system_core.resources.enable_external_resource_collection_minify');
                             if ($minify) {
-                                $sSubString = CssMin::minify($sSubString);
+                                // TODO use a service here (see MinifyJsJshrinkService) - if there are any problems with customer css content with new lib?
```
pixeljunkie, line 301: Absolutely. This should be replaceable. A service to call the minification in between should not be a big deal.

src/CoreBundle/private/library/classes/TCMSResourceCollection.class.php
```diff
@@ -297,7 +298,12 @@ protected function CreateCSSResourceCollectionFile($sFileMD5, $aCSS)
...
+
+                                $minifier = new CSS();
+                                $minifier->add($sSubString);
+
+                                $sSubString = $minifier->minify();
```
pixeljunkie, line 306: We should replace the variable name here to our standard: $substring (remove the hungarian notation)

UlrichKu: (That variable is not introduced here. Changing it would change a number of unaffected lines. The whole method is using hungarian notation.)

src/CoreBundle/private/library/classes/TCMSResourceCollection.class.php
UlrichKu: Hm, the Js minification even uses an event listener mechanism: \TCMSResourceCollection::dispatchJSMinifyEvent()

src/CoreBundle/private/library/classes/TCMSResourceCollection.class.php
pixeljunkie: ok

src/CoreBundle/private/library/classes/TCMSResourceCollection.class.php
UlrichKu: Added a service for now.

PR #458, opened 2020-06-05
Fix/590 access array offset

kzorluoglu:
| Q             | A
| ------------- | ---
| Branch        | 6.2.x
| Bug fix?      | yes
| New feature?  | no 
| BC breaks?    | no     
| Deprecations? | no
| Tests pass?   | yes
| Fixed issues  | chameleon-system/chameleon-system#590
| License       | MIT

We have now null coalescing before the array offset access and if/else refactoring.

### DIFF
--- a/src/CoreBundle/private/library/classes/dbobjects/TCMSImageEndpoint.class.php
+++ b/src/CoreBundle/private/library/classes/dbobjects/TCMSImageEndpoint.class.php
@@ -736,9 +736,10 @@ class TCMSImageEndpoint
         /** @var string|null $dataPath */
         $dataPath = $this->aData['path'] ?? '';
 
-        $sImageURL = URL_MEDIA_LIBRARY_THUMBS_PATH.$this->GetThumbPathExtension().'/'.$dataPath;
         if (false === $this->_isThumbnail) {
             $sImageURL = $this->GetImageUrlPathPrefix().$dataPath;
+        }else{
+            $sImageURL = URL_MEDIA_LIBRARY_THUMBS_PATH.$this->GetThumbPathExtension().'/'.$dataPath;
         }
 
         $sImageURL = $this->addRefreshToken($sImageURL);
--- a/src/CoreBundle/private/library/classes/dbobjects/TCMSImageEndpoint.class.php
+++ b/src/CoreBundle/private/library/classes/dbobjects/TCMSImageEndpoint.class.php
@@ -733,7 +733,7 @@ class TCMSImageEndpoint
      */
     public function GetRelativeURL()
     {
-        if (!$this->_isThumbnail) {
+        if (false === $this->_isThumbnail) {
             $sImageURL = $this->GetImageUrlPathPrefix().$this->aData['path'];
         } else {
             $sImageURL = URL_MEDIA_LIBRARY_THUMBS_PATH.$this->GetThumbPathExtension().'/'.$this->aData['path'];
--- a/src/CoreBundle/private/library/classes/dbobjects/TCMSImageEndpoint.class.php
+++ b/src/CoreBundle/private/library/classes/dbobjects/TCMSImageEndpoint.class.php
@@ -736,11 +736,11 @@ class TCMSImageEndpoint
         /** @var string|null $dataPath */
         $dataPath = $this->aData['path'] ?? '';
 
-        if (!$this->_isThumbnail) {
+        $sImageURL = URL_MEDIA_LIBRARY_THUMBS_PATH.$this->GetThumbPathExtension().'/'.$dataPath;
+        if (false === $this->_isThumbnail) {
             $sImageURL = $this->GetImageUrlPathPrefix().$dataPath;
-        } else {
-            $sImageURL = URL_MEDIA_LIBRARY_THUMBS_PATH.$this->GetThumbPathExtension().'/'.$dataPath;
         }
+
         $sImageURL = $this->addRefreshToken($sImageURL);
 
         return $sImageURL;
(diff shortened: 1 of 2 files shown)


src/CoreBundle/private/library/classes/dbobjects/TCMSImageEndpoint.class.php
```diff
@@ -733,11 +733,14 @@ public static function imagecopymerge_alpha($dst_im, $src_im, $dst_x, $dst_y, $s
      */
     public function GetRelativeURL()
     {
-        if (!$this->_isThumbnail) {
-            $sImageURL = $this->GetImageUrlPathPrefix().$this->aData['path'];
-        } else {
-            $sImageURL = URL_MEDIA_LIBRARY_THUMBS_PATH.$this->GetThumbPathExtension().'/'.$this->aData['path'];
+        /** @var string|null $dataPath */
+        $dataPath = $this->aData['path'] ?? '';
```
UlrichKu, line 737: In what case is aData a boolean?

I take it that is some earlier (access) error? It should probably be either catched already there or be explicitly checked here (as an early return?).

src/CoreBundle/private/library/classes/dbobjects/TCMSImageEndpoint.class.php
```diff
@@ -733,11 +733,14 @@ public static function imagecopymerge_alpha($dst_im, $src_im, $dst_x, $dst_y, $s
...
+
+        $sImageURL = URL_MEDIA_LIBRARY_THUMBS_PATH.$this->GetThumbPathExtension().'/'.$dataPath;
+        if (false === $this->_isThumbnail) {
```
UlrichKu, line 740: This changes the original if-else logic.

At least if this if is true then the previsous line was executed superfluously.

src/CoreBundle/private/library/classes/dbobjects/TCMSImageEndpoint.class.php
kzorluoglu: aData is a empty array() in normally request. and empty array is false in PHP 7.4.
This case attached as image in the Issue.

src/CoreBundle/private/library/classes/dbobjects/TCMSImageEndpoint.class.php
kzorluoglu: true, you'r aright. Else is back.

src/CoreBundle/private/library/classes/dbobjects/TCMSImageEndpoint.class.php
kzorluoglu: This changes have been reset, fix now directly in to the Load method.

src/CoreBundle/private/library/classes/dbobjects/TCMSImageEndpoint.class.php
```diff
@@ -733,15 +733,11 @@ public static function imagecopymerge_alpha($dst_im, $src_im, $dst_x, $dst_y, $s
...
     public function GetRelativeURL()
     {
-        /** @var string|null $dataPath */
-        $dataPath = $this->aData['path'] ?? '';
-
-        if (false === $this->_isThumbnail) {
-            $sImageURL = $this->GetImageUrlPathPrefix().$dataPath;
-        }else{
-            $sImageURL = URL_MEDIA_LIBRARY_THUMBS_PATH.$this->GetThumbPathExtension().'/'.$dataPath;
+        if (!$this->_isThumbnail) {
```
UlrichKu: Our current code convention is false ===

src/CoreBundle/private/library/classes/dbobjects/TCMSPortal.class.php
```diff
@@ -192,8 +192,10 @@ public function IsDomainOfThisPortal($sDomainName)
     public function &GetLogo()
     {
         if (!array_key_exists('_portalLogo', $this->_aCache)) {
-            $this->_aCache['_portalLogo'] = new TCMSImage();
-            $this->_aCache['_portalLogo']->Load($this->sqlData['images']);
+            $portalLogo = new TCMSImage();
+            if(false !== $portalLogo->Load($this->sqlData['images'])){
```
UlrichKu: There are some spaces missing (start and end) -> cs fix

src/CoreBundle/private/library/classes/dbobjects/TCMSPortal.class.php
```diff
@@ -192,8 +192,10 @@ public function IsDomainOfThisPortal($sDomainName)
...
+                $this->_aCache['_portalLogo'] = $portalLogo;
```
UlrichKu: The value $this->_aCache['_portalLogo'] must be set in any case (in error case null if that works?) because it is used for the return value.

src/CoreBundle/private/library/classes/dbobjects/TCMSImageEndpoint.class.php
kzorluoglu: it was just rollback. But, its updated.

src/CoreBundle/private/library/classes/dbobjects/TCMSPortal.class.php
kzorluoglu: thanks for this feedback. Re-formatted.

src/CoreBundle/private/library/classes/dbobjects/TCMSPortal.class.php
kzorluoglu: okay, changes resetted.

PR #459, opened 2020-06-09
587 store table obj correctly

UlrichKu:
| Q             | A
| ------------- | ---
| Branch        | 6.3.x
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed issues  | chameleon-system/chameleon-system#587
| License       | MIT

### DIFF
--- a/src/CoreBundle/private/library/classes/TCMSListManager/TCMSListManagerFullGroupTable.class.php
+++ b/src/CoreBundle/private/library/classes/TCMSListManager/TCMSListManagerFullGroupTable.class.php
@@ -49,7 +49,10 @@ class TCMSListManagerFullGroupTable extends TCMSListManager
         $_SESSION['_tmpCurrentTableID'] = $this->oTableConf->sqlData['id']; // needed for the callback functions...
 
         $listCacheKey = $this->GetListCacheKey();
-        $cachedTableObj = $this->getTableFromSessionCache($listCacheKey);
+        $cachedTableObj = null;
+        if (true === $this->isTableCachingEnabled()) {
+            $cachedTableObj = $this->getTableFromSessionCache($listCacheKey);
+        }
 
         $aOrderData = array();
         if (null !== $cachedTableObj) {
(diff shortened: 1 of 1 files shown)


src/CoreBundle/private/library/classes/TCMSListManager/TCMSListManagerFullGroupTable.class.php
```diff
@@ -99,21 +89,48 @@ public function Init(&$oTableConf)
...
-        $inputFilter = $this->getInputFilterUtil();
+        return $this->bListCacheEnabled && CMS_ACTIVE_BACKEND_LIST_CACHE;
+    }
 
-        if (null !== $inputFilter->getFilteredInput('_fnc')) {
-            return false;
+    private function getTableFromSessionCache(string $cacheKey): ?TFullGroupTable
+    {
+        if (false === \array_key_exists('_listObjCache', $_SESSION)) {
+            $_SESSION['_listObjCache'] = [];
```
rnuernberg, line 105: if the key does not exists, then we can early exit, can we not?

src/CoreBundle/private/library/classes/TCMSListManager/TCMSListManagerFullGroupTable.class.php
```diff
@@ -48,39 +48,29 @@ public function Init(&$oTableConf)
...
-            $aOrderData = $oOldTableObj->orderList;
+        if (null !== $cachedTableObj) {
+            $aOrderData = $cachedTableObj->orderList;
         }
 
         // table is not in cache, load it
-        if (is_null($oOldTableObj) || false === $oOldTableObj) {
+        if (null === $cachedTableObj) {
             $this->CreateTableObj();
+            $this->PostCreateTableObjectHook();
```
rnuernberg, line 65: CreateTableObj is a public method - so anyone could be calling it. By removing the PostCreateTableObjectHook you have introduced an unexpected change. Although I agree, that the call to the post create would be better placed here, I don't think we should change it in a patch level version. In any case, it would have to be documented in the upgrade documentation

src/CoreBundle/private/library/classes/TCMSListManager/TCMSListManagerFullGroupTable.class.php
UlrichKu: True. Changed.

(It was a replica of the old code which also had exactly this case.)

src/CoreBundle/private/library/classes/TCMSListManager/TCMSListManagerFullGroupTable.class.php
UlrichKu: ok, changed back

PR #460, opened 2020-06-09
ref #557: Correct loading order of js: cms.js between jquery and the rest

UlrichKu:
| Q             | A
| ------------- | ---
| Branch        | 6.3.x
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed issues  | chameleon-system/chameleon-system#557
| License       | MIT

### DIFF
--- a/src/CoreBundle/EventListener/AddGlobalHeadIncludesListener.php
+++ b/src/CoreBundle/EventListener/AddGlobalHeadIncludesListener.php
@@ -12,6 +12,7 @@
 namespace ChameleonSystem\CoreBundle\EventListener;
 
 use ChameleonSystem\CoreBundle\Event\HtmlIncludeEventInterface;
+use ChameleonSystem\CoreBundle\Service\RequestInfoServiceInterface;
 use TGlobal;
 
 /**
@@ -26,11 +27,14 @@ class AddGlobalHeadIncludesListener
     private $viewRendererSnippetDirectory;
 
     /**
-     * @param \TPkgViewRendererSnippetDirectoryInterface $viewRendererSnippetDirectory
+     * @var RequestInfoServiceInterface
      */
-    public function __construct(\TPkgViewRendererSnippetDirectoryInterface $viewRendererSnippetDirectory)
+    private $requestInfoService;
+
+    public function __construct(\TPkgViewRendererSnippetDirectoryInterface $viewRendererSnippetDirectory, RequestInfoServiceInterface $requestInfoService)
     {
         $this->viewRendererSnippetDirectory = $viewRendererSnippetDirectory;
+        $this->requestInfoService = $requestInfoService;
     }
 
     /**
@@ -38,9 +42,13 @@ class AddGlobalHeadIncludesListener
      */
     public function onGlobalHtmlHeaderInclude(HtmlIncludeEventInterface $event)
     {
-        $event->addData([
-            '<script src="'.TGlobal::GetStaticURLToWebLib('/javascript/cms.js').'" type="text/javascript"></script>',
-        ]);
+        if (true === $this->requestInfoService->isBackendMode()) {
+            $event->addData(
+                [
+                    '<script src="'.TGlobal::GetStaticURLToWebLib( '/javascript/cms.js').'" type="text/javascript"></script>',
+                ]
+            );
+        }
 
         $event->addData($this->viewRendererSnippetDirectory->getResourcesForSnippetPackage(''));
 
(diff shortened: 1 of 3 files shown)


src/CoreBundle/private/rendering/layouts/includes/cms_head_data.inc.php
```diff
@@ -32,10 +34,9 @@
...
 <script src="<?= TGlobal::GetStaticURLToWebLib('/javascript/jquery/jquery-form-4.2.2/jquery.form.min.js'); ?>" type="text/javascript"></script><!--#GLOBALRESOURCECOLLECTION#-->
-<script src="<?= TGlobal::GetStaticURLToWebLib('/javascript/cms.js'); ?>" type="text/javascript"></script><!--#GLOBALRESOURCECOLLECTION#-->
 <?php // Scrollbar JS needs to be loaded before CoreUI?>
 <script src="<?= TGlobal::GetPathTheme(); ?>/coreui/js/perfect-scrollbar.min.js" type="text/javascript"></script><!--#GLOBALRESOURCECOLLECTION#-->
 <script src="<?= TGlobal::GetPathTheme(); ?>/coreui/js/coreui.min.js" type="text/javascript"></script><!--#GLOBALRESOURCECOLLECTION#-->
 <script src="<?= TGlobal::GetPathTheme(); ?>/coreui/js/coreui-utilities.min.js" type="text/javascript"></script><!--#GLOBALRESOURCECOLLECTION#-->
 </head>
-<body class="<?=$cssClasses; ?>" <?=$bodyAttributes; ?>>
\ No newline at end of file
+<body class="<?=$cssClasses; ?>" <?=$bodyAttributes; ?>
```
kzorluoglu, line 42: i think, the body close tag ">" must be still there?

UlrichKu: Indeed. Good catch. (added)

src/CoreBundle/EventListener/AddGlobalHeadIncludesListener.php
```diff
@@ -38,12 +38,16 @@ public function __construct(\TPkgViewRendererSnippetDirectoryInterface $viewRend
      */
     public function onGlobalHtmlHeaderInclude(HtmlIncludeEventInterface $event)
     {
+        $event->addData([
+            '<script src="'.TGlobal::GetStaticURLToWebLib('/javascript/cms.js').'" type="text/javascript"></script>',
+        ]);
```
rnuernberg, line 43: this would also inject the cms.js into the frontend, would it not? I think we should avoid that, unless it is also used in the frontend.

src/CoreBundle/private/rendering/layouts/includes/cms_head_data.inc.php
```diff
@@ -32,10 +34,9 @@
 <script src="<?= TGlobal::GetStaticURLToWebLib('/bootstrap/js/bootstrap.bundle.min.js?v4.1.3'); ?>" type="text/javascript"></script><!--#GLOBALRESOURCECOLLECTION#-->
 <script src="<?= TGlobal::GetStaticURLToWebLib('/components/bootstrap3-typeahead/bootstrap3-typeahead.min.js'); ?>" type="text/javascript"></script><!--#GLOBALRESOURCECOLLECTION#-->
 <script src="<?= TGlobal::GetStaticURLToWebLib('/javascript/jquery/jquery-form-4.2.2/jquery.form.min.js'); ?>" type="text/javascript"></script><!--#GLOBALRESOURCECOLLECTION#-->
-<script src="<?= TGlobal::GetStaticURLToWebLib('/javascript/cms.js'); ?>" type="text/javascript"></script><!--#GLOBALRESOURCECOLLECTION#-->
```
rnuernberg, line 35: looks to me as if we load cms.js via resource collection - but none of the other libraries. Why the special handling of cms.js?

UlrichKu: all of these are similar - just the location of cms.js changed

UlrichKu: cms.js is more or less the "base" library for the backend and could be used/required by others: so it should be among the first (after jquery)

src/CoreBundle/EventListener/AddGlobalHeadIncludesListener.php
UlrichKu: True, checking that now.

UlrichKu:
Will be done by a "private" merge to my feature branch: https://github.com/UlrichKu/chameleon-base/pull/1

(The feature will be reverted here...)

PR #461, opened 2020-06-17
ref #595: add missing target table definition to   field 'parent' in …

rnuernberg:
| Q             | A
| ------------- | ---
| Branch        | 6.2.x
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Fixed issues  | chameleon-system/chameleon-system#595
| License       | MIT

fix definition of the parent field in the cms_document_tree table

### DIFF
--- a/src/CoreBundle/Bridge/Chameleon/Migration/Script/update-1592409240.inc.php
+++ b/src/CoreBundle/Bridge/Chameleon/Migration/Script/update-1592409240.inc.php
@@ -1,0 +1,26 @@
+<h1>Build #1592409240</h1>
+<h2>Date: 2020-06-17</h2>
+<div class="changelog">
+    - ref #595 add missing target table config to document tree parent field
+</div>
+<?php
+$documentTreeParentFieldId = TCMSLogChange::GetTableFieldId(
+    TCMSLogChange::GetTableId('cms_document_tree'),
+    'parent_id'
+);
+$query = "SELECT `fieldtype_config` FROM `cms_field_conf` WHERE `id` = :id";
+$fieldConfig = TCMSLogChange::getDatabaseConnection()->fetchColumn($query, ['id' => $documentTreeParentFieldId]);
+if (false === stripos($fieldConfig, 'connectedTableName')) {
+    $data = TCMSLogChange::createMigrationQueryData('cms_field_conf', 'de')
+        ->setFields(
+            [
+                'fieldtype_config' => 'connectedTableName=cms_document_tree',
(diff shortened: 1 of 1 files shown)


src/CoreBundle/Bridge/Chameleon/Migration/Script/update-1592409240.inc.php
```diff
@@ -0,0 +1,26 @@
...
+    TCMSLogChange::GetTableId('cms_document_tree'),
+    'parent_id'
+);
+$query = "SELECT `fieldtype_config` FROM `cms_field_conf` WHERE `id` = :id";
+$fieldConfig = TCMSLogChange::getDatabaseConnection()->fetchColumn($query, ['id' => $documentTreeParentFieldId]);
+if (false === stripos($fieldConfig, 'connectedTableName')) {
+    $data = TCMSLogChange::createMigrationQueryData('cms_field_conf', 'de')
+        ->setFields(
+            [
+                'fieldtype_config' => 'connectedTableName=cms_document_tree',
```
UlrichKu, line 17: Could there be another fieldtype_config already?

src/CoreBundle/Bridge/Chameleon/Migration/Script/update-1592409240.inc.php
```diff
@@ -0,0 +1,26 @@
+<h1>Build #1592409240</h1>
```
UlrichKu, line 1: (issue description) Why is it only sometimes crashing?

src/CoreBundle/Bridge/Chameleon/Migration/Script/update-1592409240.inc.php
rnuernberg: you mean one that configures everything but the connected table (because I check that in line 13). I think that is very very unlikely seeing as the field is broken as long as the connected table is not configured. So I would argue, that if someone added a configuration, then they would have had to also set the connected table name in order for the field to work.

src/CoreBundle/Bridge/Chameleon/Migration/Script/update-1592409240.inc.php
rnuernberg: to be honest, I have no idea why it crashes only sometimes. But is that relevant? The missconfiguration of the field leads to a broken method in the TAdb that attempts to load a 'TdbParent' class - which does not exists.

PR #464, opened 2020-07-03
ref #597: make resource collection target path and url configurable

rnuernberg:
| Q             | A
| ------------- | ---
| Branch        | 6.2.x
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed issues  | chameleon-system/chameleon-system#597
| License       | MIT

Makes the path for resource collection files configurable


src/CoreBundle/Resources/config/services.xml
```diff
@@ -5,13 +5,18 @@
 
     <parameters>
         <parameter key="vendor_dir">%kernel.root_dir%/../vendors</parameter>
+        <parameter key="chameleon_system_core.const.path_outbox" type="constant">PATH_OUTBOX</parameter>
```
UlrichKu, line 8: Could we use a bundle config here?

rnuernberg: not for the constants - since we may or may not want to use them.

The other two parameters could be moved to a bundle config - so i'll do that

rnuernberg: configuration has been added

src/CoreBundle/DependencyInjection/ChameleonSystemCoreExtension.php
```diff
@@ -57,6 +57,10 @@ public function load(array $config, ContainerBuilder $container)
         $this->configureSession($container);
 
         $this->addResources($container);
+
+        $resourceCollectionDefinition = $container->getDefinition('chameleon_system_core.resource_collector');
```
UlrichKu, line 61: Could be done with a method (after configureSession - consistent with those methods there).

rnuernberg: ok - moved

PR #465, opened 2020-07-03
Feature/598 add locking to resource collection

rnuernberg:
| Q             | A
| ------------- | ---
| Branch        | 6.2.x
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed issues  | chameleon-system/chameleon-system#598
| License       | MIT

Notice that this is based on https://github.com/chameleon-system/chameleon-base/pull/464 since we need the path to the resource directory.

### DIFF
--- a/src/CoreBundle/ResourceCollection/ResourceCollectionInformationService.php
+++ b/src/CoreBundle/ResourceCollection/ResourceCollectionInformationService.php
@@ -1,0 +1,26 @@
+<?php
+
+namespace ChameleonSystem\CoreBundle\ResourceCollection;
+
+use ChameleonSystem\CoreBundle\Interfaces\ResourceCollectorInterface;
+
+class ResourceCollectionInformationService implements ResourceCollectionInformationServiceInterface
+{
+    /**
+     * @var ResourceCollectorInterface
+     */
+    private $resourceCollector;
+
+    public function __construct(ResourceCollectorInterface $resourceCollector)
+    {
+        $this->resourceCollector = $resourceCollector;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function isActive(): bool
+    {
+        return $this->resourceCollector->IsAllowed();
+    }
+}
--- a/src/CoreBundle/ResourceCollection/ResourceCollectionInformationServiceInterface.php
+++ b/src/CoreBundle/ResourceCollection/ResourceCollectionInformationServiceInterface.php
@@ -1,0 +1,12 @@
+<?php
+
+namespace ChameleonSystem\CoreBundle\ResourceCollection;
+
+interface ResourceCollectionInformationServiceInterface
+{
+    /**
+     * Returns true if the resource collection is enabled and not locked or otherwise
+     * blocked for the current request.
+     */
+    public function isActive(): bool;
+}
--- a/src/CoreBundle/Resources/config/services.xml
+++ b/src/CoreBundle/Resources/config/services.xml
@@ -807,6 +807,12 @@
             <argument type="service" id="request_stack"/>
         </service>
 
+        <service class="ChameleonSystem\CoreBundle\ResourceCollection\ResourceCollectionInformationService"
+                 id="chameleon_system_core.resource_collector.info"
+                 public="true">
+            <argument id="chameleon_system_core.resource_collector" type="service"/>
+        </service>
+
         <service id="chameleon_system_core.resource_collector" class="TCMSResourceCollection" shared="false">
             <argument type="service" id="chameleon_system_core.filemanager"/>
             <argument type="service" id="chameleon_system_core.portal_domain_service"/>
--- a/src/CoreBundle/Resources/config/services.xml
+++ b/src/CoreBundle/Resources/config/services.xml
@@ -5,6 +5,8 @@
 
     <parameters>
         <parameter key="vendor_dir">%kernel.root_dir%/../vendors</parameter>
+        <parameter key="chameleon_system_core.const.path_outbox" type="constant">PATH_OUTBOX</parameter>
+        <parameter key="chameleon_system_core.const.url_outbox" type="constant">URL_OUTBOX</parameter>
         <parameter key="chameleon_system_core.pdo.mysql_attr_init_command" type="constant">\PDO::MYSQL_ATTR_INIT_COMMAND</parameter>
         <parameter key="chameleon_system_core.pdo.mysql_attr_compression_name" type="constant">\PDO::MYSQL_ATTR_COMPRESS</parameter>
         <parameter key="chameleon_system_core.pdo.enable_mysql_compression">false</parameter>
@@ -12,6 +14,9 @@
         <parameter key="chameleon_system_core.controller_invoke_listener.class">ChameleonSystem\CoreBundle\Controller\ChameleonControllerInvokeListener</parameter>
         <parameter key="chameleon_system_core.tpl_page_controller.class">TPkgCmsRouteControllerCmsTplPage</parameter>
         <parameter key="chameleon_system_core.render_media_properties_field">true</parameter>
+
+        <parameter key="chameleon_system_core.resource_collector.asset_base_url">%chameleon_system_core.const.url_outbox%/static</parameter>
+        <parameter key="chameleon_system_core.resource_collector.asset_base_path">%chameleon_system_core.const.path_outbox%/static</parameter>
     </parameters>
 
     <services>
@@ -806,6 +811,8 @@
             <argument type="service" id="chameleon_system_core.filemanager"/>
             <argument type="service" id="chameleon_system_core.portal_domain_service"/>
             <argument type="service" id="event_dispatcher"/>
+            <argument>%chameleon_system_core.resource_collector.asset_base_url%</argument>
+            <argument>%chameleon_system_core.resource_collector.asset_base_path%</argument>
         </service>
 
         <service id="chameleon_system_core.command.update.count" class="ChameleonSystem\CoreBundle\Command\CountUpdateCommand">
--- a/src/CoreBundle/private/library/classes/TCMSResourceCollection.class.php
+++ b/src/CoreBundle/private/library/classes/TCMSResourceCollection.class.php
@@ -189,8 +189,8 @@ class TCMSResourceCollection implements ResourceCollectorInterface
         );
         $aJSGlobal = $this->StaticContentCollector('jsglobal');
 
-        $hasCollectedCss = \count($aCSS) > 0 && \count($aGlobalCSS);
-        $hasCollectedJs = \count($aJS) > 0 && \count($aJSGlobal);
+        $hasCollectedCss = \count($aCSS) > 0 && \count($aGlobalCSS) > 0;
+        $hasCollectedJs = \count($aJS) > 0 && \count($aJSGlobal) > 0;
         if (false === $hasCollectedCss && false === $hasCollectedJs) {
             return $sPageContent;
         }
--- a/src/CoreBundle/private/library/classes/TCMSResourceCollection.class.php
+++ b/src/CoreBundle/private/library/classes/TCMSResourceCollection.class.php
@@ -45,15 +45,27 @@ class TCMSResourceCollection implements ResourceCollectorInterface
      * @var EventDispatcherInterface
      */
     private $eventDispatcher;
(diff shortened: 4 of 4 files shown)


src/CoreBundle/Resources/config/services.xml
```diff
@@ -5,13 +5,18 @@
 
     <parameters>
         <parameter key="vendor_dir">%kernel.root_dir%/../vendors</parameter>
+        <parameter key="chameleon_system_core.const.path_outbox" type="constant">PATH_OUTBOX</parameter>
```
UlrichKu, line 8: Could we use bundle configuration here?

src/CoreBundle/private/library/classes/TCMSResourceCollection.class.php
```diff
@@ -69,98 +81,161 @@ public function IsEnabled()
...
+            // do not enable in backend mode
+            return false;
+        }
+        $bIsTemplateEngine = 'true' === $bIsModuleChooser && TCMSUser::CMSUserDefined();
+        if (true === $bIsTemplateEngine) {
+            return false;
         }
 
-        return $bAllowResourceCollection;
+        if (true === $this->resourceCollectionProcessRunning()) {
```
UlrichKu, line 99: -> "isRunning()" ?

src/CoreBundle/private/library/classes/TCMSResourceCollection.class.php
```diff
@@ -734,4 +809,18 @@ protected function isLocalResource($sResource)
 
         return $bIsLocal;
     }
+
+    private function resourceCollectionProcessRunning(): bool
+    {
+        return file_exists($this->assetPath.'/lock.tmp');
```
UlrichKu, line 815: Maybe (additionally) use something quicker than this? (memcache/redis/db)?

And/or check for the existence of a result (ressource collection of another process) before overwriting?

src/CoreBundle/private/library/classes/TCMSResourceCollection.class.php
```diff
@@ -69,98 +81,161 @@ public function IsEnabled()
...
+        $resourceFilesExist = $resourceFilesExist && \file_exists($sFileCSSGlobalMD5);
+        $resourceFilesExist = $resourceFilesExist && \file_exists($sFileJSGlobalMD5);
 
-                if (2 == count($aPageParts)) {
-                    $sPageContent = $sReworkContent.'</head>'.$aPageParts[1];
-                } else {
-                    $sPageContent = $sReworkContent;
-                }
-                $sPageContent = $this->AddResourceCollectionToHead($sPageContent, $sFileCSSMD5, $sFileJSMD5, $sFileCSSGlobalMD5, $sFileJSGlobalMD5);
+        if (false === $resourceFilesExist) {
```
UlrichKu, line 207: It needs to check the lock here before setting it. (?)

src/CoreBundle/ResourceCollection/ResourceCollectionInformationService.php
```diff
@@ -0,0 +1,26 @@
...
+
+    public function __construct(ResourceCollectorInterface $resourceCollector)
+    {
+        $this->resourceCollector = $resourceCollector;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function isActive(): bool
```
UlrichKu, line 22: -> sounds like "resource collection is activated?"

-> isRunning()?

src/CoreBundle/ResourceCollection/ResourceCollectionInformationService.php
```diff
@@ -0,0 +1,26 @@
+<?php
+
+namespace ChameleonSystem\CoreBundle\ResourceCollection;
+
+use ChameleonSystem\CoreBundle\Interfaces\ResourceCollectorInterface;
+
+class ResourceCollectionInformationService implements ResourceCollectionInformationServiceInterface
```
UlrichKu, line 7: This is (currently) specific to "locking", so -> ResourceCollectionLockingService (or similar)?

src/CoreBundle/ResourceCollection/ResourceCollectionInformationServiceInterface.php
```diff
@@ -0,0 +1,12 @@
...
+
+namespace ChameleonSystem\CoreBundle\ResourceCollection;
+
+interface ResourceCollectionInformationServiceInterface
+{
+    /**
+     * Returns true if the resource collection is enabled and not locked or otherwise
+     * blocked for the current request.
+     */
+    public function isActive(): bool;
```
UlrichKu, line 11: Is this used somewhere already?

src/CoreBundle/private/library/classes/TCMSResourceCollection.class.php
```diff
@@ -69,98 +81,161 @@ public function IsEnabled()
      */
     public function IsAllowed()
```
UlrichKu, line 82: Is this (already) checked somewhere else?

What does happen when there are two concurrent requests with the second request?

src/CoreBundle/private/library/classes/TCMSResourceCollection.class.php
rnuernberg: i need this to be on a per code basis, and i need this to work even if caching is not enabled. So a file_exists looks like a reasonable solution.

As for checking if the files exists - we do that in src/CoreBundle/private/library/classes/TCMSResourceCollection.class.php lines 202-205

src/CoreBundle/private/library/classes/TCMSResourceCollection.class.php
rnuernberg: not sure what you mean. We check if all files exist - if some don't, then files will be written and we need a lock. If all files exists, no files will be written so we need no lock

src/CoreBundle/ResourceCollection/ResourceCollectionInformationService.php
rnuernberg: It is not only related to locking - it indicates if the resource collection is enabled and is collecting resources.

src/CoreBundle/ResourceCollection/ResourceCollectionInformationServiceInterface.php
rnuernberg: Yes - in templates that load resources into the header or the footer depending on the resource collection being active.

src/CoreBundle/private/library/classes/TCMSResourceCollection.class.php
rnuernberg: IsAllowed is called by the core - to decided if to trigger resource collection. It had no concept of concurrent calls so it would return true for every request wanting to generate resources. That is the change I made - that it now only allows resource collection if no other process on that node is currently writing resources

src/CoreBundle/ResourceCollection/ResourceCollectionInformationService.php
rnuernberg: no - in fact, wenn resource collection is running, this method will return false. See documentation in `ResourceCollectionInformationServiceInterface::isActive` when the method will return true.

src/CoreBundle/private/library/classes/TCMSResourceCollection.class.php
rnuernberg: isRunning is not quite correct - it is more a "is currently writing" info. So renamed methods to

hasResourceCollectionWriteProcessRunning

and 

setsResourceCollectionWriteProcessRunning

src/CoreBundle/private/library/classes/TCMSResourceCollection.class.php
rnuernberg: lock is now checked as part of the lock acquire method

src/CoreBundle/ResourceCollection/ResourceCollectionInformationServiceInterface.php
UlrichKu: But wouldn't I need to see these calls to isActive() in this PR. Because they have to be new (?)

src/CoreBundle/private/library/classes/TCMSResourceCollection.class.php
```diff
@@ -734,4 +811,18 @@ protected function isLocalResource($sResource)
 
         return $bIsLocal;
     }
+
+    private function hasResourceCollectionWriteProcessRunning(): bool
+    {
+        return file_exists($this->assetPath.'/lock.tmp');
+    }
+
+    private function setsResourceCollectionWriteProcessRunning(bool $state): bool
```
UlrichKu, line 820: -> setRes...

src/CoreBundle/private/library/classes/TCMSResourceCollection.class.php
```diff
@@ -734,4 +811,18 @@ protected function isLocalResource($sResource)
...
+    {
+        if (true === $state) {
+            return false === $this->hasResourceCollectionWriteProcessRunning() && touch($this->assetPath.'/lock.tmp');
```
UlrichKu, line 823: It's a bit surprising to have a set method do these things and also the hierarchy levels of the code are mixed ("check for file" is another helper method and the "create" isn't).

So for simplification / clarification: Why not have three methods:
isResourceCollectionProcessActive, startResourceCollectionProcess and finishResourceCollectionProcess?

src/CoreBundle/private/library/classes/TCMSResourceCollection.class.php
UlrichKu: The three methods below would also make this here more clear:

if "is running" return early.

(otherwise) start

rnuernberg: I don't understand what you mean

src/CoreBundle/ResourceCollection/ResourceCollectionInformationServiceInterface.php
rnuernberg: no - because I am using them in a project - not in the core.

src/CoreBundle/private/library/classes/TCMSResourceCollection.class.php
rnuernberg: ok

src/CoreBundle/private/library/classes/TCMSResourceCollection.class.php
rnuernberg: you are right - we should be checking for the existence of the file and not call `hasResourceCollectionWriteProcessRunning`. This way the method has only one level of abstraction and it becomes much clearer what is going on.

I changed the method along those lines.

src/CoreBundle/private/library/classes/TCMSResourceCollection.class.php
rnuernberg: ok - I understand now... The logic hier is: attempt to acquire lock - if that fails, do not run the resource collection. 

Acquiring the lock will fail if someone already has the lock - but also when we fail to write the lock file. So if we moved to the three methods you suggested, then the logic here would become even more complex (because we would still need to check if we succeeded to acquire the lock).

src/CoreBundle/private/library/classes/TCMSResourceCollection.class.php
```diff
@@ -69,98 +81,163 @@ public function IsEnabled()
...
-                    $sMinifyStatus = 'true';
-                } else {
-                    $sMinifyStatus = 'false';
-                }
-                $sFileJSMD5 = $filesPrefix.'.js.'.md5(implode(';', $aJS).$sMinifyStatus).'.js';
-                if (false === $this->CreateJSResourceCollectionFile($sFileJSMD5, $aJS)) {
-                    $sFileJSMD5 = ''; // reset file because we don`t have includes
-                }
+        $hasCollectedCss = \count($aCSS) > 0 && \count($aGlobalCSS);
+        $hasCollectedJs = \count($aJS) > 0 && \count($aJSGlobal);
```
UlrichKu, line 193: Are there > 0 missing at the end?

rnuernberg: yes - fixed it

PR #473, opened 2020-07-27
574 check module groups

UlrichKu:
| Q             | A
| ------------- | ---
| Branch        | master (7.2 or 8.0)
| Bug fix?      | yes
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed issues  | chameleon-system/chameleon-system#574
| License       | MIT


src/CoreBundle/private/core/TModelBase.class.php
```diff
@@ -194,6 +194,14 @@ public function Init()
     {
     }
 
+    /**
+     * @return bool - false if access is forbidden
+     */
+    public function checkAccessRightsOnTable(): bool
```
rnuernberg, line 200: I don't think the method makes sense on a module basis. For one, Permissions may or may not have anything to do with tables. But also, this is a class that is the base of frontend and backend modules - and as far as I can see, the check would not make much sense in the frontend.

For the backend we know the tables, and should look to the module definition, not the module, for checking permissions.

src/CoreBundle/private/library/classes/TCMSPageDefinitionFile.class.php
```diff
@@ -236,6 +241,15 @@ protected function LoadPageDefVars($sPagedef)
             if (isset($staticModuleList)) {
                 $this->staticModuleList = $staticModuleList;
             }
+            if (isset($allowedRights) && \strlen($allowedRights) > 0) {
+                $rightAbbreviations = \explode(',', $allowedRights);
+                foreach ($rightAbbreviations as $rightAbbreviation) {
```
rnuernberg, line 246: what is an abbreviation in this context?

src/CoreBundle/Controller/ChameleonController.php
```diff
@@ -289,22 +297,28 @@ protected function GeneratePage($pagedef)
...
     {
-        $pagedefData = $this->dataAccessCmsMasterPagedef->get($pagedef);
-        if (null === $pagedefData) {
+        // TODO restrict to backend? (is already the case with GetActiveUser()?)
+
+        $activeUser = \TCMSUser::GetActiveUser();
+        if (null === $activeUser) {
+            // For a number of edge cases: cronjobs, 404, ...
+
+            return true;
```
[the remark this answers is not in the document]
UlrichKu: Yes.

It's a rather low-level location (every request gets here). And cronjobs have no user but should still get access to some locations.

src/CoreBundle/Bridge/Chameleon/Migration/Script/update-1595841100.inc.php
```diff
@@ -0,0 +1,18 @@
...
+</div>
+<?php
+
+$data = TCMSLogChange::createMigrationQueryData('cms_field_conf', 'en')
+  ->setFields([
+      'modifier' => 'hidden',
+  ])
+  ->setWhereEquals([
+      'name' => 'cms_right_mlt',
+      'cms_tbl_conf_id' => TCMSLogChange::GetTableId('cms_menu_custom_item'),
```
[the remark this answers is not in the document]
UlrichKu: ok

src/CoreBundle/Bridge/Chameleon/Module/Sidebar/CustomMenuItemProvider.php
```diff
@@ -11,8 +11,37 @@
 
 namespace ChameleonSystem\CoreBundle\Bridge\Chameleon\Module\Sidebar;
 
+use ChameleonSystem\CoreBundle\DataAccess\DataAccessCmsMasterPagedefInterface;
+use ChameleonSystem\CoreBundle\Security\PageAccessCheckInterface;
```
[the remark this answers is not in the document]
UlrichKu: ok

the concrete class (was old and) had the correct name

src/CoreBundle/Controller/ChameleonController.php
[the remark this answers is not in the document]
UlrichKu: Might be debated however.

But the later method checkPageAccess() needs a user so this here is an early return and "false" does not really make sense there.

src/CoreBundle/Controller/ChameleonController.php
```diff
@@ -289,22 +297,28 @@ protected function GeneratePage($pagedef)
...
-     * @param string $pagedef
-     *
-     * @return array|bool
-     */
-    private function getPagedefData($pagedef)
+    private function checkAccess(CmsMasterPagdef $pagedef): bool
     {
-        $pagedefData = $this->dataAccessCmsMasterPagedef->get($pagedef);
-        if (null === $pagedefData) {
+        // TODO restrict to backend? (is already the case with GetActiveUser()?)
```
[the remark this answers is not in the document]
UlrichKu: streamline?

I moved it now to the ChameleonBackendController.

But am not really happy with that either as it now jugles with some protected members.

src/CoreBundle/private/library/classes/TCMSPageDefinitionFile.class.php
UlrichKu: renamed to "rightName"

src/CoreBundle/Resources/BackendPageDefs/navigationTreePlain.pagedef.php
```diff
@@ -18,4 +18,6 @@
     ],
 ];
 
+$allowedRights = 'navigation_edit';
```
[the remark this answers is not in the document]
UlrichKu: Change to array.

It would be an "OR list" so required rights would be misleading (suggesting AND).

src/CoreBundle/Controller/ChameleonController.php
```diff
@@ -289,22 +297,28 @@ protected function GeneratePage($pagedef)
...
+
+        $activeUser = \TCMSUser::GetActiveUser();
+        if (null === $activeUser) {
+            // For a number of edge cases: cronjobs, 404, ...
+
+            return true;
+        }
+
+        foreach ($this->moduleLoader->modules as $module) {
+            if (false === $module->checkAccessRightsOnTable()) {
```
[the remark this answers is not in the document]
UlrichKu: This an old legacy case (which I factored to this method): for most backend pages which are tables (table editors or table managers) the access right is defined as acess rights defined for the respective table.

And as these classes are "TModelBase" (= modules) I chose that name.

src/CoreBundle/DataModel/CmsMasterPagdef.php
```diff
@@ -17,11 +17,23 @@ class CmsMasterPagdef
...
+     */
+    private $allowedRights;
+
+    /**
+     * @param string         $id
+     * @param array          $moduleList
+     * @param string         $layoutFile
+     * @param \TdbCmsRight[] $allowedRights
+     */
+    public function __construct(string $id, array $moduleList, string $layoutFile, array $allowedRights)
```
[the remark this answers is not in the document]
UlrichKu: Actually the "name" is used in other locations.

But as this is a simple data model (and we already have the Tdb) it makes no sense to foresee external usages here.

src/CoreBundle/private/core/TModelBase.class.php
UlrichKu: (mostly this is old code refactored)

These here are the "modules" provided by TModuleLoader and they are annotated as TModelBase. And those in turn were loaded from the simple string array CmsMasterPagdef::getModuleList().

The two usecases (MTTableManager, MTTableEditor) would be TCMSModelBase but that would need to be checked by instanceof.

src/CoreBundle/Controller/ChameleonController.php
```diff
@@ -238,22 +240,27 @@ protected function sendDefaultHeaders()
...
 
         foreach ($this->moduleLoader->modules as $sSpotName => $module) {
             $this->moduleLoader->modules[$sSpotName]->InjectVirtualModuleSpots($this->moduleLoader);
         }
         reset($this->moduleLoader->modules);
 
         $this->InitializeModules();
+
+        if (false === $this->checkAccess($pagedefData)) {
+            return new Response('You have no sufficient rights to view this page.', Response::HTTP_FORBIDDEN);
```
rnuernberg, line 258: This is backend logic only. So IMHO this also needs to be moved into the BackendController

src/CoreBundle/DataModel/CmsMasterPagdef.php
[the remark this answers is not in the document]
rnuernberg: I am not looking at other usages - I think a method should only take the parameters it needs. If it only needs the IDs of the permissions, then that is all it should geht. 

In your case you are using the name field - so I would think that should be enough. But I think the point is not important enough.

rnuernberg (changes requested):

src/CoreBundle/private/core/TModelBase.class.php
UlrichKu: Moved.

However ChameleonBackendController::checkAccess() has now a warning: Potentially polymorphic call
(wrong type of \TModuleLoader::$modules)

src/CoreBundle/Controller/ChameleonController.php
UlrichKu: It depends however on the 10 lines before it - and ChameleonBackendController has nothing of this yet.

src/CoreBundle/private/core/TModelBase.class.php
UlrichKu: Moved it back: The modules in question here are "frontend modules".

UlrichKu: (ie StaticViewModule or MediaManagerBackendModule)

src/CoreBundle/Resources/BackendPageDefs/navigationTreePlain.pagedef.php
```diff
@@ -18,4 +18,6 @@
     ],
 ];
 
+$allowedRights = ['navigation_edit'];
```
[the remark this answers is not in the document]
UlrichKu: Renamed.

src/CoreBundle/private/core/TModelBase.class.php
UlrichKu: Moved it to its own interface: CheckTableAccessRightsInterface

However this is only an interface for two (legacy) cases: MTTableEditor, MTTableManager

For all else the access check is done via the pagedef (and only in ChameleonBackendController).

src/CoreBundle/DataAccess/DataAccessCmsMasterPagedefDatabase.php
```diff
@@ -25,11 +25,12 @@ public function __construct(DataAccessCmsMasterPagedefInterface $fallbackLoader,
         $this->inputFilterUtil = $inputFilterUtil;
     }
 
-    public function get(string $id): ?CmsMasterPagdef
+    public function get(string $id, ?string $type = null): ?CmsMasterPagdef
```
rnuernberg, line 28: we are changing the interface here - why? And if we are, can we do away with the default?

Also-  it looks to me as if we are passing the value deep into the chain before finally using it.... In addition, we are only using it in `DataAccessCmsMasterPagedefFile` - that would make me very hesitant to move the parameter into the interface.

If the file requires a type (but no other pagedef needs this), then I would suggest to simply add the type as part of the pagedef ID.

TYPE:ID for example. That way we can keep the interface as it is - and have `DataAccessCmsMasterPagedefFile` extract the type from the ID-String passed.

src/CoreBundle/DataModel/CmsMasterPagdef.php
rnuernberg, line 31: I would suggest changing the name of `allowedRights` into `cmsRightAllowList` everywhere

rnuernberg: oh - and this is the data model for alle pagedefs, right? Including the pagedefs for the frontend... If that is the case, then again we should avoid introducing a parameter that will be used only by a very small fraction of pages.

I would suggest extending the class to create a backend specific version.

`class BackedPageDefDataModel extends CmsMasterPagdef`

The extension can then add the attributes required for backend page handling.

src/CoreBundle/Security/BackendPageAccessCheckInterface.php
```diff
@@ -0,0 +1,24 @@
...
+
+interface BackendPageAccessCheckInterface
+{
+    /**
+     * Checks for page access of the current user/client to the current page.
+     * In case of failure this redirects to the login page.
+     *
+     * @return void
+     */
+    public function assertAccess();
```
rnuernberg, line 15: I would remove assertAccess. I think things like redirection belong into the controller, not into this service.

src/CoreBundle/Security/BackendPageAccessCheckInterface.php
```diff
@@ -0,0 +1,24 @@
...
+
+    /**
+     * @param \TdbCmsUser     $user
+     * @param CmsMasterPagdef $pagedef
+     *
+     * @return bool - false if page access is forbidden
+     */
+    public function checkPageAccess(\TdbCmsUser $user, CmsMasterPagdef $pagedef): bool;
```
rnuernberg, line 23: I would rename the method to `permitAccess`

src/CoreBundle/private/modules/MTTableEditor/MTTableEditor.class.php
```diff
@@ -217,6 +202,28 @@ public function Init()
         $this->AddURLHistory();
     }
 
+    /**
+     * {@inheritDoc}
+     */
+    public function checkAccessRightsOnTable(): bool
```
rnuernberg, line 208: I would avoid implementation details in the method name. For the call, it does not matter HOW or on WHAT the module checks access permission.

rnuernberg (changes requested):

src/CoreBundle/Controller/ChameleonBackendController.php
```diff
@@ -77,4 +79,32 @@ public function setHomePagedef(string $homePagedef): void
     {
         $this->homePagedef = $homePagedef;
     }
+
+    /**
+     * {@inheritDoc}
+     */
+    protected function checkAccess(CmsMasterPagdef $pagedef): bool
```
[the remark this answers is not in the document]
UlrichKu: "permit" would be more appropriate if this method directly throws an exception or similar 

isAllowed(page) ?

But I thinks checkAccess(page) is quite on point with what is done.

src/CoreBundle/Controller/ChameleonBackendController.php
```diff
@@ -77,4 +79,32 @@ public function setHomePagedef(string $homePagedef): void
...
+                continue;
+            }
+            if (false === $module->checkAccessRightsOnTable()) {
+                return false;
+            }
+        }
+
+        if (false === $this->pageAccessCheck->checkPageAccess($activeUser, $pagedef)) {
+            return false;
+        }
```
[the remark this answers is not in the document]
UlrichKu: It looks from a complexity point similar - there aren't normally many modules for any page.

But reordered it.

src/CoreBundle/Controller/ChameleonController.php
```diff
@@ -289,22 +296,9 @@ protected function GeneratePage($pagedef)
...
-        $pagedefData = $this->dataAccessCmsMasterPagedef->get($pagedef);
-        if (null === $pagedefData) {
-            return false;
-        }
-
-        return             [
-            'moduleList' => $pagedefData->getModuleList(),
-            'sLayoutFile' => $pagedefData->getLayoutFile(),
-        ];
+        return true;
```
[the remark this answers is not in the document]
UlrichKu: done

src/CoreBundle/Controller/ChameleonFrontendController.php
```diff
@@ -40,13 +42,15 @@ public function __construct(
...
+        BackendPageAccessCheckInterface $pageAccessCheck,
+        RequestInfoServiceInterface $requestInfoService,
         DataAccessCmsMasterPagedefInterface $dataAccessCmsMasterPagedef,
         TModuleLoader $moduleLoader,
         $viewPathManager,
         ContainerInterface $container,
         TPkgViewRendererConfigToLessMapper $configToLessMapper
     ) {
-        parent::__construct($requestStack, $eventDispatcher, $portalDomainService, $dataAccessCmsMasterPagedef, $moduleLoader, $viewPathManager);
+        parent::__construct($requestStack, $eventDispatcher, $portalDomainService, $pageAccessCheck, $requestInfoService, $dataAccessCmsMasterPagedef, $moduleLoader, $viewPathManager);
```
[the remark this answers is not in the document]
UlrichKu: ok cleaned up

there was already an old setter injection for this check interface as well

src/CoreBundle/Controller/ChameleonBackendController.php
```diff
@@ -77,4 +79,32 @@ public function setHomePagedef(string $homePagedef): void
...
+    /**
+     * {@inheritDoc}
+     */
+    protected function checkAccess(CmsMasterPagdef $pagedef): bool
+    {
+        $activeUser = \TCMSUser::GetActiveUser();
+        if (null === $activeUser) {
+            // For a number of edge cases: cronjobs, 404, ...
+
+            return true;
```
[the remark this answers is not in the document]
UlrichKu: It is at least true for cronjob execution.

I added 0 === \count($pagedef->getAllowedRights())

src/CoreBundle/Security/BackendPageAccessCheckInterface.php
UlrichKu: (it was present and called like this before in ChameleonBackendController::getResponse - only missing in the interface)

src/CoreBundle/private/modules/MTTableEditor/MTTableEditor.class.php
UlrichKu: That is then sub-issue to the new interface name.

What do you think of that? CheckTableAccessRightsInterface

src/CoreBundle/DataAccess/DataAccessCmsMasterPagedefDatabase.php
UlrichKu: Finally it ends up in DataAccessCmsMasterPagedefFile::getPageDefinitionFilePath() where the old implementation was (only) reading it from the request.

At least in those request only cases there is probably no sensible default.

As this is an interface change it would be a BC break to require an additional parameter I guess.

UlrichKu: "Hiding" the type in the id sounds hackish/wrong?

src/CoreBundle/DataModel/CmsMasterPagdef.php
UlrichKu: From an information encapsulation point this probably makes sense.

However as interfaces use this as return type (ie DataAccessCmsMasterPagedefInterface::get()) we would need at least at some points an instanceof (or a hard cast) where the else case would be a runtime error.

src/CoreBundle/private/modules/MTTableEditor/MTTableEditor.class.php
rnuernberg: why table? why use AccessCheckInterface::checkAccess() (or assertAccess). 
That way we can just use one interface (AccessCheckInterface) instead of two (BackendPageAccessCheckInterface and CheckTableAccessRightsInterface).

rnuernberg (changes requested):

src/CoreBundle/Controller/ChameleonBackendController.php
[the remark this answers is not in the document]
UlrichKu: looks ok to me in "Files changed".

The line is before the module loop.

src/CoreBundle/Controller/ChameleonController.php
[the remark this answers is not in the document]
UlrichKu: Looks ok for me on "Files changed".

src/CoreBundle/Security/BackendPageAccessCheckInterface.php
UlrichKu: It is a reason to do it (only) for consistency reasons.

At least the typing in ChameleonBackendController::getResponse() was wrong before.

More was not changed here.

src/CoreBundle/private/modules/MTTableEditor/MTTableEditor.class.php
UlrichKu: Hm, CheckTableAccessRightsInterface is still a solution for an old legacy case - and requires an "instanceof".

Whereas BackendPageAccessCheckInterface::checkPageAccess ist the more modern approach of rights on a page..

src/CoreBundle/DataModel/CmsMasterPagdef.php
UlrichKu: Add a special "file version" of the data model - and check for it in two locations.

src/CoreBundle/DataAccess/DataAccessCmsMasterPagedefDatabase.php
UlrichKu: (changed pull request target to master)

rnuernberg:
There are open comments and a conflict.

UlrichKu:
Conflicts fixed.

All open comments seem to wait for your input?

pixeljunkie:
Needs conflict fixes first.

bestform:
Closing for further review on gitlab

PR #474, opened 2020-07-28
ref #606: prevent endless recursion

rnuernberg:
When GetReadOnly calls GetHTML we get into trouble whenever someone makes the field read only (or some extension such as  \TCMSFieldBlob defaults to read only).

| Q             | A
| ------------- | ---
| Branch        | 7.0.x
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed issues  | chameleon-system/chameleon-system#606
| License       | MIT

Prevent endless recursion when a field that extends the text field is put into read only mode (such as TCMSFieldBlob)

### DIFF
--- a/src/CoreBundle/private/library/classes/TCMSFields/TCMSFieldText.class.php
+++ b/src/CoreBundle/private/library/classes/TCMSFields/TCMSFieldText.class.php
@@ -26,16 +26,21 @@ class TCMSFieldText extends TCMSField
     {
         parent::GetHTML();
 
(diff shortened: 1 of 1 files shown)


UlrichKu:
This pull request should be pointing to 6.3.x (the current feature minimum version) or even 6.2 as it is a crash bug.

rnuernberg:
replaced by https://github.com/chameleon-system/chameleon-base/pull/482

PR #476, opened 2020-08-02
ref #610: add restriction to extended lookup

rnuernberg:
| Q             | A
| ------------- | ---
| Branch        | 7.0.x
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed issues  | chameleon-system/chameleon-system#610
| License       | MIT

The extended lookup no longer supported restrictions based on the current record

### DIFF
--- a/src/CoreBundle/private/library/classes/TCMSFields/TCMSFieldExtendedLookup.class.php
+++ b/src/CoreBundle/private/library/classes/TCMSFields/TCMSFieldExtendedLookup.class.php
@@ -162,15 +162,15 @@ class TCMSFieldExtendedLookup extends TCMSFieldLookup
         $equalPosition = \mb_strpos($restrictionExpression, '=');
         $restrictionField = \trim(\mb_substr($restrictionExpression, 0, $equalPosition));
         $restrictionValue = \trim(\mb_substr($restrictionExpression, $equalPosition + 1));
-        $restrictionValue = \mb_substr($restrictionValue, 2, -2); // ignore [{}]
-        if (\property_exists($this, 'oTableRow')
-            && \is_object($this->oTableRow)
-            && \property_exists($this->oTableRow, 'sqlData')
-            && \is_array($this->oTableRow->sqlData)) {
-            if (isset($this->oTableRow->sqlData[$restrictionValue])) {
-                $restrictionValue = $this->oTableRow->sqlData[$restrictionValue];
+
+        if (\mb_strpos($restrictionValue, '[{') === 0 && '}]' === \mb_substr($restrictionValue, -2)) {
+            $restrictionValueReplaced = $this->getFieldValueFromFieldToken($restrictionValue);
+            if (null === $restrictionValueReplaced) {
+                $restrictionValueReplaced = \sprintf('field token %s not found in record.', $restrictionValue);
             }
+            $restrictionValue = $restrictionValueReplaced;
         }
+
         if ('' === $restrictionField || '' === $restrictionValue) {
             return [];
         }
@@ -179,6 +179,27 @@ class TCMSFieldExtendedLookup extends TCMSFieldLookup
             'sRestrictionField' => $restrictionField,
             'sRestriction' => $restrictionValue,
         ];
+    }
+
+    /**
+     * Returns field value in current record if token matches a known field. Otherwise, null will be returned.
+     * @param string $fieldToken - token in the form [{fieldName}]
+     * @return string|null
+     */
+    private function getFieldValueFromFieldToken(string $fieldToken): ?string
+    {
+        if (false === \property_exists($this, 'oTableRow')) {
+            return null;
+        }
+        if (false === \is_object($this->oTableRow)) {
+            return null;
+        }
+        if (false === \is_array($this->oTableRow->sqlData)) {
+            return null;
+        }
+        $fieldName = \trim(\mb_substr($fieldToken, 2, -2));
+
+        return $this->oTableRow->sqlData[$fieldName] ?? null;
     }
 
     /**
--- a/src/CoreBundle/private/library/classes/TCMSFields/TCMSFieldExtendedLookup.class.php
+++ b/src/CoreBundle/private/library/classes/TCMSFields/TCMSFieldExtendedLookup.class.php
@@ -160,10 +160,10 @@ class TCMSFieldExtendedLookup extends TCMSFieldLookup
             return [];
         }
         $equalPosition = \mb_strpos($restrictionExpression, '=');
-        $restrictionField = trim(\mb_substr($restrictionExpression, 0, $equalPosition));
-        $restrictionValue = trim(\mb_substr($restrictionExpression, $equalPosition+1));
-        $restrictionValue = mb_substr($restrictionValue, 2, -2); // ignore [{}]
-        if (\property_exists($this, 'oTableRow') && \is_object($this->oTableRow) && \property_exists($this->oTableRow, 'sqlData') && is_array($this->oTableRow->sqlData)) {
+        $restrictionField = \trim(\mb_substr($restrictionExpression, 0, $equalPosition));
+        $restrictionValue = \trim(\mb_substr($restrictionExpression, $equalPosition+1));
+        $restrictionValue = \mb_substr($restrictionValue, 2, -2); // ignore [{}]
+        if (\property_exists($this, 'oTableRow') && \is_object($this->oTableRow) && \property_exists($this->oTableRow, 'sqlData') && \is_array($this->oTableRow->sqlData)) {
             if (isset($this->oTableRow->sqlData[$restrictionValue])) {
                 $restrictionValue = $this->oTableRow->sqlData[$restrictionValue];
             }
--- a/src/CoreBundle/private/library/classes/TCMSFields/TCMSFieldExtendedLookup.class.php
+++ b/src/CoreBundle/private/library/classes/TCMSFields/TCMSFieldExtendedLookup.class.php
@@ -161,9 +161,9 @@ class TCMSFieldExtendedLookup extends TCMSFieldLookup
         }
         $equalPosition = \mb_strpos($restrictionExpression, '=');
         $restrictionField = \trim(\mb_substr($restrictionExpression, 0, $equalPosition));
-        $restrictionValue = \trim(\mb_substr($restrictionExpression, $equalPosition + 1));
+        $restrictionValue = \mb_substr($restrictionExpression, $equalPosition + 1);
 
-        if (\mb_strpos($restrictionValue, '[{') === 0 && '}]' === \mb_substr($restrictionValue, -2)) {
+        if (true === $this->isFieldToken($restrictionValue)) {
             $restrictionValueReplaced = $this->getFieldValueFromFieldToken($restrictionValue);
             if (null === $restrictionValueReplaced) {
                 $restrictionValueReplaced = \sprintf('field token %s not found in record.', $restrictionValue);
@@ -181,6 +181,12 @@ class TCMSFieldExtendedLookup extends TCMSFieldLookup
         ];
     }
 
+    private function isFieldToken(string $string): bool
+    {
+        $string = \trim($string);
+        return \mb_strpos($string, '[{') === 0 && '}]' === \mb_substr($string, -2);
+    }
+
     /**
      * Returns field value in current record if token matches a known field. Otherwise, null will be returned.
      * @param string $fieldToken - token in the form [{fieldName}]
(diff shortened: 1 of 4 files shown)


src/CoreBundle/private/library/classes/TCMSFields/TCMSFieldExtendedLookup.class.php
```diff
@@ -152,6 +150,30 @@ protected function _GetOpenWindowJS(&$oPopupTableConf)
...
+        if (null === $restrictionExpression) {
+            return [];
+        }
+        if (false === \strpos($restrictionExpression, '=')) {
+            return [];
+        }
+        $equalPosition = \mb_strpos($restrictionExpression, '=');
+        $restrictionField = trim(\mb_substr($restrictionExpression, 0, $equalPosition));
+        $restrictionValue = trim(\mb_substr($restrictionExpression, $equalPosition+1));
+        $restrictionValue = mb_substr($restrictionValue, 2, -2); // ignore [{}]
```
UlrichKu, line 165: Is this the line where something was fixed (I can't really easily tell when comparing old to new code)?

This cuts four characters regardless of their content? Maybe check first?

src/CoreBundle/private/library/classes/TCMSFields/TCMSFieldExtendedLookup.class.php
```diff
@@ -152,6 +150,30 @@ protected function _GetOpenWindowJS(&$oPopupTableConf)
...
+        if (\property_exists($this, 'oTableRow') && \is_object($this->oTableRow) && \property_exists($this->oTableRow, 'sqlData') && is_array($this->oTableRow->sqlData)) {
+            if (isset($this->oTableRow->sqlData[$restrictionValue])) {
+                $restrictionValue = $this->oTableRow->sqlData[$restrictionValue];
+            }
+        }
+        return [
+            'sRestrictionField' => $restrictionField,
+            'sRestriction' => $restrictionValue,
```
UlrichKu, line 173: (old problem) $restrictionValue (or $restrictionField) could still be empty here.

src/CoreBundle/private/library/classes/TCMSFields/TCMSFieldExtendedLookup.class.php
```diff
@@ -152,6 +150,30 @@ protected function _GetOpenWindowJS(&$oPopupTableConf)
...
+    {
+        $restrictionExpression = $this->oDefinition->GetFieldtypeConfigKey('restriction');
+        if (null === $restrictionExpression) {
+            return [];
+        }
+        if (false === \strpos($restrictionExpression, '=')) {
+            return [];
+        }
+        $equalPosition = \mb_strpos($restrictionExpression, '=');
+        $restrictionField = trim(\mb_substr($restrictionExpression, 0, $equalPosition));
```
UlrichKu, line 163: There are 3 \ missing (or 7 too much).

src/CoreBundle/private/library/classes/TCMSFields/TCMSFieldExtendedLookup.class.php
```diff
@@ -137,11 +137,9 @@ protected function GetExtendedListButtons()
     protected function _GetOpenWindowJS(&$oPopupTableConf)
     {
         $aParams = array('pagedef' => 'extendedLookupList', 'id' => $oPopupTableConf->id, 'fieldName' => $this->name, 'sourceTblConfId' => $this->oDefinition->fieldCmsTblConfId);
-        $sRestriction = $this->oDefinition->GetFieldtypeConfigKey('restriction');
-        $aRestrictionParts = explode('=', $sRestriction);
-        if (2 == count($aRestrictionParts) && property_exists($this, 'oTableRow') && is_object($this->oTableRow) && property_exists($this->oTableRow, 'sqlData') && is_array($this->oTableRow->sqlData) && array_key_exists($aRestrictionParts[0], $this->oTableRow->sqlData)) {
-            $aParams['sRestriction'] = $aRestrictionParts[1];
-            $aParams['sRestrictionField'] = $aRestrictionParts[0];
+        $restriction = $this->getTargetListRestriction();
```
UlrichKu, line 140: This is an old problem (for years)?

src/CoreBundle/private/library/classes/TCMSFields/TCMSFieldExtendedLookup.class.php
rnuernberg: There is no single line fix. The problem was, that the old code would take the restriction:

`foo=[{bar}]`

at face value. That is, it would use `foo` as the restriction field and `[{bar}]` as the restriction value. Instead, the content of [{bar}] is to be used - and to be replaced by a field value IF bar is a field name.

It may make sense to change the definition in suche a way, that `[{}]` will always assume that we are dealing with a field while any other content will be taken as a literal value.

so that `foo=[{bar}]` would mean
restrictionField = foo
restrictionValue = content of field bar

and `foo=bar` would mean
restrictionField = foo
restrictionValue = bar

But if we decided to go that way we also need to check the lookup field. That may exceed the scope of a patch release...

I would feel better though if the logic used in the lookup class was identical to the logic used here... so maybe we do need to revisit this code section

src/CoreBundle/private/library/classes/TCMSFields/TCMSFieldExtendedLookup.class.php
rnuernberg: yes. 8 years at least

src/CoreBundle/private/library/classes/TCMSFields/TCMSFieldExtendedLookup.class.php
rnuernberg: 3 slashes hinzugefügt

src/CoreBundle/private/library/classes/TCMSFields/TCMSFieldExtendedLookup.class.php
rnuernberg: ok - check added

src/CoreBundle/private/library/classes/TCMSFields/TCMSFieldExtendedLookup.class.php
rnuernberg: refactored to allow static string restrictions as well as field references - and updated field documentation to match what the field now does.

src/CoreBundle/private/library/classes/TCMSFields/TCMSFieldExtendedLookup.class.php
```diff
@@ -152,6 +150,58 @@ protected function _GetOpenWindowJS(&$oPopupTableConf)
...
+            return [];
+        }
+        $equalPosition = \mb_strpos($restrictionExpression, '=');
+        $restrictionField = \trim(\mb_substr($restrictionExpression, 0, $equalPosition));
+        $restrictionValue = \trim(\mb_substr($restrictionExpression, $equalPosition + 1));
+
+        if (\mb_strpos($restrictionValue, '[{') === 0 && '}]' === \mb_substr($restrictionValue, -2)) {
+            $restrictionValueReplaced = $this->getFieldValueFromFieldToken($restrictionValue);
+            if (null === $restrictionValueReplaced) {
+                $restrictionValueReplaced = \sprintf('field token %s not found in record.', $restrictionValue);
```
UlrichKu, line 169: This is then simply shown in the URL?

src/CoreBundle/private/library/classes/TCMSFields/TCMSFieldExtendedLookup.class.php
```diff
@@ -152,6 +150,58 @@ protected function _GetOpenWindowJS(&$oPopupTableConf)
...
+            return [];
+        }
+        if (false === \strpos($restrictionExpression, '=')) {
+            return [];
+        }
+        $equalPosition = \mb_strpos($restrictionExpression, '=');
+        $restrictionField = \trim(\mb_substr($restrictionExpression, 0, $equalPosition));
+        $restrictionValue = \trim(\mb_substr($restrictionExpression, $equalPosition + 1));
+
+        if (\mb_strpos($restrictionValue, '[{') === 0 && '}]' === \mb_substr($restrictionValue, -2)) {
```
UlrichKu, line 166: Is space (trim()) a potential problem here?

(And possible simplification: if (true === $this->isTokenString()) and/or if ($this->startsWith() && $this->endsWith()).)

src/CoreBundle/private/library/classes/TCMSFields/TCMSFieldExtendedLookup.class.php
```diff
@@ -152,6 +150,58 @@ protected function _GetOpenWindowJS(&$oPopupTableConf)
...
+        }
+        if (false === \is_object($this->oTableRow)) {
+            return null;
+        }
+        if (false === \is_array($this->oTableRow->sqlData)) {
+            return null;
+        }
+        $fieldName = \trim(\mb_substr($fieldToken, 2, -2));
+
+        return $this->oTableRow->sqlData[$fieldName] ?? null;
```
UlrichKu, line 202: Wouldn't these two lines be sufficient here (esp. ?? )?

src/CoreBundle/private/library/classes/TCMSFields/TCMSFieldExtendedLookup.class.php
rnuernberg: > Is space (trim()) a potential problem here?

changed the code to keep the trim only for the `[{}]` variation (in which case the trim is relevant).

> 
> (And possible simplification: if (true === $this->isTokenString()) and/or if ($this->startsWith() && $this->endsWith()).)

replached with `isFieldToken` and moved trim into `isFieldToken` and `getFieldValueFromFieldToken`

src/CoreBundle/private/library/classes/TCMSFields/TCMSFieldExtendedLookup.class.php
rnuernberg: You are right. I thought that the null coalescing operator would not work on nested properties and that I would need the nullsafe operator. Turns out I would need that only if there was a method call in the chain.

src/CoreBundle/private/library/classes/TCMSFields/TCMSFieldExtendedLookup.class.php
rnuernberg: Changed to now show a toaster error message as well

rnuernberg:
@UlrichKu I'll change the target branch to 6.3 once all discussions have been resolved.

src/CoreBundle/private/library/classes/TCMSFields/TCMSFieldExtendedLookup.class.php
```diff
@@ -152,6 +152,68 @@ protected function _GetOpenWindowJS(&$oPopupTableConf)
...
+            return [];
+        }
+        $equalPosition = \mb_strpos($restrictionExpression, '=');
+        $restrictionField = \trim(\mb_substr($restrictionExpression, 0, $equalPosition));
+        $restrictionValue = \mb_substr($restrictionExpression, $equalPosition + 1);
+
+        if (true === $this->isFieldToken($restrictionValue)) {
+            $restrictionValueReplaced = $this->getFieldValueFromFieldToken($restrictionValue);
+            if (null === $restrictionValueReplaced) {
+                $error = $this->getTranslationService()->trans(
```
UlrichKu, line 171: AddBackendToasterMessage expects an id (and then does the translation).

src/CoreBundle/private/library/classes/TCMSFields/TCMSFieldExtendedLookup.class.php
```diff
@@ -152,6 +152,68 @@ protected function _GetOpenWindowJS(&$oPopupTableConf)
...
+        if (false === \strpos($restrictionExpression, '=')) {
+            return [];
+        }
+        $equalPosition = \mb_strpos($restrictionExpression, '=');
+        $restrictionField = \trim(\mb_substr($restrictionExpression, 0, $equalPosition));
+        $restrictionValue = \mb_substr($restrictionExpression, $equalPosition + 1);
+
+        if (true === $this->isFieldToken($restrictionValue)) {
+            $restrictionValueReplaced = $this->getFieldValueFromFieldToken($restrictionValue);
+            if (null === $restrictionValueReplaced) {
```
UlrichKu, line 170: Can such a value (in the target table) be correctly empty or null and thus falsely trigger this error?

rnuernberg: we return null only if the field is not set or is null. Since we currently do not support null value fields, I think the code can safely stay as it is.

src/CoreBundle/private/library/classes/TCMSFields/TCMSFieldExtendedLookup.class.php
rnuernberg: ah indeed. changed

rnuernberg:
replaced by https://github.com/chameleon-system/chameleon-base/pull/483

PR #479, opened 2020-08-04
ref #500: Make corrections to the "parent" menu

UlrichKu:
| Q             | A
| ------------- | ---
| Branch        | 6.3.x
| Bug fix?      | yes
| New feature?  | no - missing feature
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed issues  | chameleon-system/chameleon-system#500
| License       | MIT

### DIFF
--- a/src/CoreBundle/private/modules/CMSTemplateEngine/CMSTemplateEngine.class.php
+++ b/src/CoreBundle/private/modules/CMSTemplateEngine/CMSTemplateEngine.class.php
@@ -475,12 +475,26 @@ class CMSTemplateEngine extends TCMSModelBase
      */
     protected function filterMainNavigation(): void
     {
-        $this->data['oMenuItems'] = $this->oTableManager->oTableEditor->GetMenuItems();
-        $this->data['oMenuItems']->RemoveItem('sItemKey', 'save');
-        $this->data['oMenuItems']->RemoveItem('sItemKey', 'copy');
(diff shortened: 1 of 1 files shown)


src/CoreBundle/private/modules/CMSTemplateEngine/CMSTemplateEngine.class.php
```diff
@@ -475,12 +475,26 @@ protected function _GetModuleInstanceList()
...
+
+        $menuItems->RemoveItem('sItemKey', 'save');
+        $menuItems->RemoveItem('sItemKey', 'copy');
+        $menuItems->RemoveItem('sItemKey', 'new');
+        $menuItems->RemoveItem('sItemKey', 'delete');
+        $menuItems->RemoveItem('sItemKey', 'copy_translation');
+
+        $tableEditorButton = $menuItems->FindItemWithProperty('sItemKey', 'edittableconf');
+        if (false !== $tableEditorButton) {
+            /**
```
pixeljunkie, line 491: The type docblock needs to move to line 488 where the variable is introduced

UlrichKu: Only here the type is safe, two lines earlier it could also be a boolean.

(And it is only needed/meaningful because FindItemWithProperty() is annotated with "object".)

PR #480, opened 2020-08-04
579 provide more memory

UlrichKu:
| Q             | A
| ------------- | ---
| Branch        | 6.3.x
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed issues  | chameleon-system/chameleon-system#579
| License       | MIT

### DIFF
--- a/src/CoreBundle/private/modules/CMSTableExport/CMSTableExport.class.php
+++ b/src/CoreBundle/private/modules/CMSTableExport/CMSTableExport.class.php
@@ -153,19 +153,23 @@ class CMSTableExport extends TCMSModelBase
     {
         $exportMemoryUnits = ServiceLocator::getParameter('chameleon_system.core.export_memory');
 
-        $inBytes = $this->unitToInt($exportMemoryUnits);
+        $configuredBytes = $this->unitToInt($exportMemoryUnits);
+        $currentBytes = $this->unitToInt(ini_get('memory_limit'));
 
-        if ($inBytes > 256*1024*1024) {
+        if ($configuredBytes > $currentBytes) {
             ini_set('memory_limit', $exportMemoryUnits);
         }
-        // else filter out bogus values
     }
 
     /**
      * Converts a number with byte unit (B / K / M / G) into an integer of bytes.
      */
-    private function unitToInt($units): int
+    private function unitToInt(string $units): int
     {
+        if ('' === $units || false === strpos('BKMG', substr($units, -1))) {
+            return 0;
+        }
+
         return (int)preg_replace_callback('/(\-?\d+)(.?)/', function ($matches) {
             return $matches[1] * (1024 ** strpos('BKMG', $matches[2]));
         }, strtoupper($units));
--- a/src/CoreBundle/private/modules/CMSTableExport/CMSTableExport.class.php
+++ b/src/CoreBundle/private/modules/CMSTableExport/CMSTableExport.class.php
@@ -108,7 +108,7 @@ class CMSTableExport extends TCMSModelBase
             $listClass = null;
         }
 
-        $this->oTableList = &$this->oTableConf->GetListObject($listClass);
+        $this->oTableList = &$this->oTableConf->GetListObject($listClass)->tableObj;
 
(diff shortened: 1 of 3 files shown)


src/CoreBundle/private/modules/CMSTableExport/CMSTableExport.class.php
```diff
@@ -99,13 +99,16 @@ public function &Execute()
      */
     public function GenerateExport()
     {
+        // TODO do this for more than this export?
+        $this->provideEnoughMemory();
```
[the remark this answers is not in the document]
UlrichKu: ok

src/CoreBundle/private/modules/CMSTableExport/CMSTableExport.class.php
```diff
@@ -145,6 +149,28 @@ public function GenerateExport()
...
+    private function provideExportMemory(): void
+    {
+        $exportMemoryUnits = ServiceLocator::getParameter('chameleon_system.core.export_memory');
+
+        $inBytes = $this->unitToInt($exportMemoryUnits);
+
+        if ($inBytes > 256*1024*1024) {
+            ini_set('memory_limit', $exportMemoryUnits);
+        }
+        // else filter out bogus values
```
rnuernberg, line 161: what is this?

src/CoreBundle/private/modules/CMSTableExport/CMSTableExport.class.php
```diff
@@ -145,6 +149,28 @@ public function GenerateExport()
         $this->getDownload($sFileType);
     }
 
+    private function provideExportMemory(): void
+    {
+        $exportMemoryUnits = ServiceLocator::getParameter('chameleon_system.core.export_memory');
+
+        $inBytes = $this->unitToInt($exportMemoryUnits);
+
+        if ($inBytes > 256*1024*1024) {
```
rnuernberg, line 158: what is this for? I assume you want to set it only if it exceeds the current memory limit?

src/CoreBundle/private/modules/CMSTableExport/CMSTableExport.class.php
UlrichKu: description of the if

src/CoreBundle/private/modules/CMSTableExport/CMSTableExport.class.php
UlrichKu: not exactly; simply filter out "bad" values ie "4M"

UlrichKu: But I could of course simply trust the user (admin)?

(And get rid of the nice unitToInt method...)

rnuernberg: I think that makes sense. If you do want to validate the input, then the compile Configuration class would be the place to do so.

src/CoreBundle/private/modules/CMSTableExport/CMSTableExport.class.php
rnuernberg: Then we should refactor this like to make it clear. This way it looks like you planed on an else but did not come around to implement it yet:

```php

$exportMemoryUnits = ServiceLocator::getParameter('chameleon_system.core.export_memory');

$inBytes = $this->unitToInt($exportMemoryUnits);

if ($inBytes <= 256*1024*1024) {
    // ignore small values
    return ;
}

ini_set('memory_limit', $exportMemoryUnits);
```

I do not think that makes any sense though. What would make sense here is to check if the value provided is grater than the current memory settings - since it makes no sense to reduce the memory for the export.

rnuernberg: and "bogus" is anything below 256m... I don't think we can assume that nobody would ever attempt to use less memory...

UlrichKu: change to check configured against current value

and added correctness check for values

PR #481, opened 2020-08-04
ref #593: Properly sort and render the checkboxes of a multiselect

UlrichKu:
| Q             | A
| ------------- | ---
| Branch        | 6.3.x
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed issues  | chameleon-system/chameleon-system#593
| License       | MIT

### DIFF
--- a/src/CoreBundle/private/library/classes/TCMSFields/TCMSFieldLookupMultiselectCheckboxes.class.php
+++ b/src/CoreBundle/private/library/classes/TCMSFields/TCMSFieldLookupMultiselectCheckboxes.class.php
@@ -75,8 +75,6 @@ class TCMSFieldLookupMultiselectCheckboxes extends TCMSFieldLookupMultiselect
             $connected = $mltRecord['connected'];
             $displayValue = $mltRecord['display_value'];
 
-            // TODO "display_value" can contain html (which is then rendered as text) - for example with gcf_CMSUserWithImage/callbackCmsUserWithImage for cms_user
-
             $editable = $mltRecord['editable'];
             if ($currentGroup !== $activeGroup) {
                 $activeGroup = $currentGroup;
@@ -164,7 +162,7 @@ class TCMSFieldLookupMultiselectCheckboxes extends TCMSFieldLookupMultiselect
                 'id' => $mltRecord->id,
                 'group' => '' === $listGroupFieldColumn ? '' : $mltRecord->sqlData[$listGroupFieldColumn],
                 'connected' => $mltRecord->isConnected($this->sTableName, $this->oTableRow->sqlData['id'], $mltTableName),
-                'display_value' => $mltRecord->GetDisplayValue(),
+                'display_value' => $mltRecord->GetName(),
                 'editable' => true,
             );
         }
--- a/src/CoreBundle/private/library/classes/TCMSFields/TCMSFieldLookupMultiselectCheckboxes.class.php
+++ b/src/CoreBundle/private/library/classes/TCMSFields/TCMSFieldLookupMultiselectCheckboxes.class.php
@@ -74,6 +74,9 @@ class TCMSFieldLookupMultiselectCheckboxes extends TCMSFieldLookupMultiselect
             $currentGroup = $mltRecord['group'];
             $connected = $mltRecord['connected'];
             $displayValue = $mltRecord['display_value'];
+
+            // TODO "display_value" can contain html (which is then rendered as text) - for example with gcf_CMSUserWithImage/callbackCmsUserWithImage for cms_user
+
             $editable = $mltRecord['editable'];
             if ($currentGroup !== $activeGroup) {
                 $activeGroup = $currentGroup;
@@ -233,17 +236,10 @@ class TCMSFieldLookupMultiselectCheckboxes extends TCMSFieldLookupMultiselect
             $oTableList->sRestriction = null; // do not include the restriction - it is part of the parent table, not the mlt!
 
             $sFilterQuery = $oTableList->FilterQuery().$this->GetMLTRecordRestrictions();
-//            $sFilterQueryOrderInfo = $oTableList->GetSortInfoAsString();
-//            if (!empty($sFilterQueryOrderInfo)) {
-//                $sFilterQuery .= ' ORDER BY '.$sFilterQueryOrderInfo;
-//            }
 
-            // TODO check for tabname and cms_field_conf ?? restrict special case to cms_field_conf?
-            // TODO also how does this relate to $bShowCustomsort above?
-
-            if (false !== \strpos($sFilterQuery, 'AS tabname') && false !== \strpos($sFilterQuery, 'cms_field_conf')) {
-                // TODO multilanguage sort! Also for tabname!
-                $sFilterQuery .= ' ORDER BY tabname ASC, `cms_field_conf`.`translation` ASC ';
+            $sFilterQueryOrderInfo = $oTableList->GetSortInfoAsString();
+            if (!empty($sFilterQueryOrderInfo)) {
+                $sFilterQuery .= ' ORDER BY '.$sFilterQueryOrderInfo;
             }
         } else {
             $sFilterQuery = parent::GetMLTFilterQuery();
--- a/src/CoreBundle/private/library/classes/TCMSFields/TCMSFieldLookupMultiselectCheckboxes.class.php
+++ b/src/CoreBundle/private/library/classes/TCMSFields/TCMSFieldLookupMultiselectCheckboxes.class.php
@@ -67,6 +67,8 @@ class TCMSFieldLookupMultiselectCheckboxes extends TCMSFieldLookupMultiselect
         $activeGroup = '';
         $hasEditPermissionForForeignTable = $this->isRecordChangingAllowed($foreignTableName);
 
+        $mltRecords = $this->sortRecordsAlphabetically($mltRecords);
+
         $inputFilter = $this->getInputFilterUtil();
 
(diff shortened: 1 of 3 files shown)


src/CoreBundle/private/library/classes/TCMSFields/TCMSFieldLookupMultiselectCheckboxes.class.php
```diff
@@ -232,9 +233,17 @@ protected function GetMLTFilterQuery()
             $oTableList->sRestriction = null; // do not include the restriction - it is part of the parent table, not the mlt!
 
             $sFilterQuery = $oTableList->FilterQuery().$this->GetMLTRecordRestrictions();
-            $sFilterQueryOrderInfo = $oTableList->GetSortInfoAsString();
-            if (!empty($sFilterQueryOrderInfo)) {
-                $sFilterQuery .= ' ORDER BY '.$sFilterQueryOrderInfo;
+//            $sFilterQueryOrderInfo = $oTableList->GetSortInfoAsString();
```
pixeljunkie, line 236: We can't remove this. This is needed to allow custom sort orders.

src/CoreBundle/private/library/classes/TCMSFields/TCMSFieldLookupMultiselectCheckboxes.class.php
```diff
@@ -232,9 +233,17 @@ protected function GetMLTFilterQuery()
...
+//            if (!empty($sFilterQueryOrderInfo)) {
+//                $sFilterQuery .= ' ORDER BY '.$sFilterQueryOrderInfo;
+//            }
+
+            // TODO check for tabname and cms_field_conf ?? restrict special case to cms_field_conf?
+            // TODO also how does this relate to $bShowCustomsort above?
+
+            if (false !== \strpos($sFilterQuery, 'AS tabname') && false !== \strpos($sFilterQuery, 'cms_field_conf')) {
+                // TODO multilanguage sort! Also for tabname!
+                $sFilterQuery .= ' ORDER BY tabname ASC, `cms_field_conf`.`translation` ASC ';
```
pixeljunkie, line 246: I think $bShowCustomsort needs to force a sort by a position field, because in this mode the user changes the positions of a record by a sort order field. This can't work with additional sorting on top.

src/CoreBundle/private/library/classes/TCMSFields/TCMSFieldLookupMultiselectCheckboxes.class.php
UlrichKu: Selection_236

This sorting (the one of cms_tbl_conf: "position ASC") is the main reason why it currently looks broken.
It defies the check in the rendering code: if ($currentGroup !== $activeGroup)
As there is no group (tab) ordering in the data then.

src/CoreBundle/private/library/classes/TCMSFields/TCMSFieldLookupMultiselectCheckboxes.class.php
UlrichKu: (also see above)

Isn't "custom sort order" a topic for the actual data of a table?
But in this case ("show fields from table") we are only talking about fields and not the table data?

UlrichKu: I am not sure however as for the general name of the methods (FetchMLTRecords() -> GetMLTFilterQuery()) if this location has to serve two very different purposes and thus cannot work for both.

pixeljunkie:
We checkt the code together and found out, that it looks like the change needs to move to: TCMSFieldLookupMultiselectCheckboxesSelectFieldsFromTable

src/CoreBundle/private/library/classes/TCMSFields/TCMSFieldLookupMultiselectCheckboxesSelectFieldsFromTable.class.php
```diff
@@ -14,6 +14,27 @@
...
+        // NOTE these conditions are _very_ specific to the current configuration of cms_field_conf
+
+        if (false !== \strpos($filterQuery, 'AS tabname') && false !== \strpos($filterQuery, 'cms_field_conf')) {
+            $orderClauseCheck = 'ORDER BY position ASC';
+            $orderPosition = \strpos($filterQuery, $orderClauseCheck);
+
+            // TODO "tabname" is the configuration "group field" or "group field column name" of the table - however that must be present in the custom list query.
+
+            if (false !== $orderPosition) {
+                // NOTE translations are handled later (EntityList?)
```
pixeljunkie, line 30: Yes, the list is rendered with Tdb objects, so translations will be done there. But the query needs to handle the translations, too. Because the order by would not be alphabetical, if it is not  the base language. 

Something like:
ORDER BY ... `cms_field_conf`.'.$titleFieldWithTranslationSuffix.' ASC'

src/CoreBundle/private/library/classes/TCMSFields/TCMSFieldLookupMultiselectCheckboxesSelectFieldsFromTable.class.php
```diff
@@ -14,6 +14,27 @@
...
+    {
+        $filterQuery = parent::GetMLTFilterQuery();
+
+        // NOTE these conditions are _very_ specific to the current configuration of cms_field_conf
+
+        if (false !== \strpos($filterQuery, 'AS tabname') && false !== \strpos($filterQuery, 'cms_field_conf')) {
+            $orderClauseCheck = 'ORDER BY position ASC';
+            $orderPosition = \strpos($filterQuery, $orderClauseCheck);
+
+            // TODO "tabname" is the configuration "group field" or "group field column name" of the table - however that must be present in the custom list query.
```
pixeljunkie, line 27: Looks like this is always the case. Because it is THE most core like table... there should not be changes made here by users i think. As long as we check for the existence of tabname this is not breaking things here.

src/CoreBundle/private/library/classes/TCMSFields/TCMSFieldLookupMultiselectCheckboxes.class.php
```diff
@@ -73,6 +74,9 @@ public function GetHTML()
             $currentGroup = $mltRecord['group'];
             $connected = $mltRecord['connected'];
             $displayValue = $mltRecord['display_value'];
+
+            // TODO "display_value" can contain html (which is then rendered as text) - for example with gcf_CMSUserWithImage/callbackCmsUserWithImage for cms_user
```
pixeljunkie, line 78: Therefore i would suggest using GetName here and not the display value, because the display value is meant for normal lists. 
But... what about Trees, that output a <ul> construct with the breadcrumb? Maybe we can`t encode the output here as HTML. On the other hand.... we did not change things here. So i would let it as before with encoding. 
Or move the output to a method so we can extend it if necessary for fields with this problem (never seen one yet).

src/CoreBundle/private/library/classes/TCMSFields/TCMSFieldLookupMultiselectCheckboxesSelectFieldsFromTable.class.php
UlrichKu: Works in demo shop (with two languages).

src/CoreBundle/private/library/classes/TCMSFields/TCMSFieldLookupMultiselectCheckboxesSelectFieldsFromTable.class.php
UlrichKu: I meant this more like there is a possibility to do this for other cases dynamically (correct).

I removed the comment.

src/CoreBundle/private/library/classes/TCMSFields/TCMSFieldLookupMultiselectCheckboxes.class.php
UlrichKu: Changed it to GetName().

Did you see trees that output a checkbox matrix (as child elements)?

src/CoreBundle/private/library/classes/TCMSFields/TCMSFieldLookupMultiselectCheckboxesSelectFieldsFromTable.class.php
pixeljunkie: Strange. How? Maybe there is some magic fieldname replacement before running the query for all translated fields?

src/CoreBundle/private/library/classes/TCMSFields/TCMSFieldLookupMultiselectCheckboxes.class.php
pixeljunkie: No. Never. The only table with special HTML output in my mind is cms_user, but we don`t need an icon here. The name is enough. So GetName should be better in all cases.

src/CoreBundle/private/library/classes/TCMSFields/TCMSFieldLookupMultiselectCheckboxesSelectFieldsFromTable.class.php
UlrichKu: Removed it now again.

And sort the data in code (only for checkboxes).

PR #490, opened 2020-08-14
ref #620: NavigationTreeSingleSelectWysiwyg: remove useless services

ckleis:
| Q             | A
| ------------- | ---
| Branch        | features / 6.3.x for bug fixes
| Bug fix?      | yes
| New feature?  | no 
| BC breaks?    | no
| Deprecations? | no
| Fixed issues  | chameleon-system/chameleon-system#620
| License       | MIT

After a Symfony and PHP update, it is no longer tolerated that a child class has different service parameters than its parent. This leads to an error now.
I have removed the unnecessary services in NavigationTreeSingleSelectWysiwyg

### DIFF
--- a/src/CoreBundle/Resources/config/services.xml
+++ b/src/CoreBundle/Resources/config/services.xml
@@ -950,8 +950,6 @@
 
         <service id="chameleon_system_core.module.navigation_tree_single_select_wysiwyg" class="ChameleonSystem\CoreBundle\Bridge\Chameleon\Module\NavigationTreeSingleSelectWysiwyg\NavigationTreeSingleSelectWysiwyg" shared="false">
             <tag name="chameleon_system.module"/>
-            <argument id="database_connection" type="service"/>
-            <argument id="event_dispatcher" type="service"/>
             <argument id="chameleon_system_core.util.input_filter" type="service"/>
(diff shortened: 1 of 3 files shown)


src/CoreBundle/Resources/config/services.xml
```diff
@@ -950,8 +950,6 @@
 
         <service id="chameleon_system_core.module.navigation_tree_single_select_wysiwyg" class="ChameleonSystem\CoreBundle\Bridge\Chameleon\Module\NavigationTreeSingleSelectWysiwyg\NavigationTreeSingleSelectWysiwyg" shared="false">
```
UlrichKu, line 951: You can further simplify (and correct) this by giving a "parent":

parent="chameleon_system_core.module.navigation_tree_single_select"
(behind class)

All the dependencies except the request_stack can and must be removed then here.

UlrichKu:
It is allowed that child classes (services) have additional parameters. But they must be last - like request_stack in the current examle.

The first two constructor parameters just didn't fit the parent constructor anymore - after the change there.

src/CoreBundle/Resources/config/services.xml
ckleis: OK, done.

PR #500, opened 2020-09-07
mkdir error details with a exception

kzorluoglu:
| Q             | A
| ------------- | ---
| Branch        | 7.0.x
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes  
| Fixed issues  | chameleon-system/chameleon-system#632
| License       | MIT

When we try to create a new folder, we get only: " Warning: mkdir(): No such file or directory" as Error Message, but its without Path, Details and etc.

This Fix, pack the Folder Path and the Error,which mkdir() function returned, informations in an new Exception. And then we can see, which Path/Folder is wrong/missing in the Project.

### DIFF
--- a/src/CmsFileManagerBundle/objects/TPkgCmsFileManager_FileSystem.class.php
+++ b/src/CmsFileManagerBundle/objects/TPkgCmsFileManager_FileSystem.class.php
@@ -112,7 +112,11 @@ class TPkgCmsFileManager_FileSystem implements IPkgCmsFileManager
             trigger_error('File permissions are handled by the administrator. Do not try to set the mode by hand.', E_USER_DEPRECATED);
         }
 
-        return mkdir($path, 0777, $recursive);
+        try {
+            return mkdir($path, 0777, $recursive);
+        } catch (Exception $e) {
+            throw new Exception(sprintf("Cannot creating folder %s. Error: %s", $path, $e->getMessage()));
+        }
     }
 
     /**
--- a/src/CmsFileManagerBundle/objects/TPkgCmsFileManager_FileSystem.class.php
+++ b/src/CmsFileManagerBundle/objects/TPkgCmsFileManager_FileSystem.class.php
@@ -115,7 +115,7 @@ class TPkgCmsFileManager_FileSystem implements IPkgCmsFileManager
         try {
             return mkdir($path, 0777, $recursive);
         } catch (Exception $e) {
-            throw new Exception(sprintf("Cannot create folder %s. Error: %s", $path, $e->getMessage()));
+            throw new Exception(sprintf("Cannot create folder %s. Error: %s", $path, $e->getMessage()), $e->getCode(), $e);
         }
     }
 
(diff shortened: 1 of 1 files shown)


src/CmsFileManagerBundle/objects/TPkgCmsFileManager_FileSystem.class.php
```diff
@@ -112,7 +112,11 @@ public function mkdir($path, $mode = 0777, $recursive = false)
             trigger_error('File permissions are handled by the administrator. Do not try to set the mode by hand.', E_USER_DEPRECATED);
         }
 
-        return mkdir($path, 0777, $recursive);
+        try {
+            return mkdir($path, 0777, $recursive);
+        } catch (Exception $e) {
+            throw new Exception(sprintf("Cannot creating folder %s. Error: %s", $path, $e->getMessage()));
```
bestform, line 118: "Cannot create folder"

bestform (changes requested):

src/CmsFileManagerBundle/objects/TPkgCmsFileManager_FileSystem.class.php
kzorluoglu: Thanks for Review, fixed.

UlrichKu:
At least the stack trace of that exception should give enough information I think.

Selection_246

kzorluoglu:
you'r right, but i think, much better if you the path directly in the error message see.

bestform (changes requested):
We should not extract all the information from the cought Exception but instead just add our helpful message and provide the new Exception with the original as a parent. The constructor takes a parent exception as an argument.

kzorluoglu:
Thanks for the Information, i  just updated with previous throwable, for the exception chaining.

kzorluoglu:
and, the wrong reference ids from commits also with issue ids replaced.

src/CmsFileManagerBundle/objects/TPkgCmsFileManager_FileSystem.class.php
```diff
@@ -112,7 +112,11 @@ public function mkdir($path, $mode = 0777, $recursive = false)
             trigger_error('File permissions are handled by the administrator. Do not try to set the mode by hand.', E_USER_DEPRECATED);
         }
 
-        return mkdir($path, 0777, $recursive);
+        try {
+            return mkdir($path, 0777, $recursive);
+        } catch (Exception $e) {
+            throw new Exception(sprintf("Cannot create folder %s. Error: %s", $path, $e->getMessage()), $e->getCode(), $e);
```
UlrichKu, line 118: The part "Error %s" is not needed here.

If interesting/interested this is part of the parent exception.

kzorluoglu: okay, fixed.

PR #507, opened 2020-09-28
643 newsletter unsubscribe link email parameter

fabianbarth-esono:
| Q             | A
| ------------- | ---
| Branch        | 6.3.x
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed issues  | chameleon-system/chameleon-system#643
| License       | MIT

Instead of using the email address to get newsletter subscription from unsubscribe link, we use the ID of the subscription (= newsletter_user) now. This will only affect cases here CHAMELEON_PKG_NEWSLETTER_NEW_MODULE is false. Older links with the e-mail address will continue to work.

### DIFF
--- a/src/NewsletterBundle/objects/WebModules/MTPkgNewsletterSignoutCore/MTPkgNewsletterSignoutCore.class.php
+++ b/src/NewsletterBundle/objects/WebModules/MTPkgNewsletterSignoutCore/MTPkgNewsletterSignoutCore.class.php
@@ -17,6 +17,7 @@ class MTPkgNewsletterSignoutCore extends TUserCustomModelBase
 {
     const URL_PARAM_DATA = 'aPkgNewsletter';
     const URL_PARAM_MAIL = 'mail';
+    public const URL_PARAM_NEWSLETTER_USER_ID = \TdbPkgNewsletterUser::URL_USER_ID_PARAMETER;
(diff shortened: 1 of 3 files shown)


src/NewsletterBundle/objects/WebModules/MTPkgNewsletterSignoutCore/MTPkgNewsletterSignoutCore.class.php
```diff
@@ -73,6 +85,13 @@ protected function UnsubscribeUser($sPkgNewsletterUserMail = null, $sPkgNewslett
     {
         $bError = false;
         $aUserData = $this->global->GetUserData(self::URL_PARAM_DATA);
+        if (false === is_array($aUserData)) {
+            $aUserData = [];
+        }
+        $newsletterUserId = null;
+        if (array_key_exists(self::URL_PARAM_NEWSLETTER_USER_ID, $aUserData)) {
```
UlrichKu, line 92: true ===

fabianbarth-esono: Fixed

PR #510, opened 2020-10-02
Do not fetch length from post data if a value for $postData is given

bestform:
fixes #586

| Q             | A
| ------------- | ---
| Branch        | 7.0.x
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes 
| Fixed issues  | chameleon-system/chameleon-system#586
| License       | MIT

When constructing the SQL for the new field, the core should either value the given data OR fetch the information from the post data. By doing both it was possible to break the SQL when mixing the type from the given data and the length from the post data, ignoring the fact that an empty length value in the field conf data might be a valid value.

### DIFF
--- a/src/CoreBundle/private/library/classes/TCMSFields/TCMSField.class.php
+++ b/src/CoreBundle/private/library/classes/TCMSFields/TCMSField.class.php
@@ -745,10 +745,11 @@ class TCMSField implements TCMSFieldVisitableInterface
         $connection = $this->getDatabaseConnection();
         $inputFilterUtil = $this->getInputFilterUtil();
 
-        $lengthSet = $inputFilterUtil->getFilteredInput('length_set');
+        $lengthSet = '';
+        $cmsFieldTypeId = '';
+        $fieldDefaultValue = '';
 
         if (null !== $postData) {
-            $fieldDefaultValue = '';
             if (isset($postData['field_default_value'])) {
                 $fieldDefaultValue = $postData['field_default_value'];
             }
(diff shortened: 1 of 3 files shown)


src/CoreBundle/private/library/classes/TCMSFields/TCMSField.class.php
```diff
@@ -745,10 +745,11 @@ public function _GetSQLDefinition(&$postData = null)
         $connection = $this->getDatabaseConnection();
         $inputFilterUtil = $this->getInputFilterUtil();
 
-        $lengthSet = $inputFilterUtil->getFilteredInput('length_set');
+        $lengthSet = '';
+        $cmsFieldTypeId = '';
+        $fieldDefaultValue = '';
 
         if (null !== $postData) {
```
UlrichKu, line 752: Rename this variable?

It is highly confusing with respect to the special case $inputFilterUtil...

src/CoreBundle/private/library/classes/TCMSFields/TCMSField.class.php
```diff
@@ -759,6 +760,7 @@ public function _GetSQLDefinition(&$postData = null)
         } else {
```
UlrichKu, line 760: What happens if $lengthSet remains empty?

bestform: That is ok. Not every type has a length. That was the problem in the first place: it was set for a type, where it should have been empty.

src/CoreBundle/private/library/classes/TCMSFields/TCMSField.class.php
bestform: Done. It is now called `fieldDefinition`

PR #511, opened 2020-10-07
ref #531: show error when SQL error occurs due to table settings conf…

fabianbarth-esono:
…iguration error

| Q             | A
| ------------- | ---
| Branch        | 6.3.x
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed issues  | chameleon-system/chameleon-system#531
| License       | MIT

### DIFF
--- a/src/CoreBundle/Resources/translations/admin.en.xliff
+++ b/src/CoreBundle/Resources/translations/admin.en.xliff
@@ -2606,6 +2606,11 @@
                 <source>chameleon_system_core.navigation_tree.cancel_selection</source>
                 <target>Cancel</target>
             </trans-unit>
+
+            <trans-unit id="chameleon_system_core.record_list.sql_error">
+                <source>chameleon_system_core.record_list.sql_error</source>
+                <target>A database error occurred creating the record list. Most likely, this is a configuration error in table settings. The error has been logged.</target>
+            </trans-unit>
         </body>
     </file>
 </xliff>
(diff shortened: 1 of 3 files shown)


src/CoreBundle/Resources/translations/admin.en.xliff
```diff
@@ -2606,6 +2606,11 @@
                 <source>chameleon_system_core.navigation_tree.cancel_selection</source>
                 <target>Cancel</target>
             </trans-unit>
+
+            <trans-unit id="chameleon_system_core.record_list.sql_error">
+                <source>chameleon_system_core.record_list.sql_error</source>
+                <target>A database error occurred creating the record list. Most likely, this is a configuration error in table settings. The error has been logged.</target>
```
UlrichKu, line 2612: Isn't this "most likely" part a bit too specific?

I think "A database error occurred while creating the record list." would be sufficient.

That it is logged (and thus there are more informations to be found) should be general consensus for errors, I guess.

fabianbarth-esono: My experience is, that errors are usually not logged. Including the information that is actually is logged here, could potentially  save someone quite some time. 
I don't know if a more general message would be enough to immediately make you think of the table settings and I don't see the advantage of a less specific message?

PR #513, opened 2020-10-26
644 color picker

fabianbarth-esono:
| Q             | A
| ------------- | ---
| Branch        | 7.0.x
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed issues  | chameleon-system/chameleon-system#644
| License       | MIT

### DIFF
--- a/src/CoreBundle/private/library/classes/TCMSFields/TCMSFieldColorpicker.class.php
+++ b/src/CoreBundle/private/library/classes/TCMSFields/TCMSFieldColorpicker.class.php
@@ -21,13 +21,13 @@ class TCMSFieldColorpicker extends TCMSField
             $value = '#'.$value;
         }
 
-        return '<div class="input-group input-group-sm" id="colorPickerContainer'.TGlobal::OutHTML($this->name).'">
-    <span class="input-group-append">
-        <span class="input-group-text colorpicker-input-addon"><i></i></span>
-    </span>    
-    <input type="text" value="'.TGlobal::OutHTML($value).'" class="form-control" id="'.TGlobal::OutHTML($this->name).'" name="'.TGlobal::OutHTML($this->name).'" />
-</div>
-';
+        return '
+            <div class="input-group input-group-sm" id="colorPickerContainer'.TGlobal::OutHTML($this->name).'">
+                <input type="hidden" value="'.TGlobal::OutHTML(
+                $value
+            ).'" id="'.TGlobal::OutHTML($this->name).'" name="'.TGlobal::OutHTML($this->name).'">
+            </div>
+        ';
     }
 
     public function _GetHTMLValue()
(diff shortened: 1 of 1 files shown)


src/CoreBundle/private/library/classes/TCMSFields/TCMSFieldColorpicker.class.php
```diff
@@ -59,11 +59,51 @@ public function GetCMSHtmlFooterIncludes()
...
-    </script>";
+            $(function() {
+                $('#colorPickerContainer".TGlobal::OutHTML($this->name)."').colorpicker({
+                  format: 'hex', 
+                  useHashPrefix: true,
+                  autoInputFallback: false,
+                  inline: true,
+                  container: true,
+                  template: '<div class=\"colorpicker\">' +
+                  '<div class=\"colorpicker-saturation\"><i class=\"colorpicker-guide\"></i></div>' +
```
UlrichKu, line 72: Quoting of " is necessary here?

src/CoreBundle/private/library/classes/TCMSFields/TCMSFieldColorpicker.class.php
```diff
@@ -59,11 +59,51 @@ public function GetCMSHtmlFooterIncludes()
     {
         $includes = parent::GetCMSHtmlFooterIncludes();
         $includes[] = '<script src="'.TGlobal::GetStaticURLToWebLib('/javascript/bootstrap-colorpicker-3.0.3/js/bootstrap-colorpicker.min.js').'" type="text/javascript"></script>';
+
         $includes[] = "<script>
-    $(function() {
-        $('#colorPickerContainer".TGlobal::OutHTML($this->name)."').colorpicker({ format: 'hex', useHashPrefix: true, autoInputFallback: false });
-    });
-    </script>";
+            $(function() {
```
UlrichKu, line 64: You could use a "text block" (<<<JS) here?

src/CoreBundle/private/library/classes/TCMSFields/TCMSFieldColorpicker.class.php
fabianbarth-esono: Yes, the quoting is php, not JS

src/CoreBundle/private/library/classes/TCMSFields/TCMSFieldColorpicker.class.php
fabianbarth-esono: I could, but I don't know if it does us any favors because there are a couple of PHP expressions in there. These would have to become variables which wouldn't make much sense on their own, as they would contain parts of the JavaScript code.

Do you think I should change it nonetheless?

UlrichKu: I was also wondering about that.

But if there is no easy solution for these expressions I wouldn't do it.

PR #516, opened 2020-11-18
ref #591: Fix condition for insert case

UlrichKu:
| Q             | A
| ------------- | ---
| Branch        | 6.3.x
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed issues  | chameleon-system/chameleon-system#591
| License       | MIT

### DIFF
--- a/src/CoreBundle/private/library/classes/TCMSTableEditor/TCMSTableEditorEndPoint.class.php
+++ b/src/CoreBundle/private/library/classes/TCMSTableEditor/TCMSTableEditorEndPoint.class.php
@@ -1779,15 +1779,13 @@ class TCMSTableEditorEndPoint
 
         if ($databaseChanged) {
             $this->LoadDataFromDatabase();
-            if (true === $bIsUpdateCall && true === $this->isRecordingActive() && \count($dataForChangeRecorder) > 0) {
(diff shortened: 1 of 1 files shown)


src/CoreBundle/private/library/classes/TCMSTableEditor/TCMSTableEditorEndPoint.class.php
```diff
@@ -1779,15 +1779,13 @@ protected function _WriteDataToDatabase(&$oFields, &$oData, $bDataFromDatabase =
...
-            if (true === $bIsUpdateCall && true === $this->isRecordingActive() && \count($dataForChangeRecorder) > 0) {
-                if (\count($dataForChangeRecorder) > 0) {
-                    $this->writePostWriteLogChangeData(
-                        $bIsUpdateCall,
-                        $dataForChangeRecorder,
-                        $whereConditions,
-                        $setLanguageFields
-                    );
-                }
+            if (true === $this->isRecordingActive() && \count($dataForChangeRecorder) > 0) {
```
j6s, line 1782: Very minor detail here: Wouldn't it make more sense to swap the conditions since `count` is the less expensive operation?

PR #517, opened 2020-11-24
Cleanup order list initialization in TCMSListManagerFullGroupTable

j6s:
| Q             | A
| ------------- | ---
| Branch        | 6.3
| Bug fix?      | no
| New feature?  | no 
| BC breaks?    | no     
| Deprecations? | no 
| Fixed issues  | https://github.com/chameleon-system/chameleon-system/issues/657   
| License       | MIT

While debugging I found this instance, where `$aOrderData` is initialized only if a certain condition applies
but never used when that condition applies - the only usage of it is it's default value.


j6s:
Opening new PR with correct branch name

PR #519, opened 2020-11-24
ref #657: Cleanup order list initialization in TCMSListManagerFullGroupTable

j6s:
| Q             | A
| ------------- | ---
| Branch        | ~6.3~ 7.0
| Bug fix?      | no
| New feature?  | no 
| BC breaks?    | no     
| Deprecations? | no 
| Fixed issues  | https://github.com/chameleon-system/chameleon-system/issues/657   
| License       | MIT

While debugging I found this instance, where `$aOrderData` is initialized only if a certain condition applies
but never used when that condition applies - the only usage of it is it's default value.

### DIFF
--- a/src/CoreBundle/private/library/classes/TCMSListManager/TCMSListManagerFullGroupTable.class.php
+++ b/src/CoreBundle/private/library/classes/TCMSListManager/TCMSListManagerFullGroupTable.class.php
@@ -54,15 +54,10 @@ class TCMSListManagerFullGroupTable extends TCMSListManager
             $cachedTableObj = $this->getTableFromSessionCache($listCacheKey);
         }
 
-        $aOrderData = array();
-        if (null !== $cachedTableObj) {
-            $aOrderData = $cachedTableObj->orderList;
-        }
-
         // table is not in cache, load it
         if (null === $cachedTableObj) {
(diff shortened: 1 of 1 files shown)


src/CoreBundle/private/library/classes/TCMSListManager/TCMSListManagerFullGroupTable.class.php
```diff
@@ -54,15 +54,10 @@ public function Init(&$oTableConf)
             $cachedTableObj = $this->getTableFromSessionCache($listCacheKey);
         }
 
-        $aOrderData = array();
-        if (null !== $cachedTableObj) {
-            $aOrderData = $cachedTableObj->orderList;
```
UlrichKu, line 59: Couldn't the $cachedTableObj contain something?

j6s: It can, but `$aOrderData` is only ever used in the statement below, that has `if (null === $cachedTableObj)` wrapped around it - which means it is only ever used in it's initial, empty array state.

UlrichKu: Ah, I see: it is properly used in the else case that follows.

PR #521, opened 2020-11-26
Fix/538 add space between icon and name of tcmsfieldcmsuser field

FinnBrackrock:
| Q             | A
| ------------- | ---
| Branch        | 7.0.x
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no     
| Deprecations? | no 
| Tests pass?   | yes    
| Fixed issues  | chameleon-system/chameleon-system#538
| License       | MIT

Just added a space at the right place, so the field does not look all squished up


bestform:
@FinnBrackrock The oldest version we will support from now on will be 7.0.x. Could you please update the request accordingly? Thanks.

PR #522, opened 2020-11-27
ref #661: Allow cross domain user impersonation

j6s:
| Q             | A
| ------------- | ---
| Branch        | master
| Bug fix?      | no
| New feature?  | yes 
| BC breaks?    | no     
| Deprecations? | no 
| Tests pass?   | yes    
| Fixed issues  | chameleon-system/chameleon-system#661   
| License       | MIT

This adds the ability to impersonate users from the backend even if the
portal runs on a different URL. This is achieved using JWT inspired
short lived temporary tokens transmitted in URLs in order to initialize
the session on the correct domain.


src/ExtranetBundle/LoginByTransferToken/TransferTokenService.php
```diff
@@ -0,0 +1,82 @@
...
+            $this->secret,
+            0,
+            $this->initializationVector()
+        );
+
+        return base64_encode($encrypted);
+    }
+
+    /** {@inheritdoc} */
+    public function validateTransferToken(string $token): ?string
```
bestform, line 48: the name of this method does not imply that it will return a userid on success. Sure, it does validate the token in the process but a caller might expect a boolean return value.

src/ExtranetBundle/Resources/config/services.xml
```diff
@@ -35,6 +39,31 @@
...
+            <argument type="service" id="chameleon_system_core.page_service" />
+            <argument type="service" id="request_stack" />
+        </service>
+        <service
+                id="chameleon_system_extranet.login_by_transfer_token.transfer_token_service"
+                class="ChameleonSystem\ExtranetBundle\LoginByTransferToken\TransferTokenService"
+                public="true"
+        >
+            <argument type="service" id="chameleon_system_core.system_time_provider" />
+            <argument>%kernel.secret%</argument>
```
bestform, line 64: I know the secret is meant to be used like this but I see a potential security issue here, because in a default installation this token is not created randomly. It is set to a concrete string like "pleaseChangeMe". So I would suspect that there will be installations that use a predictable secret. Of course this is a user error but it can be exploited quite easily if the attacker just tries the default value. So we should either use something else as the secret or fail to create tokens if the secret has the default value.

src/ExtranetBundle/objects/TCMSTableEditor/TableEditorExtranetUser.php
```diff
@@ -35,58 +40,111 @@ protected function GetCustomMenuItems()
...
         }
-        $extranetUser->DirectLoginWithoutPassword($extranetUser->GetName(), $portal->id);
-        $url = $this->getPageService()->getLinkToPortalHomePageRelative(array(), $portal);
+
+        return TdbCmsPortal::GetNewInstance($extranetUser->fieldCmsPortalId);
+    }
+
+    private function redirectUserToTokenLoginOnPortal(string $userId, TdbCmsPortal $portal): void
+    {
+        $token = $this->transferTokenService()->createTransferTokenForUser($userId, 120);
```
bestform, line 139: I guess the default time could be shorter. As the user will log in immediately after creating the token, 120 seconds seems a bit long. The token is very sensitive data and if leaked somehow 120 seconds is plenty time for an attacker to exploit.

bestform (changes requested):

src/ExtranetBundle/LoginByTransferToken/TransferTokenService.php
j6s: Fair enonugh - I renamed it to `getUserIdFromTransferToken` which should communicate that a bit better.

src/ExtranetBundle/objects/TCMSTableEditor/TableEditorExtranetUser.php
j6s: :heavy_check_mark: Lowered it to 20 seconds

src/ExtranetBundle/Resources/config/services.xml
j6s: Is there a unique per installation, constant over time property we could feasibly use for this?

j6s: :heavy_check_mark: 

* The service now fails to decode a token if the secret is the default one, logging that failure
* The button in the table editor also is disabled now if that's the case.

src/ExtranetBundle/LoginByTransferToken/TransferTokenService.php
```diff
@@ -92,6 +105,21 @@ private function decodeToken(string $token): ?array
...
     }
 
+    private function isSecretIsDefaultFromQuickstartTemplate(): bool
+    {
+        if ('!ThisTokenIsNotSoSecretChangeIt!' !== $this->secret) {
+            return false;
+        }
+
+        $this->logger->error('
+            Refusing to encode or decode transfer transfer token with default secret.
```
bestform, line 115: "transfer transfer" :)

src/ExtranetBundle/LoginByTransferToken/TransferTokenService.php
```diff
@@ -92,6 +105,21 @@ private function decodeToken(string $token): ?array
         return $data;
     }
 
+    private function isSecretIsDefaultFromQuickstartTemplate(): bool
```
bestform, line 108: I find it kindof irritating to repeat the "is" in the method name. Wouldn't it be easier to read `isSecretDefaultFromQuickstartTemplate`?

bestform (changes requested):

src/ExtranetBundle/objects/TCMSTableEditor/TableEditorExtranetUser.php
```diff
@@ -210,4 +215,11 @@ private function router(): PortalAndLanguageAwareRouterInterface
     {
         return \ChameleonSystem\CoreBundle\ServiceLocator::get('chameleon_system_core.router.chameleon_frontend');
     }
+
+    private function isKernelSecretDefault(): bool
+    {
+        $secret = \ChameleonSystem\CoreBundle\ServiceLocator::getParameter('kernel.secret');
+
+        return '!ThisTokenIsNotSoSecretChangeIt!' === $secret;
```
bestform, line 223: Maybe we should define this string in a constant somewhere. I do not mind the duplicated logic but the duplicated string should be avoided.

bestform (changes requested):

src/ExtranetBundle/LoginByTransferToken/TransferTokenService.php
j6s: :heavy_check_mark: Fixed

src/ExtranetBundle/LoginByTransferToken/TransferTokenService.php
j6s: That's what happens when you are unsure which naming scheme you want to use: You use both at the same time.

:heavy_check_mark: Fixed by renaming the method to `isReadyToEncodeTokens` which also resolves the issue of duplication

src/ExtranetBundle/objects/TCMSTableEditor/TableEditorExtranetUser.php
j6s: :heavy_check_mark: The service now has a `isReadyToEncodeTokens` methods which is used in the talbe editor in order to prevent duplicate logic

src/ExtranetBundle/LoginByTransferToken/LoginController.php
```diff
@@ -0,0 +1,74 @@
...
+namespace ChameleonSystem\ExtranetBundle\LoginByTransferToken;
+
+use ChameleonSystem\CoreBundle\Service\PageServiceInterface;
+use ChameleonSystem\ExtranetBundle\Interfaces\ExtranetUserProviderInterface;
+use Symfony\Component\HttpFoundation\RedirectResponse;
+use Symfony\Component\HttpFoundation\RequestStack;
+use Symfony\Component\HttpFoundation\Response;
+use Symfony\Component\HttpKernel\Exception\AccessDeniedHttpException;
+
+class LoginController
```
UlrichKu, line 12: LoginByTokenController ?

Otherwise it suggests that every login is handled here.

src/ExtranetBundle/LoginByTransferToken/LoginController.php
```diff
@@ -0,0 +1,74 @@
...
+{
+    /** @var TransferTokenServiceInterface */
+    protected $transferTokenService;
+
+    /** @var ExtranetUserProviderInterface */
+    protected $extranetUserProvider;
```
UlrichKu, line 18: They don't need to be protected.

src/ExtranetBundle/LoginByTransferToken/RouteGenerator.php
```diff
@@ -0,0 +1,30 @@
...
+namespace ChameleonSystem\ExtranetBundle\LoginByTransferToken;
+
+use esono\pkgCmsRouting\CollectionGeneratorInterface;
+use Symfony\Component\Routing\Route;
+use Symfony\Component\Routing\RouteCollection;
+
+class RouteGenerator implements CollectionGeneratorInterface
+{
+    public const ROUTE_NAME = 'chameleon_system_extranet.login_by_transfer_token';
+    protected const CONTROLLER_SERVICE = 'chameleon_system_extranet.login_by_transfer_token.login_controller';
```
UlrichKu, line 12: No protected needed (?)

src/ExtranetBundle/LoginByTransferToken/TransferTokenService.php
```diff
@@ -0,0 +1,130 @@
...
+        $data = json_decode($json, true);
+        if (false === is_array($data)) {
+            return null;
+        }
+
+        return $data;
+    }
+
+
+    public function isReadyToEncodeTokens(): bool
```
UlrichKu, line 111: Must this be public? (No inheritdoc, called here)

src/ExtranetBundle/LoginByTransferToken/TransferTokenService.php
```diff
@@ -0,0 +1,130 @@
+<?php
+
+namespace ChameleonSystem\ExtranetBundle\LoginByTransferToken;
+
+use ChameleonSystem\CoreBundle\Interfaces\TimeProviderInterface;
+use Psr\Log\LoggerInterface;
+
+class TransferTokenService implements TransferTokenServiceInterface
```
UlrichKu, line 8: License header is missing.

src/ExtranetBundle/LoginByTransferToken/TransferTokenService.php
```diff
@@ -0,0 +1,130 @@
...
+        if ($this->timeProvider->getUnixTimestamp() > $data['expires']) {
+            return null;
+        }
+
+        return $data['userId'];
+    }
+
+    private function encodeToken(array $data): string
+    {
+        $data['_salt'] = md5(random_bytes(16));
```
UlrichKu, line 67: Why do secret things need to be md5 ?

Also in initializationVector().

src/ExtranetBundle/objects/TCMSTableEditor/TableEditorExtranetUser.php
```diff
@@ -35,58 +42,119 @@ protected function GetCustomMenuItems()
...
+        $extranetUser = $this->loadExtranetUser($userId);
+        $portal = $this->portalForExtranetUser($extranetUser);
+        $this->redirectUserToTokenLoginOnPortal($userId, $portal);
+    }
+
+    private function isBackendUserLoggedInWithPermission(string $permission): bool
+    {
+        $cmsUser = TCMSUser::GetActiveUser();
+
+        return $cmsUser && $cmsUser->oAccessManager && $cmsUser->oAccessManager->PermitFunction($permission);
```
UlrichKu, line 118: null !== 

true ===

src/ExtranetBundle/LoginByTransferToken/TransferTokenService.php
```diff
@@ -0,0 +1,130 @@
+<?php
```
UlrichKu, line 1: Naming: What does this transfer? It sounds as if it transfers a token?

src/ExtranetBundle/LoginByTransferToken/LoginController.php
j6s: Fair.

:heavy_check_mark: Renamed to `LoginByTokenController`

src/ExtranetBundle/LoginByTransferToken/LoginController.php
j6s: :heavy_check_mark: Using `private` in place of `protected`. Old habits coming through here.

src/ExtranetBundle/LoginByTransferToken/RouteGenerator.php
j6s: :heavy_check_mark: Using `private` in place of `protected`. Old habits coming through here.

src/ExtranetBundle/LoginByTransferToken/TransferTokenService.php
j6s: Yes, it needs to be `public` and it's also part of the interface. It is also called in `TableEditorExtranetUser` in order to disable the login button if the transfer token service is not ready.

:heavy_check_mark: Added `@inheritDoc` annotation

src/ExtranetBundle/LoginByTransferToken/TransferTokenService.php
j6s: :heavy_check_mark: Added license headers in all files

src/ExtranetBundle/LoginByTransferToken/TransferTokenService.php
j6s: It doesn't have to be - the important part of this is converting the output of `random_bytes` into a regular string that does not have any weirdness in it. We don't have to do that, but doing so prevents future headaches when some invalid character is in the string. Basically what this does is generate a random hex-string.

We could use `base64encode` in these places - however, `md5` was chosen because it is a very well optimized implementation of this behavior.

src/ExtranetBundle/objects/TCMSTableEditor/TableEditorExtranetUser.php
j6s: :heavy_check_mark: Now using yoda-style comparisons

src/ExtranetBundle/LoginByTransferToken/TransferTokenService.php
j6s: It's a service that handles tokens in order to log extranet users in - in order to transfer between two domains for example. If this was a service to transfer tokens I would expect it to be named `TokenTransferService`.

src/ExtranetBundle/LoginByTransferToken/RouteGenerator.php
```diff
@@ -0,0 +1,39 @@
...
+    /**
+     * @param array $config
+     *
+     * @return RouteCollection
+     */
+    public function getCollection($config, \TdbCmsPortal $portal, \TdbCmsLanguage $language)
+    {
+        $routeCollection = new RouteCollection();
+        $routeCollection->add(self::ROUTE_NAME, new Route(
+            '/_login_by_transfer_token_/{token}',
```
UlrichKu, line 33: Wouldn't a simple routing.yml (without RouteGenerator class) be sufficient here?

src/ExtranetBundle/LoginByTransferToken/TransferTokenService.php
UlrichKu: So the main purpose is login?

LoginByTokenService?

"Transfer" would be an edge case / implementation detail?

src/ExtranetBundle/LoginByTransferToken/RouteGenerator.php
j6s: I copied that pattern from somewhere else because at the time I didn't know that `pkg_cms_routing` also supported YAML files.

:heavy_check_mark: Fixed

src/ExtranetBundle/LoginByTransferToken/TransferTokenService.php
UlrichKu: ok

j6s: :heavy_check_mark: Renamed

PR #524, opened 2020-12-07
ref 609: add outline: none; to nav item, so that the weird looking ou…

FinnBrackrock:
…tline, that chrome adds by default if an element is focused, disapears

| Q             | A
| ------------- | ---
| Branch        | 7.0.x
| Bug fix?      | yes
| New feature?  |no 
| BC breaks?    | no     
| Deprecations? | no 
| Tests pass?   | yes
| Fixed issues  | chameleon-system/chameleon-system#609
| License       | MIT


bestform:
As we do not support 6.x anymore, could you please update this PR to go into 7.0.x? Thanks.  @FinnBrackrock

PR #525, opened 2020-12-08
ref #652: Replace named constructors with __construct

j6s:
| Q             | A
| ------------- | ---
| Branch        | 7.1
| Bug fix?      | no
| New feature?  | no 
| BC breaks?    | no     
| Deprecations? | yes 
| Fixed issues  | chameleon-system/chameleon-system#652   
| License       | MIT

Replaces all deprecated named constructors with `__construct` and introduces a backwards compatibility shim in order to prevent breakage of subclasses still calling the old constructor. The shim will call the new constructor and log a deprecation warning into a new `depcrecation` channel.


pixeljunkie:
This change should move to 7.1/master i would suggest.

PR #528, opened 2020-12-22
Fix/667 record mlt insertion

j6s:
| Q             | A
| ------------- | ---
| Branch        | 6.3
| Bug fix?      | yes
| New feature?  | no 
| BC breaks?    | no     
| Deprecations? | no 
| Tests pass?   | yes    
| Fixed issues  | chameleon-system/chameleon-system#667   
| License       | MIT

In `TCMSTableEditorManager::Insert`, instead of manually speaking to the database in order to insert MLT fields, The wrapped TableEditor is used to do so as that method also informs the migration recorder, if active


bestform:
This looks good to me. Could you edit the target branch to 7.0.x as this is now the oldest version we actively support? Thanks.

src/CoreBundle/private/library/classes/TCMSTableEditor/TCMSTableEditorManager.class.php
```diff
@@ -502,6 +496,16 @@ public function RemoveMLTConnection($sFieldName, $iConnectedID)
         }
     }
 
+    private function addInverseRestrictingMLTConnection(string $restrictionField, string $restriction): void
+    {
+        $sourceTable = substr($restrictionField, 0, -4);
```
UlrichKu, line 501: MLT connections can also be configured (?) then the name is arbitrary and doesn't have to end with _mlt.

There should be some method somewhere there gives the (target) table name for a field.

j6s: I copied the code over from where it was before - but you are right, this cannot work with customized MLT Table names

j6s: Surprisingly, this part of the code works fine: The iframe for the relation is the table editor with a restriction field added to it. This restriction field does not actually exist on the target table but correctly resolves to the source table's table editor which is then used to add the relation. However, this part a bit further down does not work as expected:

```php
$tableEditor = new TCMSTableEditorManager();
$tableEditor->Init(TTools::GetCMSTableId($sourceTable), $restriction);
$tableEditor->AddMLTConnection($targetTable . '_mlt', $this->sId);
```

Specifically the last line expects the column to adhere to the `{TARGET_TABLE}_mlt` naming convention.

j6s: :heavy_check_mark: Fixed by also passing through the `field` property containing the origin field name and using that in the table editor. Can you carefully review https://github.com/chameleon-system/chameleon-base/pull/528/commits/c25392998cb0a4db6f7ef6cf3313275b3e2982bb in particular since I am not 100% sure which unintended consequences the change to `TCMSListManagerEndPoint`  could have?

src/CoreBundle/private/library/classes/TCMSListManager/TCMSListManagerEndPoint.class.php
```diff
@@ -545,7 +545,7 @@ public function GetHiddenFieldsHook()
     {
         $oGlobal = TGlobal::instance();
         $aAdditionalParameterData = array();
-        $aAdditionalParameters = array('sRestrictionField', 'sRestriction', 'bIsLoadedFromIFrame');
+        $aAdditionalParameters = array('sRestrictionField', 'sRestriction', 'bIsLoadedFromIFrame', 'field');
```
UlrichKu, line 548: Is field different from sRestrictionField?

j6s: Yes - but it's not obvious from the naming.

* `sRestrictionField` contains the name of the field that contains the restriction on the target table (The field pointing back to the 'parent')
* `field` contains the name of the field on the 'parent record' of which the children are being edited

Basically, if you have a `article` that has an MLT relation to `category` in the field `special_categories`

* `sRestrictionField` will be `article_id` on `category` (even if that field doesn't "really" exist)
* `field` will refer to `special_categories` on `article`

PR #530, opened 2021-01-07
update version contraints to include newer versions

bestform:
| Q             | A
| ------------- | ---
| Branch        | 7.1.x
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed issues  | chameleon-system/chameleon-system#670 / chameleon-system/chameleon-system#671
| License       | MIT

The version constraints have been loosened to allow patch level updates. For the dbal dependency the constraint has been set to how it was in previous chameleon versions so it includes minor versions.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -17,7 +17,7 @@
         "ckeditor/ckeditor": "4.6.2",
         "coreui/coreui": "^2.1",
         "dmhendricks/file-icon-vectors": "~1.0",
-        "doctrine/dbal": "~2.9.0",
+        "doctrine/dbal": "~2.9",
         "doctrine/doctrine-bundle": "2.0.7",
         "ext-curl": "*",
         "ext-pdo": "*",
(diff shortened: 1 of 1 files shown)


composer.json
```diff
@@ -17,7 +17,7 @@
         "ckeditor/ckeditor": "4.6.2",
         "coreui/coreui": "^2.1",
         "dmhendricks/file-icon-vectors": "~1.0",
-        "doctrine/dbal": "~2.9.0",
+        "doctrine/dbal": "~2.9",
```
rnuernberg, line 20: this will result in composer update automatically pulling every minor version. I would have assumed, that we only want to update patch level updates?

bestform: This is how it was in previous versions. I guess this should be ok, but you may convince me otherwise.

UlrichKu: (That's why it can have been 2.12 as noted in the ticket https://github.com/chameleon-system/chameleon-system/issues/670)

rnuernberg: ok - I had a look hat the dbal planing. 2.x should remain backward compatible. So I guess ~2.9 is ok. However, I would suggest that we require ~2.12 since it contains a lot of changes that aid in preparing an upgrade to 3.0 later. That way we can start to prepare for the 3.0 upgrade.

PR #531, opened 2021-01-14
fix: make trusted proxy handling compatible with symfony 4

bestform:
| Q             | A
| ------------- | ---
| Branch        | 7.1.x
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no (see comment about custom header names
| Deprecations? | no
| Tests pass?   | yes
| Fixed issues  | chameleon-system/chameleon-system#669
| License       | MIT

Custom header names aren't supported anymore in symfony so we will
remove them as well. If one really needs to handle this edge case it is
advisable to map to the default names in the entry point.

### DIFF
--- a/src/CoreBundle/ChameleonHttpKernel.php
+++ b/src/CoreBundle/ChameleonHttpKernel.php
@@ -21,22 +21,6 @@ class ChameleonHttpKernel extends HttpKernel
      * @var string
      */
     private $trustedProxies;
-    /**
-     * @var string
-     */
-    private $trustedHeaderClientIp;
-    /**
-     * @var string
-     */
-    private $trustedHeaderClientHost;
-    /**
-     * @var string
-     */
-    private $trustedHeaderClientPort;
-    /**
-     * @var string
-     */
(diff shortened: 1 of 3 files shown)


src/CoreBundle/ChameleonHttpKernel.php
```diff
@@ -53,23 +37,10 @@ public function handle(Request $request, $type = HttpKernelInterface::MASTER_REQ
                 }
             }
             if (count($aTrustedProxies) > 0) {
-                Request::setTrustedProxies($aTrustedProxies);
+                Request::setTrustedProxies($aTrustedProxies, Request::getTrustedHeaderSet());
```
UlrichKu, line 40: Request::getTrustedHeaderSet() is never set to any value?

bestform: Yes it is. Any service can set it and if it hasn't been set, yet, it is the default. Because this is a mandatory parameter and we do not want to change it here, we just provide the current state.

PR #533, opened 2021-01-21
Feature/668 php8 compatibility

j6s:
| Q             | A
| ------------- | ---
| Branch        | 7.1.x
| Bug fix?      | no
| New feature?  | no 
| BC breaks?    | potentially (see below)     
| Deprecations? | no 
| Tests pass?   | yes    
| Fixed issues  | chameleon-system/chameleon-system#668   
| License       | MIT

Changes for supporting PHP8 in chameleon. Most of the code changes are PHPUnit updates.

### Potential backwards compatibility changes

There is one change in actual code, that is important here:
Before:
```php
    public static function StripTextWordSave($length = null, $sText, $sAddToText = '')
```

After:
```php
    public static function StripTextWordSave($length, $sText, $sAddToText = '')
```

declaring optional arguments before non-optional ones is not valid anymore in PHP8. However, I don't quite know if this changes behaviour in PHP7.x.


composer.json
```diff
@@ -27,18 +27,18 @@
         "kartik-v/strength-meter": "~1.1.0",
         "matthiasmullie/minify": "~1.3.0",
         "monolog/monolog": "~1.24",
-        "php": "^7.1",
+        "php": "^7.1|^8.0",
```
UlrichKu, line 30: (tbd Should this already be 7.4 for example?)

j6s: Bumped it to 7.4 and updated the minimum versions of dependencies accordingly.

rnuernberg:
I can not see how the change in TTools causes problems in 7.4. The only way to trigger the default would be to call the method with no parameters - which will trigger a warning anyway since the second parameter is not optional.

PR #536, opened 2021-01-27
648 use module views path

UlrichKu:
| Q             | A
| ------------- | ---
| Branch        | 7.0.x
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed issues  | chameleon-system/chameleon-system#648
| License       | MIT

### DIFF
--- a/src/ViewRendererBundle/objects/helper/TPkgViewRendererSnippetDirectory.class.php
+++ b/src/ViewRendererBundle/objects/helper/TPkgViewRendererSnippetDirectory.class.php
@@ -389,6 +389,9 @@ class TPkgViewRendererSnippetDirectory implements TPkgViewRendererSnippetDirecto
                 $snippetChain = $oTheme->getSnippetChainAsArray();
             }
         }
+
+        // TODO quite unfortunate here that there is no "this is backend" (only the fact that is has no portal and thus no snippet chain paths)
+
         if (false !== CHAMELEON_PATH_THEMES && count($snippetChain) > 0) {
             $aBasePaths[$sPortalCacheId] = array();
             foreach ($snippetChain as $element) {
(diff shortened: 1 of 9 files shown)


src/ViewRendererBundle/objects/helper/TPkgViewRendererSnippetDirectory.class.php
```diff
@@ -408,6 +408,22 @@ public function getBasePaths($oPortal = null, $sBaseDirectory = null)
             }
         }
 
+        if (true === $this->requestInfoService->isBackendMode()) {
+            $allBundles = $this->kernel->getBundles();
```
j6s, line 412: In the interest of keeping methods easy to read, it would probably make sense to extract this block of code into a method, so it becomes something like

```php
if (true === $this->requestInfoService->isBackendMode()) {
    $aBasePaths[$sPortalCacheId] = array_merge(
        $aBasePaths[$sPortalCacheId],
        $this->getBasePathsForBackendModules()
    );
}
```

UlrichKu: Move (all) to a method.

j6s (approved):

src/ViewRendererBundle/objects/helper/TPkgViewRendererSnippetDirectory.class.php
```diff
@@ -408,6 +408,12 @@ public function getBasePaths($oPortal = null, $sBaseDirectory = null)
             }
         }
 
+        if (true === $this->requestInfoService->isBackendMode()) {
+            $aBasePaths[$sPortalCacheId] = $this->addBundlePaths($aBasePaths[$sPortalCacheId], $sBaseDirectory);
+
+            // TODO or also use a "snippet chain" field for the backend theme? (table cms_config_themes)
```
j6s, line 414: _Discuss here_

rnuernberg: We could use themes if we assigned a theme via cms_config. 

That solution would make the handling consistent with our frontend handling - making it easier to understand. At the same time it makes it much easier to add/overwrite twig files for the backend.

rnuernberg (changes requested):

src/ViewRendererBundle/objects/helper/TPkgViewRendererSnippetDirectory.class.php
UlrichKu: Ok.

(Frontend) themeing used here.

PR #538, opened 2021-02-03
ref #641: Reset paging for different search triggers

UlrichKu:
| Q             | A
| ------------- | ---
| Branch        | 7.0.x
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed issues  | chameleon-system/chameleon-system#641
| License       | MIT

### DIFF
--- a/src/CoreBundle/Resources/public/javascript/table.js
+++ b/src/CoreBundle/Resources/public/javascript/table.js
@@ -108,7 +108,6 @@ function ChangeListMarking(fieldValue, formName) {
 }
 
 function resetPagingAndSearch(event) {
-
     if (typeof event !== "undefined" && typeof event.originalEvent !== "undefined") {
         event = event.originalEvent; // extract from jquery event
     }
@@ -142,7 +141,7 @@ $(document).ready(function () {
 
     CHAMELEON.CORE.initializeEntryAutocomplete($('input#searchLookup'));
 
-    var $searchFields = $("input[name='_search_word']");
+    var $searchFields = $(".entry-search-field");
     $searchFields.on("keypress", resetPagingAndSearch); // NOTE "input" is not called for a RETURN (which immediately sends the form)
     $searchFields.closest("form").find("input[type='button']").on("click", resetPagingAndSearch);
 });
--- a/src/CoreBundle/Resources/public/javascript/table.js
+++ b/src/CoreBundle/Resources/public/javascript/table.js
@@ -107,6 +107,30 @@ function ChangeListMarking(fieldValue, formName) {
     );
 }
 
+function resetPagingAndSearch(event) {
+
+    if (typeof event !== "undefined" && typeof event.originalEvent !== "undefined") {
+        event = event.originalEvent; // extract from jquery event
+    }
+
+    if (event instanceof KeyboardEvent) {
+        if (event.key !== "Enter") {
+            return;
+        }
+    }
+
+    $form = $(event.target).closest("form");
+
+    if (0 === $form.length) {
+        return;
+    }
+
+    event.preventDefault();
+
+    $form[0]._startRecord.value = 0;
+    $form.submit();
+}
+
 $(document).ready(function () {
     CHAMELEON.CORE.handleFormAndLinkTargetsInModals();
 
@@ -117,4 +141,8 @@ $(document).ready(function () {
     });
 
     CHAMELEON.CORE.initializeEntryAutocomplete($('input#searchLookup'));
+
+    var $searchFields = $("input[name='_search_word']");
+    $searchFields.on("keypress", resetPagingAndSearch); // NOTE "input" is not called for a RETURN (which immediately sends the form)
+    $searchFields.closest("form").find("input[type='button']").on("click", resetPagingAndSearch);
 });
--- a/src/CoreBundle/private/library/classes/TFullGroupTable/TFullGroupTable.class.php
+++ b/src/CoreBundle/private/library/classes/TFullGroupTable/TFullGroupTable.class.php
@@ -868,6 +868,19 @@ class TFullGroupTable extends TGroupTable
 
             $filterContent .= '<div class="form-group mr-2 typeahead-relative">';
 
+            $filterContent .= '<script>
+                function resetPagingAndSearch(event) {
+                    if (event !== undefined && typeof event.key !== "undefined") {
+                        if (event.key !== "Enter") {
+                            return;
+                        }
+                    }
+                    
+                    document.'.$this->listName.'._startRecord.value = 0;
+                    document.'.$this->listName.'.submit();
+                }
+                </script>';
+
             $formatString = '<input id="searchLookup" name="_search_word" class="form-control form-control-sm" placeholder="%s" value="%s" autocomplete="off" data-source-url="%s" data-record-url="%s" data-onclick-function="%s">';
             $filterContent .= sprintf(
                 $formatString,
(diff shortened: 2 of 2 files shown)


src/CoreBundle/private/library/classes/TFullGroupTable/TFullGroupTable.class.php
```diff
@@ -868,6 +868,19 @@ protected function _BuildFilterSection()
 
             $filterContent .= '<div class="form-group mr-2 typeahead-relative">';
 
+            $filterContent .= '<script>
+                function resetPagingAndSearch(event) {
+                    if (event !== undefined && typeof event.key !== "undefined") {
```
j6s, line 873: `if (event instanceof KeyboardEvent)` would make the intent a lot clearer.

src/CoreBundle/private/library/classes/TFullGroupTable/TFullGroupTable.class.php
```diff
@@ -878,11 +891,16 @@ protected function _BuildFilterSection()
                 TGlobal::OutHTML($this->onClick)
             );
 
+            // We need the key event; that is not given when simply specifying "onkeypress" in the input:
+            $filterContent .= '<script>
+                    document.querySelector("#searchLookup").addEventListener("keypress", resetPagingAndSearch);
```
j6s, line 896: Do we have cases, where multiple tables are on the page at the same time? If yes, then this would only attach a listener to the first one. In fact, if there can be multiple lists on a page, then the definition of `resetPagingAndSearch` will also be duplicated - in that case that should also be moved into `document.{LIST_NAME}`.

j6s: Also: The `keypress` event seems to be deprecated. `input` seems to be the preferred event nowadays.

src/CoreBundle/private/library/classes/TFullGroupTable/TFullGroupTable.class.php
```diff
@@ -868,6 +868,19 @@ protected function _BuildFilterSection()
 
             $filterContent .= '<div class="form-group mr-2 typeahead-relative">';
 
+            $filterContent .= '<script>
+                function resetPagingAndSearch(event) {
+                    if (event !== undefined && typeof event.key !== "undefined") {
+                        if (event.key !== "Enter") {
+                            return;
+                        }
```
j6s, line 876: If it is the enter key, `event.preventDefault()` should probably be called, just to prevent race-conditions, where the browsers default behaviour is to submit the form while we also do that programatically.

j6s (changes requested):

src/CoreBundle/private/library/classes/TFullGroupTable/TFullGroupTable.class.php
UlrichKu: ok

src/CoreBundle/private/library/classes/TFullGroupTable/TFullGroupTable.class.php
UlrichKu: ok

src/CoreBundle/private/library/classes/TFullGroupTable/TFullGroupTable.class.php
UlrichKu: No multiple tables.

Changed to input (event).

src/CoreBundle/private/library/classes/TFullGroupTable/TFullGroupTable.class.php
j6s: Alright.

j6s (approved):

src/CoreBundle/private/library/classes/TFullGroupTable/TFullGroupTable.class.php
```diff
@@ -868,6 +868,19 @@ protected function _BuildFilterSection()
 
             $filterContent .= '<div class="form-group mr-2 typeahead-relative">';
 
+            $filterContent .= '<script>
+                function resetPagingAndSearch(event) {
```
rnuernberg, line 872: The group tabled may be used multiple times on one page - the Method should be moved into the appropriate js file (`chameleon/blackbox/javascript/table.js`)

rnuernberg (changes requested):

src/CoreBundle/private/library/classes/TFullGroupTable/TFullGroupTable.class.php
UlrichKu: This "in-line" with the existing old JS that is partly replaced here.

Especially here is a need for a (static) parameter to that JS: $this->listName

Can it be mulitply on one page? I think every table is in its own iframe?

Also it does work with a single id (searchLookup) for the search field.

rnuernberg: you can get the list name by using `${this}` (you will find the list name as the input field `_listName` AND as the name and id of the containing form.

The old JS would not cause a collision since it did not define a function - the new code does. So we need to take care of that issue.

The core may currently only use one instance of the list table on one page - but the list table itself is setup in suche a way, that multiple instances of it can be used on one page. I see no reason to limit this as a side effect - especially when the problem can be solved so easily.

UlrichKu: Wasn't easy (javascript!) but I moved it to a file and made it dynamic.

src/CoreBundle/Resources/public/javascript/table.js
```diff
@@ -117,4 +141,8 @@ $(document).ready(function () {
     });
 
     CHAMELEON.CORE.initializeEntryAutocomplete($('input#searchLookup'));
+
+    var $searchFields = $("input[name='_search_word']");
```
rnuernberg, line 145: we must only enable the event on the _search_word field of the list form. Either by limiting the selection to the correct form, or by adding a css class that marks the field clearly as being the search field. Otherwise we risk adding the event listener to fields for which it is not intended.

src/CoreBundle/Resources/public/javascript/table.js
```diff
@@ -117,4 +141,8 @@ $(document).ready(function () {
...
+    $searchFields.on("keypress", resetPagingAndSearch); // NOTE "input" is not called for a RETURN (which immediately sends the form)
+    $searchFields.closest("form").find("input[type='button']").on("click", resetPagingAndSearch);
```
rnuernberg, line 147: Same point as above. we do not want to add the event to just any form on the page - but only to a full group table form. 
I would suggest adding a class to clearly mark it and then selecting based on that

rnuernberg (changes requested):

src/CoreBundle/Resources/public/javascript/table.js
UlrichKu: Did that.

However the collision likelihood is now as big or bigger as this class name could also be used somewhere else.

UlrichKu:
I noticed that there are other cases where this change here might apply: all other (search) form fields. For example every grouping dropdown (product type on product table).

I'd much rather use the "submit" event on that form. That would make the logic much simipler. However also this might not be the correct location depending on the form input element.

Also I noticed that it is now very inconsistent with the move of parts of the (new) javascript to the file. The other older half (ie grouping dropdown or "switch page") is still in TFullGroupTable.php.

src/CoreBundle/Resources/public/javascript/table.js
rnuernberg: The idea was to use a css class that is unique for this purpose. But I think this is sufficiently unique

PR #539, opened 2021-02-09
ref #681: Check registration result

j6s:
| Q             | A
| ------------- | ---
| Branch        | 7.0
| Bug fix?      | yes
| New feature?  | no 
| BC breaks?    | no     
| Deprecations? |no 
| Tests pass?   | yes    
| License       | MIT

Adds a check for the return value of `ExtranetUser::Register` to the 'success' criteria when registering a new user in order to prevent errors in the registration process from accidentally redirecting the user to the success URL while their new account was not created.

### DIFF
--- a/src/ExtranetBundle/objects/WebModules/MTExtranetCore/MTExtranetCoreEndPoint.class.php
+++ b/src/ExtranetBundle/objects/WebModules/MTExtranetCore/MTExtranetCoreEndPoint.class.php
@@ -220,7 +220,11 @@ class MTExtranetCoreEndPoint extends TUserCustomModelBase
                 }
             }
 
-            if ($bDataValid && true === $oUser->Register()) {
+            if (true === $bDataValid) {
+                $bDataValid = $oUser->Register();
+            }
+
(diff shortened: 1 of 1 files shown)


src/ExtranetBundle/objects/WebModules/MTExtranetCore/MTExtranetCoreEndPoint.class.php
```diff
@@ -220,9 +220,7 @@ public function Register($sSuccessURL = null, $sFailureURL = null)
                 }
             }
 
-            if ($bDataValid) {
-                $oUser->Register();
-
+            if ($bDataValid && true === $oUser->Register()) {
```
UlrichKu, line 223: I'd make two lines of this:

`
if (true === $bDataValid) {
    $bDataValid = $oUser->Register();
}

if (true === $bDataValid) {
....

`

j6s: ok

PR #541, opened 2021-02-25
690 show menu location in table settings

UlrichKu:
| Q             | A
| ------------- | ---
| Branch        | 7.0.x
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed issues  | chameleon-system/chameleon-system#690
| License       | MIT

### DIFF
--- a/src/CoreBundle/Field/FieldSidebarConnected.php
+++ b/src/CoreBundle/Field/FieldSidebarConnected.php
@@ -35,7 +35,7 @@ class FieldSidebarConnected extends \TCMSFieldVarchar
     /**
      * {@inheritDoc}
      */
-    public function GetReadOnly(){
+    public function GetReadOnly() {
         $menuPath = $this->getMenuPath($this->recordId);
 
         return sprintf('<div class="form-control form-control-sm" readonly>%s</div>', $menuPath);
--- a/src/CoreBundle/Field/FieldSidebarConnected.php
+++ b/src/CoreBundle/Field/FieldSidebarConnected.php
@@ -1,0 +1,89 @@
+<?php
+
+/*
+ * This file is part of the Chameleon System (https://www.chameleonsystem.com).
+ *
+ * (c) ESONO AG (https://www.esono.de)
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace ChameleonSystem\CoreBundle\Field;
+
+use ChameleonSystem\CoreBundle\DataAccess\MenuItemDataAccessInterface;
+use ChameleonSystem\CoreBundle\DataModel\MenuCategoryAndItem;
+use ChameleonSystem\CoreBundle\ServiceLocator;
+
+class FieldSidebarConnected extends \TCMSFieldVarchar
+{
+    /**
+     * @var MenuItemDataAccessInterface
+     */
+    private $menuItemDataAccess;
+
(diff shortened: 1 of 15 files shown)


src/CoreBundle/private/library/classes/dbobjects/TCMSTableConf.class.php
```diff
@@ -490,23 +490,29 @@ public function ExportTable()
...
-     * fetches the TdbFooBaa class name and loads it for current table config record.
+     * @param string|null $id - the id of the record to load, if any
      *
-     * @param string|null $recordID - record to load
-     *
-     * @return TCMSRecord
+     * @return TCMSRecord|null - null if the class does not exist or the loading of the id fails if given
      */
-    public function GetTableObjectInstance($recordID = null)
+    public function GetTableObjectInstance($id = null)
```
rnuernberg, line 497: why are we changing this method in this context? Also: we are changing the signature here, which we should only do in a minor update.

src/CoreBundle/Field/FieldSidebarConnected.php
```diff
@@ -0,0 +1,89 @@
...
+        if (null === $menuItemDataAccess) {
+            $menuItemDataAccess = ServiceLocator::get('chameleon_system_core.sidebar.menu_access');
+        }
+        $this->menuItemDataAccess = $menuItemDataAccess;
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    public function GetReadOnly(){
```
rnuernberg, line 38: use php-cs-fixer  to enforce style

src/CoreBundle/Field/FieldSidebarConnected.php
```diff
@@ -0,0 +1,89 @@
...
+        $menuPath = $this->getMenuPath($this->recordId);
+
+        return sprintf('<div class="form-control form-control-sm" readonly>%s</div>', $menuPath);
+    }
+
+    private function getMenuPath(string $tableId): string
+    {
+        $menuEntry = $this->getMatchingEntryFromMenu($tableId);
```
rnuernberg, line 46: There could be multiple assignments. Why only show one?

rnuernberg (changes requested):

src/CoreBundle/private/library/classes/dbobjects/TCMSTableConf.class.php
UlrichKu: Changes to the method:

Two checks (eg class_exists) were added. Otherwise there will be (non-prohibitable) crashes in it.

I revereted the renaming of the parameter. However that would not have had any (breaking) consequences.

src/CoreBundle/Field/FieldSidebarConnected.php
UlrichKu: Showing multiple would make it more complicated.

Also there are hardly cases with mutliple menu entries for just one table.

One use-case for this function here would be to be able to easily access the datasets of a table in the backend. For this one path (entry here) is enough.

src/CoreBundle/private/library/classes/dbobjects/TCMSTableConf.class.php
UlrichKu: However you are right that it is not really related here...

pixeljunkie:
Outdated. Will be added to 7.1.x

PR #544, opened 2021-03-10
578 reset session correctly

UlrichKu:
| Q             | A
| ------------- | ---
| Branch        | 7.0.x
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed issues  | chameleon-system/chameleon-system#578
| License       | MIT

### DIFF
--- a/src/ExtranetBundle/objects/db/TDataExtranetUser.class.php
+++ b/src/ExtranetBundle/objects/db/TDataExtranetUser.class.php
@@ -649,8 +649,11 @@ class TDataExtranetUser extends TDataExtranetUserAutoParent
                         $new[$key] = $all[$key];
                     }
                 }
-                $_SESSION = array();
-                $session->replace($new);
+
+                $session->invalidate(); // also clears $_SESSION
+                if (\count($new) > 0) {
+                    $session->replace($new);
+                }
             }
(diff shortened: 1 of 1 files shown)


src/ExtranetBundle/objects/db/TDataExtranetUser.class.php
```diff
@@ -649,8 +649,11 @@ public function Logout()
                         $new[$key] = $all[$key];
                     }
                 }
-                $_SESSION = array();
-                $session->replace($new);
+
+                $session->invalidate(); // also clears $_SESSION
+                if (\count($new) > 0) {
+                    $session->replace($new);
+                }
```
bestform, line 656: Can you explain, why this solves the problem? I am missing some context. A small business summary would be helpful. :)

UlrichKu: The "invalidate" is the important change here: replacing the session via SessionInterface::replace does not directly affect the $_SESSION variable and it is out of sync then. However other code in chameleon-base accesses that variable directly.

invalidate mitigates that problem.

PR #551, opened 2021-04-19
ref #699: Correct constructors after 652

UlrichKu:
| Q             | A
| ------------- | ---
| Branch        | 7.1.x
| Bug fix?      | yes
| New feature?  | no 
| BC breaks?    | no     
| Deprecations? | no 
| Tests pass?   | yes    
| Fixed issues  | chameleon-system/chameleon-system#699
| License       | MIT

Test note: Additionally changes for documents and media items.

### DIFF
--- a/src/CoreBundle/private/library/classes/dbobjects/TCMSDocumentManagerTreeNode.class.php
+++ b/src/CoreBundle/private/library/classes/dbobjects/TCMSDocumentManagerTreeNode.class.php
@@ -20,7 +20,7 @@ class TCMSDocumentManagerTreeNode extends TCMSTreeNode
 
     public function __construct($id = null)
     {
-        parent::__construct('cms_document_tree', $id);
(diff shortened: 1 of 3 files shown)


j6s:
After being confused by this for a while I can now see where this stems from:

https://github.com/chameleon-system/chameleon-base/pull/525/files#diff-4943f047c85f3d247746fac0318ced0557febe8ab0089e613a4157dbdcdd2317R21

This is the (my) change that introduced this behavior. Before my changes, the node classes 'skipped' the constructor of `TCMSTreeNode` and called the `TCMSRecord` one directly.

This fix seems good but it makes me wonder if I missed similar things in that PR. I'll double check it.

j6s:
I did not find any instances of the constructor chain changing in that PR except for the 3 that are being fixed here.

j6s (approved):

PR #557, opened 2021-05-27
ref #681: Add correct return type annotation to TDataExtranetUser::Register

j6s:
| Q             | A
| ------------- | ---
| Branch        | 7.0.x
| Bug fix?      | yes
| New feature?  | no 
| BC breaks?    | no     
| Deprecations? | no 
| License       | MIT

See also: #539 - that's where this issue was introduced.

Before this, TDataExtranetUser::Register was annotated as `@return bool`
but returned the results of `TDataExtranetUser::Save`, which returns the
id or `false`.

This commit changes the return types to `string|false` and fixes it's
incorrect usage while registering a user.


j6s:
This was already fixed in https://github.com/chameleon-system/chameleon-base/pull/559 - closing this one

PR #558, opened 2021-05-27
ref #539: Convert input data to SQL format when registering a user

j6s:
| Q             | A
| ------------- | ---
| Branch        | 7.0.x 
| Bug fix?      | yes
| New feature?  | no 
| BC breaks?    | no     
| Deprecations? | no 
| Fixed issues  | chameleon-system/chameleon-system#539   
| License       | MIT

During testing it emerged, that a users birthdate is not saved
correctly, as it is sent to the database in the format `DD.MM.YYYY`
instead of `YYYY-MM-DD`. Supplying the second argument to `LoadFromRow`
enables conversion of data formats, including dates.

### DIFF
--- a/src/ExtranetBundle/objects/WebModules/MTExtranetCore/MTExtranetCoreEndPoint.class.php
+++ b/src/ExtranetBundle/objects/WebModules/MTExtranetCore/MTExtranetCoreEndPoint.class.php
@@ -201,7 +201,7 @@ class MTExtranetCoreEndPoint extends TUserCustomModelBase
             $this->PrepareSubmittedData($aData);
             $extranetUserProvider->reset();
             $oUser = $extranetUserProvider->getActiveUser();
-            $oUser->LoadFromRow($aData);
+            $oUser->LoadFromRow($aData, true);
 
             $bDataValid = $this->ValidateUserLoginData();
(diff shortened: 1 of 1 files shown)


src/ExtranetBundle/objects/WebModules/MTExtranetCore/MTExtranetCoreEndPoint.class.php
```diff
@@ -201,7 +201,7 @@ public function Register($sSuccessURL = null, $sFailureURL = null)
             $this->PrepareSubmittedData($aData);
             $extranetUserProvider->reset();
             $oUser = $extranetUserProvider->getActiveUser();
-            $oUser->LoadFromRow($aData);
+            $oUser->LoadFromRow($aData, true);
```
rnuernberg, line 204: This may have unexpected consequences for existing installations. For example, if the input field is of type `date`, then the date will arrive in iso format (yyyy-mm-dd). I would expect the `LoadFromRow` to break this, if `true` is passed.

`PrepareSubmittedData` may be a better hook to change the date input format if required - or, alternatively the frontend input could be changed to type date.

rnuernberg (changes requested):

PR #559, opened 2021-05-28
ref #681: redirect after success register fix

kzorluoglu:
| Q             | A
| ------------- | ---
| Branch        | 7.1.x
| Bug fix?      | yes
| New feature?  | no 
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| License       | MIT

### DIFF
--- a/src/ExtranetBundle/objects/WebModules/MTExtranetCore/MTExtranetCoreEndPoint.class.php
+++ b/src/ExtranetBundle/objects/WebModules/MTExtranetCore/MTExtranetCoreEndPoint.class.php
@@ -221,8 +221,7 @@ class MTExtranetCoreEndPoint extends TUserCustomModelBase
             }
 
             if (true === $bDataValid) {
-                $registrationResult = $oUser->Register();
-                $bDataValid = false !== $registrationResult;
+                $bDataValid = $oUser->Register();
             }
 
             if (true === $bDataValid) {
--- a/src/ExtranetBundle/objects/WebModules/MTExtranetCore/MTExtranetCoreEndPoint.class.php
+++ b/src/ExtranetBundle/objects/WebModules/MTExtranetCore/MTExtranetCoreEndPoint.class.php
@@ -221,10 +221,11 @@ class MTExtranetCoreEndPoint extends TUserCustomModelBase
             }
 
             if (true === $bDataValid) {
-                $bDataValid = $oUser->Register();
+                $registrationResult = $oUser->Register();
+                $bDataValid = false !== $registrationResult;
             }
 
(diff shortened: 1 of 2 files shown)


src/ExtranetBundle/objects/WebModules/MTExtranetCore/MTExtranetCoreEndPoint.class.php
```diff
@@ -221,7 +221,8 @@ public function Register($sSuccessURL = null, $sFailureURL = null)
             }
 
             if (true === $bDataValid) {
-                $bDataValid = $oUser->Register();
+                $registrationResult = $oUser->Register();
```
UlrichKu, line 224: The return type of \TDataExtranetUser::Register() should also be fixed then.

kzorluoglu: I fixed first this method, but later restored it again. I thing, this change/fix can produce an error in several places.

UlrichKu: Why? It is annotated as bool but in 99% of the cases it returns string.

kzorluoglu: i talked with @fabianbarth-esono about the refactoring for the Method Register, but we both thinked, that should be in many place error generate. But, alright, i have the method refactored and committed.

UlrichKu: The initial code was ok.

I meant only the the return *type* annotation - not the implementation of the method.

kzorluoglu: Alright, Annotation updated and initial code is again here.

src/ExtranetBundle/objects/db/TDataExtranetUser.class.php
```diff
@@ -225,7 +225,7 @@ public function Save($bForceUserConfirm = false)
      * @param bool $bForceUserConfirmMail
      * @param bool $bAutoLoginAfterRegistration
      *
-     * @return bool
+     * @return bool|string
```
UlrichKu, line 228: + "the id or false for an error" (?)

kzorluoglu: done

PR #568, opened 2021-08-05
714/handle missing linked page object

pixeljunkie:
| Q             | A
| ------------- | ---
| Branch        | 7.0.x
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed issues  | chameleon-system/chameleon-system#714
| License       | MIT

Added a check for the linked page to prevent a crash.
Used the opportunity to replace the deprecated TGlobal::Translate calls with the new translation service and fix some typos. (separate commit)

### DIFF
--- a/src/CoreBundle/private/library/classes/TCMSFields/TCMSFieldPageTreeNode.class.php
+++ b/src/CoreBundle/private/library/classes/TCMSFields/TCMSFieldPageTreeNode.class.php
@@ -96,10 +96,15 @@ class TCMSFieldPageTreeNode extends TCMSFieldTreeNode
             return $this->getTranslationService()->trans('chameleon_system_core.field_page_tree_node.no_node_assigned');
         }
         // Retrieve portal through referenced linked page.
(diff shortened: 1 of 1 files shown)


src/CoreBundle/private/library/classes/TCMSFields/TCMSFieldPageTreeNode.class.php
```diff
@@ -352,4 +355,10 @@ private function getUrlUtilService(): UrlUtil
     {
         return ServiceLocator::get('chameleon_system_core.util.url');
     }
+
+    private function getTranslationService(): ChameleonTranslator
```
UlrichKu, line 359: That would be TranslatorInterface and "translator" as id.

However I would all translation changes out of this fix: There are extremely many of those calls and it makes probably sense to replace them in one go in the Core.

pixeljunkie: I removed the translation refactoring.

PR #569, opened 2021-08-06
ref #711: allow flash message consuming without global consumer

pixeljunkie:
| Q             | A
| ------------- | ---
| Branch        | 7.0.x
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed issues  | chameleon-system/chameleon-system#711
| License       | MIT

### DIFF
--- a/src/CoreBundle/private/library/classes/TCMSMessageManager/TCMSMessageManager.class.php
+++ b/src/CoreBundle/private/library/classes/TCMSMessageManager/TCMSMessageManager.class.php
@@ -224,11 +224,11 @@ class TCMSMessageManager
      *
      * @param string $sConsumerName
      * @param bool $bRemove
-     * @param bool $includeGlobalConsumer
+     * @param bool $includeGlobal
      *
      * @return TIterator
      */
-    public function ConsumeMessages($sConsumerName, $bRemove = true, bool $includeGlobalConsumer = true)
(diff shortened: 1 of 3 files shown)


src/CoreBundle/private/library/classes/TCMSMessageManager/TCMSMessageManager.class.php
```diff
@@ -223,11 +223,12 @@ protected function AddBackendMessage($sConsumerName, $sMessageCode, $aMessageCod
...
      *
      * @param string $sConsumerName
-     * @param bool   $bRemove
+     * @param bool $bRemove
+     * @param bool $includeGlobalConsumer
      *
      * @return TIterator
      */
-    public function ConsumeMessages($sConsumerName, $bRemove = true)
+    public function ConsumeMessages($sConsumerName, $bRemove = true, bool $includeGlobalConsumer = true)
```
UlrichKu, line 231: "include global consumer name messages" is what the new name means. (it's not the consumer that is included)

I would call it "includeGlobal" for short.

pixeljunkie: Done

PR #576, opened 2021-09-15
ref #726: Make methods static

j6s:
| Q             | A
| ------------- | ---
| Branch        | 7.1.x
| Bug fix?      | yes
| New feature?  | no 
| BC breaks?    | no     
| Deprecations? | no 
| Fixed issues  | chameleon-system/chameleon-system#726   
| License       | MIT

Calling non-static methods statically is not allowed anymore in PHP8. This changes `TCMSRecord::callBackUuid` to be a static method. This should be safe since the only invocation of it I could find comes from that same class in `TCMSRecord::getCellFormattingFunction`. While I was changing the methods I added return type annotations since they are fairly straight forward in this case.


UlrichKu:
Will be solved by https://github.com/chameleon-system/chameleon-system/issues/749

PR #578, opened 2021-09-15
ref #724: Check if value is string before unserializing

j6s:
| Q             | A
| ------------- | ---
| Branch        | 7.1.x
| Bug fix?      | yes
| New feature?  | no 
| BC breaks?    | no     
| Deprecations? | no 
| Fixed issues  | chameleon-system/chameleon-system#724   
| License       | MIT

In PHP8, attempting to unserialize a non-string (such as an array in this case, as the value has been previously unserialized) raises a TypeError as opposed to a warning in PHP7.4. This adds checks to the two instances I have found where this is an issue.

### DIFF
--- a/src/CoreBundle/private/rendering/objectviews/TCMSFields/TCMSFieldBlob/postload.view.php
+++ b/src/CoreBundle/private/rendering/objectviews/TCMSFields/TCMSFieldBlob/postload.view.php
@@ -10,7 +10,7 @@ if (isset($oDefinition->sqlData['is_translatable']) && '1' == $oDefinition->sqlD
 ?>
     if ($this->sqlData['<?= $sFieldDatabaseName; ?>'] === serialize(false)) {
         $this->sqlData['<?= $sFieldDatabaseName; ?>'] = false;  // special case - false was serialized
-    } else {
+    } elseif(is_string($this->sqlData['<?= $sFieldDatabaseName; ?>'])) {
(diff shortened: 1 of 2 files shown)


src/CoreBundle/private/rendering/objectviews/TCMSFields/TCMSFieldBlob/postload.view.php
```diff
@@ -10,7 +10,7 @@
 ?>
     if ($this->sqlData['<?= $sFieldDatabaseName; ?>'] === serialize(false)) {
         $this->sqlData['<?= $sFieldDatabaseName; ?>'] = false;  // special case - false was serialized
-    } else {
+    } elseif(is_string($this->sqlData['<?= $sFieldDatabaseName; ?>'])) {
```
UlrichKu, line 13: What else will it be?

What prior call is it that sometimes (?) decodes it before?

j6s: The Problem are the lines below it: `$this->sqlData[$sFieldDatabaseName]` is overridden by it's unserialized value. This means that after calling `PostLoadHook` once, the field in `sqlData` may be filled with an array instead of a string. When `PostLoadHook` is called again `unserialize` fails in PHP8 (Warning in PHP7).

I am unsure why this method is called twice, but adding a type-check to ensure that `unserialize` is always passed a string fixed it.

PR #580, opened 2021-10-04
ref #648: Do not use Tdb to avoid problems during (first) update

UlrichKu:
| Q             | A
| ------------- | ---
| Branch        | 7-0-x
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed issues  | chameleon-system/chameleon-system#648
| License       | MIT

### DIFF
--- a/src/CoreBundle/Service/CmsConfigDataAccess.php
+++ b/src/CoreBundle/Service/CmsConfigDataAccess.php
@@ -1,0 +1,58 @@
+<?php
+
+/*
+ * This file is part of the Chameleon System (https://www.chameleonsystem.com).
+ *
+ * (c) ESONO AG (https://www.esono.de)
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace ChameleonSystem\CoreBundle\Service;
+
+use Doctrine\DBAL\Connection;
+use Doctrine\DBAL\Exception;
+
+class CmsConfigDataAccess implements CmsConfigDataAccessInterface
+{
+    /**
+     * @var Connection
+     */
+    private $connection;
+
+    public function __construct(Connection $connection)
+    {
+        $this->connection = $connection;
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    public function getBackendTheme(): ?\TdbPkgCmsTheme
+    {
+        // TODO ? use a runtime-cache; this might be called several times...
+
+        try {
+            $themeId = $this->connection->fetchOne('SELECT `pkg_cms_theme_id` FROM `cms_config`');
+        } catch (Exception $exception) {
+            // The field might still be missing
+
+            // TODO ? log it still
+
+            return null;
+        }
+
+        if (false === $themeId) {
+            return null;
+        }
+
+        $theme = \TdbPkgCmsTheme::GetNewInstance();
+
+        if (false === $theme->Load($themeId)) {
+            return null;
+        }
+
+        return $theme;
+    }
+}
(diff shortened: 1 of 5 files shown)


src/CoreBundle/Service/CmsConfigDataAccess.php
```diff
@@ -0,0 +1,58 @@
...
+    {
+        $this->connection = $connection;
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    public function getBackendTheme(): ?\TdbPkgCmsTheme
+    {
+        // TODO ? use a runtime-cache; this might be called several times...
```
rnuernberg, line 34: please check if it is called multiple times. If so, add a runtime cache decorator.

src/CoreBundle/Service/CmsConfigDataAccess.php
```diff
@@ -0,0 +1,58 @@
...
+
+        try {
+            $themeId = $this->connection->fetchOne('SELECT `pkg_cms_theme_id` FROM `cms_config`');
+        } catch (Exception $exception) {
+            // The field might still be missing
+
+            // TODO ? log it still
```
rnuernberg, line 41: We should not swallow exceptions. Please add error log entry

src/CoreBundle/Service/CmsConfigDataAccess.php
```diff
@@ -0,0 +1,58 @@
...
+            return null;
+        }
+
+        if (false === $themeId) {
+            return null;
+        }
+
+        $theme = \TdbPkgCmsTheme::GetNewInstance();
+
+        if (false === $theme->Load($themeId)) {
```
rnuernberg, line 52: Instead of accessing the database twice, we can combine the request. Fetch the theme row from database - and then use the `LoadFromRow` method on the `TdbPkgCmsTheme` class.

rnuernberg (changes requested):

src/CoreBundle/Service/CmsConfigDataAccess.php
UlrichKu: ok

src/CoreBundle/Service/CmsConfigDataAccess.php
UlrichKu: With a subquery?

select * from theme where id = (select id from config)?

src/CoreBundle/Service/CmsConfigDataAccess.php
UlrichKu: Well it is. Normally 3 times.

However I don't get a cache decorator working (is always non-shared).

src/CoreBundle/Service/CmsConfigDataAccess.php
rnuernberg: This should do the trick.
```sql
SELECT `pkg_cms_theme`.*
   FROM `pkg_cms_theme` INNER JOIN `cms_config` ON `pkg_cms_theme`.`id` = `cms_config`.`pkg_cms_theme_id`
```

src/CoreBundle/Service/CmsConfigDataAccess.php
UlrichKu: It seems to be actually independent requests. So a runtime cache is not necessary /wouldn't work.

src/CoreBundle/Service/CmsConfigDataAccess.php
UlrichKu: ok

PR #597, opened 2022-01-27
Convert possible integer type to be a string for equation…

michaelcmichaelis:
… test, avoiding a conversion of a string id to an integer

| Q             | A
| ------------- | ---
| Branch        | 7.0.x
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Fixed issues  | chameleon-system/chameleon-system#758
| License       | MIT

Fixes Twig ver. 2 comparison behaviour of PHP when comparing numbers and strings.


UlrichKu:
Does not appear in Twig 3.0 (and thus Chameleon 7.1) and this workaround should be removed there after merging up.

PR #605, opened 2022-03-25
ref #764: Extract theme handling to have a better control about that

UlrichKu:
| Q             | A
| ------------- | ---
| Branch        | 7.0.x
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no - hopefully not: everything old should still behave the same (doesn't use new override)
| Deprecations? | no
| Tests pass?   | yes
| Fixed issues  | chameleon-system/chameleon-system#764
| License       | MIT


UlrichKu:
This should be a hotfix.

PR #606, opened 2022-03-28
ref #764: Extract theme handling to have a better control about that

UlrichKu:
| Q             | A
| ------------- | ---
| Branch        | 7.0.x
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no - hopefully not: everything old should still behave the same (doesn't use new override)
| Deprecations? | no
| Tests pass?   | yes
| Fixed issues  | chameleon-system/chameleon-system#764
| License       | MIT

### DIFF
--- a/src/CoreBundle/Bridge/Chameleon/Migration/Script/update-1613466542.inc.php
+++ b/src/CoreBundle/Bridge/Chameleon/Migration/Script/update-1613466542.inc.php
@@ -1,0 +1,16 @@
+<h1>Build #1613466542</h1>
+<h2>Date: 2021-03-22</h2>
+<div class="changelog">
+    - #764: Fix backend theme snippet chain content (it now should have ALL paths)
+</div>
+<?php
+
+$backendThemeId = '5f047d9b-0c20-0bfb-2dce-f8193653965c';
+
+// Make sure to have the same starting point
+TCMSLogChange::removeFromSnippetChain('@ChameleonSystemCoreBundle/Resources/views', [$backendThemeId]);
+
+// Add correctly ordered "default" elements (now removed from \TPkgViewRendererSnippetDirectory::getBasePaths())
+TCMSLogChange::addToSnippetChain('@ChameleonSystemCoreBundle/Resources/views', '^', [$backendThemeId]);
+TCMSLogChange::addToSnippetChain('../extensions', '@ChameleonSystemCoreBundle/Resources/views', [$backendThemeId]);
+
--- a/src/CoreBundle/Bridge/Chameleon/Migration/Script/update-1613466542.inc.php
+++ b/src/CoreBundle/Bridge/Chameleon/Migration/Script/update-1613466542.inc.php
@@ -5,12 +5,17 @@
 </div>
 <?php
 
-$backendThemeId = '5f047d9b-0c20-0bfb-2dce-f8193653965c';
+$connection = TCMSLogChange::getDatabaseConnection();
 
-// Make sure to have the same starting point
-TCMSLogChange::removeFromSnippetChain('@ChameleonSystemCoreBundle/Resources/views', [$backendThemeId]);
+$backendThemeId = $connection->fetchOne('SELECT `pkg_cms_theme_id` FROM `cms_config`');
 
-// Add correctly ordered "default" elements (now removed from \TPkgViewRendererSnippetDirectory::getBasePaths())
-TCMSLogChange::addToSnippetChain('@ChameleonSystemCoreBundle/Resources/views', '^', [$backendThemeId]);
-TCMSLogChange::addToSnippetChain('../extensions', '@ChameleonSystemCoreBundle/Resources/views', [$backendThemeId]);
+if (false !== $backendThemeId) {
+    // Make sure to have the same starting point
+    TCMSLogChange::removeFromSnippetChain('@ChameleonSystemCoreBundle/Resources/views', [$backendThemeId]);
+    TCMSLogChange::removeFromSnippetChain('../extensions', [$backendThemeId]);
+
+    // Add correctly ordered "default" elements (now removed from \TPkgViewRendererSnippetDirectory::getBasePaths())
+    TCMSLogChange::addToSnippetChain('@ChameleonSystemCoreBundle/Resources/views', '^', [$backendThemeId]);
+    TCMSLogChange::addToSnippetChain('../extensions', '@ChameleonSystemCoreBundle/Resources/views', [$backendThemeId]);
+}
 
--- a/src/ViewRenderer/Interfaces/ThemeServiceInterface.php
+++ b/src/ViewRenderer/Interfaces/ThemeServiceInterface.php
@@ -20,6 +20,4 @@ interface ThemeServiceInterface
      * @return \TdbPkgCmsTheme|null
      */
     public function getTheme(?\TdbCmsPortal $portal): ?\TdbPkgCmsTheme;
-
-    public function setOverrideTheme(?\TdbPkgCmsTheme $themeOverride): void;
 }
--- a/src/ViewRenderer/Interfaces/ThemeServiceInterface.php
+++ b/src/ViewRenderer/Interfaces/ThemeServiceInterface.php
@@ -1,0 +1,25 @@
+<?php
+
+/*
+ * This file is part of the Chameleon System (https://www.chameleonsystem.com).
+ *
+ * (c) ESONO AG (https://www.esono.de)
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace ChameleonSystem\ViewRenderer\Interfaces;
+
+interface ThemeServiceInterface
+{
+    /**
+     * Gets the current theme.
+     * This correctly considers the portal or backend but can also be overwritten by setting the theme beforehand - which is then always returned.
+     *
+     * @return \TdbPkgCmsTheme|null
+     */
+    public function getTheme(?\TdbCmsPortal $portal): ?\TdbPkgCmsTheme;
+
+    public function setOverrideTheme(?\TdbPkgCmsTheme $themeOverride): void;
+}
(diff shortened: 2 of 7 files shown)


src/CoreBundle/Bridge/Chameleon/Migration/Script/update-1613466542.inc.php
```diff
@@ -0,0 +1,16 @@
...
+<?php
+
+$backendThemeId = '5f047d9b-0c20-0bfb-2dce-f8193653965c';
+
+// Make sure to have the same starting point
+TCMSLogChange::removeFromSnippetChain('@ChameleonSystemCoreBundle/Resources/views', [$backendThemeId]);
+
+// Add correctly ordered "default" elements (now removed from \TPkgViewRendererSnippetDirectory::getBasePaths())
+TCMSLogChange::addToSnippetChain('@ChameleonSystemCoreBundle/Resources/views', '^', [$backendThemeId]);
+TCMSLogChange::addToSnippetChain('../extensions', '@ChameleonSystemCoreBundle/Resources/views', [$backendThemeId]);
```
rnuernberg, line 15: ca we be sure, that the theme does not already contain `../extensions`? I think it would be safer to run `removeFromSnippetChain` on it first.

src/ViewRenderer/Interfaces/ThemeServiceInterface.php
```diff
@@ -0,0 +1,25 @@
...
+{
+    /**
+     * Gets the current theme.
+     * This correctly considers the portal or backend but can also be overwritten by setting the theme beforehand - which is then always returned.
+     *
+     * @return \TdbPkgCmsTheme|null
+     */
+    public function getTheme(?\TdbCmsPortal $portal): ?\TdbPkgCmsTheme;
+
+    public function setOverrideTheme(?\TdbPkgCmsTheme $themeOverride): void;
```
rnuernberg, line 24: I would remove this method from the interface (but keep it in the concrete implementation). This is a special case we need only for the console command that is used to generate all themes. We do not want to make that option available to other code.

src/ViewRendererBundle/objects/helper/TPkgViewRendererLessCompiler.php
```diff
@@ -139,15 +131,18 @@ public function getCompiledCssFilenamePattern()
...
-
         $theme = null;
-        if (is_object($portal)) {
+        if (null !== $portal) {
             $theme = $portal->GetFieldPkgCmsTheme();
         }
 
-        if (!$theme || empty($theme->fieldLessFile)) {
+        $resourceCollector = new TPkgViewRendererSnippetResourceCollector();
+        $this->themeService->setOverrideTheme($theme);
```
rnuernberg, line 140: why are we replacing the theme here? I think the only place we need to do that is in the console command that compiles less for all portals and the backend.

src/ViewRendererBundle/objects/helper/TPkgViewRendererSnippetDirectory.class.php
```diff
@@ -28,33 +28,22 @@ class TPkgViewRendererSnippetDirectory implements TPkgViewRendererSnippetDirecto
...
-    private $kernel;
-    /**
-     * @var CmsConfigDataAccessInterface
-     */
-    private $cmsConfigDataAccess;
+
+    private RequestInfoServiceInterface $requestInfoService;
+    private KernelInterface $kernel;
+    private ThemeServiceInterface $themeService;
+    private PortalDomainServiceInterface $portalDomainService;
```
rnuernberg, line 35: Typed properties were introduced in php 7.4. Chameleon 7.0 has a minimum requirement of php 7.1. If you intend to use this fix in 7.0 then you need to undo this change.

rnuernberg (changes requested):

src/CoreBundle/Bridge/Chameleon/Migration/Script/update-1613466542.inc.php
UlrichKu: ok

src/CoreBundle/Bridge/Chameleon/Migration/Script/update-1613466542.inc.php
```diff
@@ -0,0 +1,16 @@
+<h1>Build #1613466542</h1>
+<h2>Date: 2021-03-22</h2>
+<div class="changelog">
+    - #764: Fix backend theme snippet chain content (it now should have ALL paths)
+</div>
+<?php
+
+$backendThemeId = '5f047d9b-0c20-0bfb-2dce-f8193653965c';
```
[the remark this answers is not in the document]
UlrichKu: ok

src/ViewRendererBundle/objects/helper/TPkgViewRendererLessCompiler.php
UlrichKu: ok

src/ViewRendererBundle/objects/helper/TPkgViewRendererSnippetDirectory.class.php
UlrichKu: ok

src/ViewRendererBundle/Command/CompileLessCommand.php
```diff
@@ -64,7 +70,10 @@ protected function execute(InputInterface $input, OutputInterface $output)
         $portalList = TdbCmsPortalList::GetList();
         $minifyCss = true === $input->getOption('minify-css');
         while ($portal = $portalList->Next()) {
+            $this->themeService->setOverrideTheme($portal->GetFieldPkgCmsTheme());
             $css = $this->lessCompiler->getGeneratedCssForPortal($portal, $minifyCss);
+            $this->themeService->setOverrideTheme(null);
+
```
rnuernberg, line 76: The command must also compile less for the backend - as an additional step

src/ViewRendererBundle/Service/ThemeService.php
```diff
@@ -0,0 +1,61 @@
...
+    /**
+     * {@inheritDoc}
+     */
+    public function getTheme(?\TdbCmsPortal $portal): ?\TdbPkgCmsTheme
+    {
+        if (null !== $this->themeOverride) {
+            return $this->themeOverride;
+        }
+
+        if (null === $portal) {
```
rnuernberg, line 41: This should match the old logic.

1. if backend mode, return backend theme
2. if portal, return portal theme
3. otherwise, get active portal and return corresponding theme.
4. still nothing, return null

src/ViewRendererBundle/objects/helper/TPkgViewRendererLessCompiler.php
```diff
@@ -139,15 +129,15 @@ public function getCompiledCssFilenamePattern()
      */
     protected function getLessFilesFromSnippetsAndTheme($portal)
     {
-        $resourceCollector = new TPkgViewRendererSnippetResourceCollector();
-        $resources = $resourceCollector->getLessResources($portal, CMS_SNIPPET_PATH);
-
         $theme = null;
-        if (is_object($portal)) {
+        if (null !== $portal) {
```
rnuernberg, line 133: here we want to get the theme from the theme service - not from the portal. Otherwise we have no way of returning the less files for the backend theme.

rnuernberg (changes requested):

src/ViewRendererBundle/Command/CompileLessCommand.php
UlrichKu: There is no less in the backend.

(no loading, no location)

src/ViewRendererBundle/objects/helper/TPkgViewRendererLessCompiler.php
UlrichKu: Hm, also this location expects everything to be "frontend". So when no explicit additional less file is configured for this theme "/assets/less/chameleon.less" is loaded. And that is normally used for frontend less.

src/ViewRendererBundle/Service/ThemeService.php
UlrichKu: (the old logic was: use portal first, if none is present load the active one)

src/ViewRendererBundle/objects/helper/TPkgViewRendererLessCompiler.php
rnuernberg: The service that provides the theme will work here, will it not? If so, then please do use it.

src/ViewRendererBundle/Command/CompileLessCommand.php
rnuernberg: A theme supports less, does it not? So we should compile it here (if the theme has less). Even if it is not currently loaded. In fact, we should probably load it too (if that is easily done)

src/ViewRendererBundle/Service/ThemeService.php
rnuernberg: it looks unchanged. Please move 
```php
        if (true === $this->requestInfoService->isBackendMode()) {
            return $this->cmsConfigDataAccess->getBackendTheme();
        }
```

to line 44

rnuernberg (changes requested):
There are open comments that need to be addressed

src/ViewRendererBundle/Service/ThemeService.php
UlrichKu: Done

src/ViewRendererBundle/objects/helper/TPkgViewRendererLessCompiler.php
UlrichKu: Done

src/ViewRendererBundle/Command/CompileLessCommand.php
UlrichKu: At the moment a portal supports less.

Everything from TPkgViewRendererLessCompiler::getGeneratedCssForPortal() onwards needs a portal.

This would be in any case a new feature / issue.

PR #609, opened 2022-03-30
756 log error info 3

UlrichKu:
| Q             | A
| ------------- | ---
| Branch        | 7.1.x
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed issues  | chameleon-system/chameleon-system#756
| License       | MIT


src/MediaManagerBundle/Bridge/Chameleon/BackendModule/MediaManagerBackendModule.php
```diff
@@ -630,7 +629,7 @@ protected function renderDetail()
         try {
             $contentHtml = $this->responseVariableReplacer->replaceVariables($contentHtml);
         } catch (TokenInjectionFailedException $e) {
-            $this->returnGeneralErrorMessageForAjax();
+            $this->returnGeneralErrorMessageForAjax('MediaManagerBackendModule: Couldn\'t replace variables');
```
bestform, line 632: We could also provide the exception as the new interface allows an optional exception.

src/MediaManagerBundle/Bridge/Chameleon/BackendModule/MediaManagerBackendModule.php
```diff
@@ -495,8 +499,10 @@ protected function renderList()
         $this->returnAsAjaxResponse($listReturn);
     }
 
-    private function returnGeneralErrorMessageForAjax()
+    private function returnGeneralErrorMessageForAjax(?string $logMessage = null, ?\Throwable $exception = null)
```
bestform, line 502: The new interface in combination with the old name is somewhat missleading. `returnGeneralErrorMessageForAjax` in combination with an optional error message seems like this message would be returned via ajax - which is not the case. So we should either rename this method to indicate what the message is actually used for. Or (and I would prefer this) let the method just do what it did before and use the logger before this method is called. This makes it clear when reading code that there will be a message logged and afterwards there will be a general error returned to the user.

bestform (changes requested):

src/MediaManagerBundle/Bridge/Chameleon/BackendModule/MediaManagerBackendModule.php
UlrichKu: ok

src/MediaManagerBundle/Bridge/Chameleon/BackendModule/MediaManagerBackendModule.php
UlrichKu: changed the method name

(it is called more than 20 times...)

bestform:
There are conflicts now. I will happily merge when they are resolved.

PR #610, opened 2022-03-31
ref #653: Always only have \n as line ending (after save)

UlrichKu:
| Q             | A
| ------------- | ---
| Branch        | 7.1.x
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | yes probably no: anything that relies on \r\n being present
| Deprecations? | no
| Tests pass?   | yes
| Fixed issues  | chameleon-system/chameleon-system#653
| License       | MIT

### DIFF
--- a/src/CoreBundle/private/library/classes/TCMSFields/TCMSFieldWYSIWYG.class.php
+++ b/src/CoreBundle/private/library/classes/TCMSFields/TCMSFieldWYSIWYG.class.php
@@ -711,6 +711,20 @@ class TCMSFieldWYSIWYG extends TCMSFieldText
     }
 
     /**
+     * {@inheritDoc}
+     */
+    public function ConvertPostDataToSQL()
+    {
+        $data = parent::ConvertPostDataToSQL();
+
+        // make sure line endings are always consistent (simple \n)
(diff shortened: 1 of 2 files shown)


src/CoreBundle/private/library/classes/TCMSFields/TCMSFieldWYSIWYG.class.php
```diff
@@ -710,6 +710,20 @@ protected function getEditorWidth()
         }
     }
 
+    /**
+     * {@inheritDoc}
+     */
+    public function ConvertPostDataToSQL()
```
rnuernberg, line 716: The issue mentions the problem for text fields, not just wysiwyg. So I would suggest moving the logic into the parent `TCMSFieldText`

rnuernberg (changes requested):

src/CoreBundle/private/library/classes/TCMSFields/TCMSFieldWYSIWYG.class.php
UlrichKu: Done

PR #617, opened 2022-04-14
ref #751: Allow easy disabling of jquery

UlrichKu:
| Q             | A
| ------------- | ---
| Branch        | 7.1.x
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed issues  | chameleon-system/chameleon-system#751
| License       | MIT

### DIFF
--- a/src/CoreBundle/EventListener/AddJqueryIncludeListener.php
+++ b/src/CoreBundle/EventListener/AddJqueryIncludeListener.php
@@ -21,18 +21,22 @@ class AddJqueryIncludeListener
      */
     public function onGlobalHtmlHeaderInclude(HtmlIncludeEventInterface $event)
     {
+        $jqueryInclude = '';
         if (true === TGlobal::IsCMSMode()) {
-            $sInclude = '<script src="'.TGlobal::GetStaticURL('/chameleon/blackbox/javascript/jquery/jquery-3.3.1.min.js').'" type="text/javascript"></script><!--#GLOBALRESOURCECOLLECTION#-->';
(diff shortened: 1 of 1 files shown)


src/CoreBundle/EventListener/AddJqueryIncludeListener.php
```diff
@@ -21,18 +21,22 @@ class AddJqueryIncludeListener
...
+                $jqueryInclude .= "\n<script>window.jQuery || document.write('".str_replace('/', '\\/', $sJQueryLocal)."')</script>";
+            } else if (CHAMELEON_URL_JQUERY !== false) {
+                $jqueryInclude = '<script src="'.TGlobal::GetStaticURL(CHAMELEON_URL_JQUERY).'" type="text/javascript"></script><!--#GLOBALRESOURCECOLLECTION#-->';
             }
         }
-        $event->addData(array($sInclude));
+
+        if ('' !== $jqueryInclude) {
+            $event->addData([$jqueryInclude]);
+        }
```
bestform, line 40: Instead of checking for an emtpy string, couldn't we just add the data in every branch of the if statements above? This would make it simpler to read in my opinion.

bestform (changes requested):

src/CoreBundle/EventListener/AddJqueryIncludeListener.php
UlrichKu: Done

bestform (approved):

PR #620, opened 2022-04-22
ref #771: Fix static code analysis issues found in chameleon-shop

j6s:
| Q             | A
| ------------- | ---
| Branch        | 7.0.x
| Bug fix?      | no
| New feature?  | no 
| BC breaks?    | no     
| Deprecations? | no 
| Tests pass?   | yes    
| Fixed issues  | chameleon-system/chameleon-system#771
| License       | MIT

Fixes a couple of typing / code analysis issues that were uncovered while working on https://github.com/chameleon-system/chameleon-shop/pull/120 .

There are no functional changes in this PR - It mainly consists of docblock changes. I did change actual code in 3 places - but only in Tdb-generation logic in order for the Tdb's to have correct docblocks. I will add annotations to all of those places.

### DIFF
--- a/src/CoreBundle/private/library/classes/TCMSFields/TCMSFieldBlob.class.php
+++ b/src/CoreBundle/private/library/classes/TCMSFields/TCMSFieldBlob.class.php
@@ -68,4 +68,12 @@ class TCMSFieldBlob extends TCMSFieldText
 
         return '<pre style="white-space:pre">'.$html.'</pre>';
     }
+
+    protected function GetFieldWriterData()
+    {
+        $data = parent::GetFieldWriterData();
+        $data['sFieldType'] = 'mixed';
+
+        return $data;
+    }
 }
--- a/src/CoreBundle/private/library/classes/TCMSFields/TCMSFieldBoolean.class.php
+++ b/src/CoreBundle/private/library/classes/TCMSFields/TCMSFieldBoolean.class.php
@@ -83,7 +83,7 @@ class TCMSFieldBoolean extends TCMSFieldOption
     {
         $aData = parent::GetFieldWriterData();
 
-        $aData['sFieldType'] = 'boolean';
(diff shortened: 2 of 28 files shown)


src/CoreBundle/private/library/classes/TCMSFields/TCMSFieldBlob.class.php
```diff
@@ -68,4 +68,12 @@ public function GetReadOnly()
 
         return '<pre style="white-space:pre">'.$html.'</pre>';
     }
+
+    protected function GetFieldWriterData()
```
j6s, line 72: A blob field is deserialized when loading the data - it could contain any type of data - thus overwriting it's docblock type with `mixed` instead of `string` to reflect that

src/CoreBundle/private/library/classes/TCMSFields/TCMSFieldBoolean.class.php
```diff
@@ -83,7 +83,7 @@ protected function GetFieldWriterData()
     {
         $aData = parent::GetFieldWriterData();
 
-        $aData['sFieldType'] = 'boolean';
+        $aData['sFieldType'] = 'bool';
```
j6s, line 86: Using the short version of `boolean`

src/CoreBundle/private/library/classes/TCMSFields/TCMSFieldNumber.class.php
```diff
@@ -102,4 +102,11 @@ public function _GetFieldWidth()
 
         return $iFieldMaxLength;
     }
+
+    protected function GetFieldWriterData() {
+        $data = parent::GetFieldWriterData();
+        $data['sFieldType'] = 'int';
```
j6s, line 108: Number fields contain integers not strings (parent sets type `string`)

PR #629, opened 2022-05-05
ref #772: Remove unwanted (and unusable) code

UlrichKu:
| Q             | A
| ------------- | ---
| Branch        | master
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed issues  | chameleon-system/chameleon-system#772
| License       | MIT

### DIFF
--- a/src/PkgCoreBundle/objects/TPkgCmsCoreSendToHost.class.php
+++ b/src/PkgCoreBundle/objects/TPkgCmsCoreSendToHost.class.php
@@ -123,23 +123,6 @@ class TPkgCmsCoreSendToHost
     {
         if (null === $this->lastResponseHeaderVariables) {
             $this->lastResponseHeaderVariables = array();
-            $rHeader = TPkgCmsStringUtilities::getStringStream($this->getLastResponseHeader());
-            $this->lastResponseCodeRaw = fgets($rHeader);
-
-            $aTmp = explode("\n", $this->getLastResponseHeader());
-            foreach ($aTmp as $sHeaderLine) {
-                $iSplit = strpos($sHeaderLine, ':');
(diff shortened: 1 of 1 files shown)


src/PkgCoreBundle/objects/TPkgCmsCoreSendToHost.class.php
```diff
@@ -123,23 +123,6 @@ public function getLastResponseHeaderVariable($sVarName)
...
-                if (isset($this->lastResponseHeaderVariables[$sHeaderVarName])) {
-                    if (!is_array($this->lastResponseHeaderVariables[$sHeaderVarName])) {
-                        $this->lastResponseHeaderVariables[$sHeaderVarName] = array($this->lastResponseHeaderVariables[$sHeaderVarName]);
-                    }
-                    $this->lastResponseHeaderVariables[$sHeaderVarName][] = $sHeaderVarValue;
-                } else {
-                    $this->lastResponseHeaderVariables[$sHeaderVarName] = $sHeaderVarValue;
-                }
-            }
         }
```
bestform, line 126: we should not set an empty array if it wasn't set before. This private attribute has the characteristics of a runtime cache. So setting it to an empty array may signal to other methods that the cache was successfully built, which it wasn't in that case. We could just return null instead and leave the attribute unset.

bestform (changes requested):

src/PkgCoreBundle/objects/TPkgCmsCoreSendToHost.class.php
UlrichKu: There is no "error" in this case. This is plain initialization.

bestform (approved):

PR #632, opened 2022-05-05
ref #721: Get interfaces in line (allow null everywhere)

UlrichKu:
| Q             | A
| ------------- | ---
| Branch        | master
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed issues  | chameleon-system/chameleon-system#721
| License       | MIT


bestform:
This is quite the change and a massive BC break. Before this change every call to get the new portal or portal domain could safely assume that the value could not be null. Now this is not the case anymore. I would assume this will break a lot of existing code. Or am I wrong?

bestform:
Am I missing something here? The event never allowed the *new* portal or domain to be null. Now it does. So code working with the event will break if we now allow null values.

UlrichKu:
> So code working with the event will break if we now allow null values.

The code also breaks now ("one level up").

What would be your suggestion for a beneficial solution?

PR #635, opened 2022-05-11
779 switch link buttons

UlrichKu:
| Q             | A
| ------------- | ---
| Branch        | 7.1.x
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed issues  | chameleon-system/chameleon-system#779
| License       | MIT

### DIFF
--- a/src/CoreBundle/Resources/public/javascript/tableEditor.js
+++ b/src/CoreBundle/Resources/public/javascript/tableEditor.js
@@ -944,7 +944,7 @@ CHAMELEON.CORE.MTTableEditor.initSwitchToEntryButtons = function () {
             return;
         }
 
-        // NOTE a change listener of the selectElement (= "the other way around") does not work as that is masked by select2 pseudo element (TODO remove select2)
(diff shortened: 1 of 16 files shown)


src/CoreBundle/Resources/public/javascript/tableEditor.js
```diff
@@ -937,6 +937,30 @@ CHAMELEON.CORE.MTTableEditor.idButtonCopyToClipboard = function () {
...
+                return;
+            }
+
+            const href = new URL(linkButton.getAttribute("href"), document.location.origin);
+            href.searchParams.set('id', selectElement.value);
+
+            linkButton.setAttribute("href", href.toString());
+        });
+    });
+};
```
bestform, line 962: Why are we modifying the links after the fact. Can't this logic live directly in the twig template that introduced those buttons?

UlrichKu: If the select option changes so must the target of the link (button).

PR #638, opened 2022-05-19
766 geocoding

j6s:
| Q             | A
| ------------- | ---
| Branch        | 7.0.x
| Bug fix?      | yes
| New feature?  | no 
| BC breaks?    | no     
| Deprecations? | no 
| Tests pass?   | yes    
| Fixed issues  | chameleon-system/chameleon-system#766   
| License       | MIT

This replaces the usage of google maps geocoder with nominatim since google requires a paid subscription.
Note: This PR adds the `symfony/http-client` **component** which is only available with 4.4. However, it does not
have any external dependencies this is fine. In fact, before this change `symfony/http-client` was already installed
through a dependency of `chameleon-base` in version 5.4.


src/CoreBundle/Resources/public/fields/TCMSFieldGMapCoordinate/TCMSFieldGMapCoordinate.js
```diff
@@ -123,32 +155,75 @@ CHAMELEON.CORE.TCMSFieldGMapCoordinate =
...
+     * @param {string} query
+     * @return {Promise<{ lat: float, lng: float }|null, string>}
+     */
+    geocode: function(query) {
+        /**
+         * @see https://nominatim.org/release-docs/develop/api/Search/
+         * Using geojson output instead of default JSON in order to be compatible with more tools
+         * (e.g. selfhosted photon geocoder) by using a standardized data format.
+         */
+        var url = new URL('https://nominatim.openstreetmap.org/search');
```
pixeljunkie, line 204: If selfhosting is an option, we should make the API URL configurable.

src/CoreBundle/private/library/classes/TGoogleMap/TGoogleMapEndPoint.class.php
```diff
@@ -218,25 +221,52 @@ public function hookMenueLinks($b)
...
+     * @return array{latitude: float, longitude: float}|false - contains latitude and longitude coordinate
      */
     public function locatePlace($sPlace = '')
     {
-        $aCoordinates = false;
-        $sURL = 'http://maps.googleapis.com/maps/api/geocode/json?address='.urlencode($sPlace);
-
-        $sResponse = file_get_contents($sURL);
-        if (empty($sResponse)) {
+        $url = sprintf('https://nominatim.openstreetmap.org/search?%s', http_build_query([
```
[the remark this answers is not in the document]
j6s: Fair enough. I made the endpoint configurable and changed the frontend code from calling the endpoint directly to calling a controller, which then does geocoding from PHP. This was done in order to have consistent geocoding behaviour in case we want to replace geocoding behaviour with something else

src/CoreBundle/DependencyInjection/ChameleonSystemCoreExtension.php
```diff
@@ -211,6 +212,16 @@ private function addGoogleApiConfig(array $googleApiConfig, ContainerBuilder $co
...
     }
 
+    private function addGeocoderConfig(array $geocoderConfig, ContainerBuilder $container): void
+    {
+        $definition = $container->getDefinition('chameleon_system_core.geocoding.geocoder');
+        $definition->setArgument(0, $geocoderConfig['geo_json_endpoint']);
+
+        $container->setParameter('chameleon_system_core.geocoding.attribution.show', $geocoderConfig['attribution']['show']);
+        $container->setParameter('chameleon_system_core.geocoding.attribution.name', 'test' . $geocoderConfig['attribution']['name']);
+        $container->setParameter('chameleon_system_core.geocoding.attribution.url', 'test' . $geocoderConfig['attribution']['url']);
```
bestform, line 222: I guess the test was there for debugging purposes? :)

bestform (changes requested):

src/CoreBundle/DependencyInjection/ChameleonSystemCoreExtension.php
j6s: `test` string has been removed

PR #645, opened 2022-06-21
ref #784 add none session extranet user service, to allow setting act…

pSpeck:
…ive user without active session.

| Q             | A
| ------------- | ---
| Branch        | 7.1.x
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed issues  | chameleon-system/chameleon-system#784
| License       | MIT

### DIFF
--- a/src/ExtranetBundle/Service/ExtranetUserStaticProvider.php
+++ b/src/ExtranetBundle/Service/ExtranetUserStaticProvider.php
@@ -1,0 +1,59 @@
+<?php
+
+/*
+ * This file is part of the Chameleon System (https://www.chameleonsystem.com).
+ *
+ * (c) ESONO AG (https://www.esono.de)
+ *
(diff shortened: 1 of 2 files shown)


src/ExtranetBundle/Service/ExtranetUserStaticProvider.php
```diff
@@ -0,0 +1,59 @@
...
+
+    /**
+     * {@inheritdoc}
+     */
+    public function reset()
+    {
+        if (null !== $this->staticUser) {
+            $this->staticUser = null;
+
+            return;
```
UlrichKu, line 54: That seems problematic: To actually reset the current user you have to call reset() twice. First for the decorator and only then can it reach the real UserProvider.

PR #649, opened 2022-07-04
only check for portal if the request is a page request

bestform:
fixes #787

| Q             | A
| ------------- | ---
| Branch        | 7.1.x
| Bug fix?      | yes
| New feature?  | no 
| BC breaks?    | no     
| Deprecations? | no 
| Tests pass?   | yes    
| Fixed issues  | chameleon-system/chameleon-system#787
| License       | MIT

Check the path of the request to decide if it is a resource request. If so do not check for the portal of the user.

Resources cannot be limited to a specific portal so this is no security regression.

### DIFF
--- a/src/ExtranetBundle/objects/db/TDataExtranetUser.class.php
+++ b/src/ExtranetBundle/objects/db/TDataExtranetUser.class.php
@@ -936,6 +936,20 @@ class TDataExtranetUser extends TDataExtranetUserAutoParent
         return array('table', 'oBillingAddress', 'oShippingAddress', 'aSessionData', 'sqlData', 'pageAccessCache');
     }
 
+    // isRequestForResource will try to identify the request based on its path.
+    // This method is quite specific and only used when deciding if it should check the active portal of the user
+    // on session validation. See https://github.com/chameleon-system/chameleon-system/issues/787 for details.
+    private function isRequestForResource(): bool {
+        $pathInfo = $this->getCurrentRequest()->getPathInfo();
+
+        $isResourceRequest =
+            str_starts_with($pathInfo, '/_wdt') ||
+            str_ends_with($pathInfo, '.css') ||
+            str_ends_with($pathInfo, '.jpg');
+
+        return $isResourceRequest;
+    }
+
     /**
      * validates the user session data.
      *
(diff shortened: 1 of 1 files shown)


src/ExtranetBundle/objects/db/TDataExtranetUser.class.php
```diff
@@ -936,6 +936,20 @@ public function __sleep()
...
+    // isRequestForResource will try to identify the request based on its path.
+    // This method is quite specific and only used when deciding if it should check the active portal of the user
+    // on session validation. See https://github.com/chameleon-system/chameleon-system/issues/787 for details.
+    private function isRequestForResource(): bool {
+        $pathInfo = $this->getCurrentRequest()->getPathInfo();
+
+        $isResourceRequest =
+            str_starts_with($pathInfo, '/_wdt') ||
+            str_ends_with($pathInfo, '.css') ||
+            str_ends_with($pathInfo, '.jpg');
```
pixeljunkie, line 948: I would suggest to set paths instead: /chameleon/mediapool, /chameleon/outbox, /_wdt.
Otherwise you would need a full list of file extensions. Like .png, .webp, .pdf and other.

src/ExtranetBundle/objects/db/TDataExtranetUser.class.php
```diff
@@ -960,7 +974,7 @@ public function ValidateSessionData()
             if ($aUserData) {
                 $iCurrentTime = time();
                 $dataValid = (($iCurrentTime - $aUserData['login_timestamp']) < $oExtranetConfig->fieldSessionlife);
-                if ($dataValid && CHAMELEON_EXTRANET_USER_IS_PORTAL_DEPENDANT && !TGlobal::IsCMSMode()) {
+                if ($dataValid && CHAMELEON_EXTRANET_USER_IS_PORTAL_DEPENDANT && !TGlobal::IsCMSMode() && !$this->isRequestForResource()) {
```
UlrichKu, line 977: Checking for example for the "master request" doesn't help here?

src/ExtranetBundle/objects/db/TDataExtranetUser.class.php
UlrichKu: There might be other (future, custom) resources with other suffixes or paths (*.js for example)?

UlrichKu (changes requested):

src/ExtranetBundle/objects/db/TDataExtranetUser.class.php
bestform: The path is a good idea. I will test that. But for the other file formats: as far as I know none of them would trigger a kernel request as we are not handling those formats internally.

src/ExtranetBundle/objects/db/TDataExtranetUser.class.php
bestform: This IS the master request. We are not dealing with sub requests here. It is the actual request done by the browser when loading a resource referenced in the html

src/ExtranetBundle/objects/db/TDataExtranetUser.class.php
UlrichKu: gif is also a possible request (missing image).

Is the pathInfo always lower case?

bestform:
We've found a better solution. A new PR has been made

PR #650, opened 2022-07-05
415/better error handling in module loading

pixeljunkie:
| Q             | A
| ------------- | ---
| Branch        | 7.1.x
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed issues  | chameleon-system/chameleon-system#415
| License       | MIT

Adds better exception handling if a module is configured to an ID that does not exist as module.
This may happen if a theme is installed.

### DIFF
--- a/src/CoreBundle/private/core/TModuleLoader.class.php
+++ b/src/CoreBundle/private/core/TModuleLoader.class.php
@@ -353,7 +353,7 @@ class TModuleLoader
             $this->logModuleException($e, $spotName);
         } catch (\Exception $e) {
             if (_DEVELOPMENT_MODE) {
-                throw new ModuleExecutionFailedException('Error in module execution: '.$e->getMessage().' in file: '.$e->getFile().' on line: '.$e->getLine(), 0, $e);
+                throw new ModuleExecutionFailedException(sprintf('Error in module execution: %s in file: %s on line: %d',$e->getMessage(), $e->getFile(), $e->getLine()), 0, $e);
             }
 
             $this->logModuleException($e, $spotName);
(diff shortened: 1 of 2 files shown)


src/CoreBundle/private/library/classes/dbobjects/TCMSTPLModule.class.php
```diff
@@ -210,7 +210,13 @@ public function getViewMapperConfig()
         if (null !== $viewMapperConfig) {
             return $viewMapperConfig;
         }
-        $viewMapperConfig = new ViewMapperConfig($this->sqlData['view_mapper_config']);
+
+        try {
+            $viewMapperConfig = $this->sqlData['view_mapper_config'];
```
UlrichKu, line 215: What kind of Exceptions can happen here?

a \array_key_exists would be sufficient?

src/CoreBundle/private/core/TModuleLoader.class.php
```diff
@@ -353,7 +353,7 @@ public function GetModule($spotName, $bReturnString = false, $sCustomWrapping =
             $this->logModuleException($e, $spotName);
         } catch (\Exception $e) {
             if (_DEVELOPMENT_MODE) {
-                throw new ModuleExecutionFailedException('Error in module execution: '.$e->getMessage(), 0, $e);
+                throw new ModuleExecutionFailedException('Error in module execution: '.$e->getMessage().' in file: '.$e->getFile().' on line: '.$e->getLine(), 0, $e);
```
UlrichKu, line 356: Use sprintf here?

src/CoreBundle/private/library/classes/dbobjects/TCMSTPLModule.class.php
pixeljunkie: It is false by default (if not initialized). I changed the check to handle bool and missing array key here.

PR #656, opened 2022-07-22
ref #784: Return an empty user if there is no session

UlrichKu:
| Q             | A
| ------------- | ---
| Branch        | 7.1.x
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | yes
| Tests pass?   | yes
| Fixed issues  | chameleon-system/chameleon-system#784
| License       | MIT


src/ExtranetBundle/Service/ExtranetUserProvider.php
```diff
@@ -37,53 +37,43 @@ public function __construct(RequestStack $requestStack)
...
-        $hasSession = false;
         $request = $this->requestStack->getCurrentRequest();
-        if (null !== $request) {
-            $hasSession = $request->hasSession();
+        if (null === $request) {
+            return null;
+        }
+        if (false === $request->getSession()->isStarted()) {
+            return null;
         }
```
bestform, line 75: We should check for `$request->hasSession()` before getting the session for the request as it might not be set.

bestform (changes requested):

src/ExtranetBundle/Service/ExtranetUserProvider.php
UlrichKu: ok

PR #415, opened 2020-01-03
Feature/107 upgrade backend navigation tree

ckleis:
| Q             | A
| ------------- | ---
| Branch        | 6.3.x
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Fixed issues  | chameleon-system/chameleon-system#107
| License       | MIT

Upgrade of Plugin JSTREE.

This new PR replaces the broken https://github.com/chameleon-system/chameleon-base/pull/402
The comments from Pixeljunkie from the previous PR are already included here.

PR #657, opened 2022-07-25
ref #415: throw useful error if a mapper config is missing

UlrichKu:
| Q             | A
| ------------- | ---
| Branch        | 7.1.x
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed issues  | chameleon-system/chameleon-system#415
| License       | MIT

### DIFF
--- a/src/CoreBundle/private/core/TModelBase.class.php
+++ b/src/CoreBundle/private/core/TModelBase.class.php
@@ -700,10 +700,7 @@ class TModelBase
             if (null === $sViewPathReference) {
                 throw new MapperException(sprintf('No config found for %s for view %s', $this->aModuleConfig['model'], $sViewName));
             }
-            $mappers = $viewMapperConfig->getMappersForConfig($sViewName);
-            if (null === $mappers) {
-                $mappers = []; // the actual error is checked above with the view path reference
-            }
+            $mappers = $viewMapperConfig->getMappersForConfig($sViewName) ?? [];
             foreach ($mappers as $sMapperName) {
                 $oRenderer->addMapperFromIdentifier(
                     $sMapperName,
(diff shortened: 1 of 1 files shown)


src/CoreBundle/private/core/TModelBase.class.php
```diff
@@ -697,6 +697,9 @@ final public function __invoke(Request $request, $isLegacyModule)
             $sViewName = $this->aModuleConfig['view'];
             $viewMapperConfig = $oModule->getViewMapperConfig();
             $mappers = $viewMapperConfig->getMappersForConfig($sViewName);
+            if (null === $mappers) {
+                throw new MapperException(sprintf('No mappers found/config not found for %s for view %s', $this->aModuleConfig['model'], $sViewName));
+            }
```
bestform, line 702: A return of `null` indicates that the whole config is missing. It is possible for a config to have no mappers and this would not be an error condition.
Therefore in addition we should check `$sViewPathReference` in line 708 for null as this would indicate the same thing - being there isn't a config for this view name.

bestform (changes requested):

src/CoreBundle/private/core/TModelBase.class.php
UlrichKu: I moved the check to the view path (and that up).

bestform: ah yes, this is better. But I would still throw an exception if the the mappers are null, because this would indicate an error that can happen after restructuring the code. Currently it is a manual invariant and might be ok, but in the future it might not be.

bestform (changes requested):

src/CoreBundle/private/core/TModelBase.class.php
UlrichKu: I wouldn't say that null is different from an empty array here.

Both signify, there "isn't anything".

Both simply do not iterate now.

bestform: Yes, but the implementation uses `null` as a sentinal value for "there is no such config". No mappers would be indicated by an empty array. Yes, this is bad API design, but what can you do. ;)

UlrichKu: Don't know the implementation from the outside...

UlrichKu: Inserted the same check again.

bestform (approved):

PR #658, opened 2022-07-25
use different session for different portals

bestform:
if CHAMELEON_EXTRANET_USER_IS_PORTAL_DEPENDANT is set.

| Q             | A
| ------------- | ---
| Branch        | 7.1.x
| Bug fix?      | yes
| New feature?  | no 
| BC breaks?    | no     
| Deprecations? | no 
| Tests pass?   | yes    
| Fixed issues  | chameleon-system/chameleon-system#787  
| License       | MIT

We are now using a different session cookie for each portal if CHAMELEON_EXTRANET_USER_IS_PORTAL_DEPENDANT is set. This not only fixes the bug in #787, but also finally allows the user to be logged in to each portal separately - as it should be.

### DIFF
--- a/CHANGELOG-7.1.md
+++ b/CHANGELOG-7.1.md
@@ -3,8 +3,17 @@ CHANGELOG For 7.1.x
 
 # New Features
 
+In v7.1.4 there are separate sessions created for frontend and backend.
+If `CHAMELEON_EXTRANET_USER_IS_PORTAL_DEPENDANT` is set to `true`, each portal will have its own session as well.
+
 # Changed Features
 
 ## Field Type "Document manager" extends `TCMSFieldLookupMultiselect`
 
 This changes public methods - especially `GetMLTTableName`.
+
+# Changed Interfaces
+
+In v7.1.4 `\ChameleonSystem\CoreBundle\Service\RequestInfoServiceInterface` has an additional method:
+`isPreviewMode();`
+Implementations should return true if the current request is a frontend request and the current page is a preview page.
--- a/CHANGELOG-7.1.md
+++ b/CHANGELOG-7.1.md
@@ -16,4 +16,4 @@ This changes public methods - especially `GetMLTTableName`.
 
 In v7.1.4 `\ChameleonSystem\CoreBundle\Service\RequestInfoServiceInterface` has an additional method:
 `isPreviewMode();`
-Implementations should return true if the current request is a frontend request and the current page is a preview page.
+Implementations should return true if the current request is a frontend request and the preview mode is active.
--- a/src/CoreBundle/Service/RequestInfoService.php
+++ b/src/CoreBundle/Service/RequestInfoService.php
@@ -146,7 +146,7 @@ class RequestInfoService implements RequestInfoServiceInterface
             return false;
         }
 
-        // todo: `__rpeviewmode` should be the only way to enable this. Refactor all places where the preview attribute is set as `preview` instead of `__previewmode'
+        // todo: `__previewmode` should be the only way to enable this. Refactor all places where the preview attribute is set as `preview` instead of `__previewmode'
         $this->isPreviewModeCache = false === \TGlobal::IsCMSMode() &&
             ('true' === $request->get('__previewmode') || 'true' === $request->get('preview'));
 
--- a/src/CoreBundle/Service/RequestInfoService.php
+++ b/src/CoreBundle/Service/RequestInfoService.php
@@ -54,7 +54,10 @@ class RequestInfoService implements RequestInfoServiceInterface
      * @var bool
      */
     private $isCmsTemplateEngineEditModeCache;
-
+    /**
+     * @var bool
+     */
+    private $isPreviewModeCache;
     /**
      * @param RequestStack                 $requestStack
      * @param PortalDomainServiceInterface $portalDomainService
@@ -127,6 +130,27 @@ class RequestInfoService implements RequestInfoServiceInterface
         $this->isCmsTemplateEngineEditModeCache = false === \TGlobal::IsCMSMode() && 'true' === $request->get('__modulechooser');
 
         return $this->isCmsTemplateEngineEditModeCache;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function isPreviewMode()
+    {
+        if (null !== $this->isPreviewModeCache) {
+            return $this->isPreviewModeCache;
+        }
+
+        $request = $this->requestStack->getCurrentRequest();
+        if (null === $request) {
+            return false;
+        }
+
+        // todo: `__rpeviewmode` should be the only way to enable this. Refactor all places where the preview attribute is set as `preview` instead of `__previewmode'
+        $this->isPreviewModeCache = false === \TGlobal::IsCMSMode() &&
+            ('true' === $request->get('__previewmode') || 'true' === $request->get('preview'));
+
+        return $this->isPreviewModeCache;
     }
 
     /**
--- a/src/CoreBundle/Service/RequestInfoService.php
+++ b/src/CoreBundle/Service/RequestInfoService.php
@@ -51,11 +51,11 @@ class RequestInfoService implements RequestInfoServiceInterface
      */
     private $chameleonRequestType;
     /**
-     * @var bool
+     * @var bool|null
      */
     private $isCmsTemplateEngineEditModeCache;
     /**
-     * @var bool
+     * @var bool|null
      */
     private $isPreviewModeCache;
     /**
--- a/src/CoreBundle/Service/RequestInfoService.php
+++ b/src/CoreBundle/Service/RequestInfoService.php
@@ -148,7 +148,7 @@ class RequestInfoService implements RequestInfoServiceInterface
 
         // todo: `__previewmode` should be the only way to enable this. Refactor all places where the preview attribute is set as `preview` instead of `__previewmode'
         $this->isPreviewModeCache = false === \TGlobal::IsCMSMode() &&
-            ('true' === $request->get('__previewmode') || 'true' === $request->get('preview'));
+            ('true' === $request->query->get('__previewmode') || 'true' === $request->query->get('preview'));
 
         return $this->isPreviewModeCache;
     }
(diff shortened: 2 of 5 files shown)


src/CoreBundle/Session/ChameleonSessionManager.php
```diff
@@ -63,17 +64,19 @@ class ChameleonSessionManager implements ChameleonSessionManagerInterface
...
 
     /**
-     * @param RequestStack             $requestStack
-     * @param Connection               $databaseConnection
-     * @param PDO                      $sessionPdo
-     * @param ContainerInterface       $container
-     * @param int                      $metaDataTimeout
-     * @param array                    $sessionOptions
-     * @param InputFilterUtilInterface $inputFilterUtil
+     * @param RequestStack                  $requestStack
```
j6s, line 70: I know this is just a reformat, but this docblock adds 0 value to the code since it just repeats the type hints. It could be narrowed down to just one type, which should be more clearly defined though: The array format of `$sessionOptions` (Which is `array<string, mixed>` I suppose?)

src/CoreBundle/Session/ChameleonSessionManager.php
```diff
@@ -175,4 +182,21 @@ public function isSessionStarting()
     {
         return $this->isSessionStarting;
     }
+
+    /**
+     * @return string
+     */
+    private function getCookieNameSuffix()
+    {
+        if (!CHAMELEON_EXTRANET_USER_IS_PORTAL_DEPENDANT) {
```
j6s, line 191: Can we assume that the constant is always defined or would adding a `defined(...)` check make sense?

src/CoreBundle/Session/ChameleonSessionManager.php
```diff
@@ -142,6 +146,9 @@ public function boot()
 
         $session = new \TPKgCmsSession($sessionStorage);
 
+        $cookieName = "PHPSESSID" . $this->getCookieNameSuffix();
```
j6s, line 149: This is a bit of a nitpick, but `PHPSESSID` should probably be moved into a constant. Sadly, PHP doesn't seem to have a built-in one for it.

j6s:
In general, this changes the behaviour of portals using this constant a bit, so I suppose this should be mentioned in a changelog?

src/CoreBundle/Session/ChameleonSessionManager.php
bestform: Yes, I thought about that for a second but then forgot. Done.

src/CoreBundle/Session/ChameleonSessionManager.php
bestform: Yes, we've got a default in every project in `/src/CoreBundle/Resources/config/defaults.inc.php:651`

src/CoreBundle/Session/ChameleonSessionManager.php
bestform: done

bestform:
> In general, this changes the behaviour of portals using this constant a bit, so I suppose this should be mentioned in a changelog?

Considering this basically fixes the usage of this constant I guess a Changelog isn't really needed. The user does not need to change anything. The difference is that it works now when the second portal is defined by a path. This was just broken before.

src/CoreBundle/Session/ChameleonSessionManager.php
j6s: Ah, we missed `$metaDataTimeout` which does not have a type annotation - it is also required in the docblock

src/CoreBundle/Service/RequestInfoService.php
```diff
@@ -129,6 +132,27 @@ public function isCmsTemplateEngineEditMode()
...
+        if (null !== $this->isPreviewModeCache) {
+            return $this->isPreviewModeCache;
+        }
+
+        $request = $this->requestStack->getCurrentRequest();
+        if (null === $request) {
+            return false;
+        }
+
+        // todo: `__rpeviewmode` should be the only way to enable this. Refactor all places where the preview attribute is set as `preview` instead of `__previewmode'
```
UlrichKu, line 149:
```suggestion
        // todo: `__previewmode` should be the only way to enable this. Refactor all places where the preview attribute is set as `preview` instead of `__previewmode'
```

src/CoreBundle/Service/RequestInfoService.php
```diff
@@ -129,6 +132,27 @@ public function isCmsTemplateEngineEditMode()
...
+        $this->isPreviewModeCache = false === \TGlobal::IsCMSMode() &&
+            ('true' === $request->get('__previewmode') || 'true' === $request->get('preview'));
```
UlrichKu, line 151: Use InputFilterUtil here?

This is always a GET parameter? (could use the specific method then)

src/CoreBundle/Service/RequestInfoService.php
```diff
@@ -54,7 +54,10 @@ class RequestInfoService implements RequestInfoServiceInterface
      * @var bool
      */
     private $isCmsTemplateEngineEditModeCache;
-
+    /**
+     * @var bool
+     */
+    private $isPreviewModeCache;
```
j6s, line 60: Actually, this is `bool|null`, not just `bool`

src/CoreBundle/Session/ChameleonSessionManager.php
```diff
@@ -175,4 +178,35 @@ public function isSessionStarting()
...
+     * @return string
+     */
+    private function getCookieName()
+    {
+        if ($this->requestInfoService->isBackendMode()) {
+            return self::SESSION_BASE_NAME . 'CMS';
+        }
+
+        if ($this->requestInfoService->isCmsTemplateEngineEditMode()) {
+            $this->logger->notice('Using backend session because we are in template engine edit mode');
```
UlrichKu, line 192: Why (notice) logging)? This is normal system function and nothing noteable?

you could join the 3 cases then which makes it more clear that the behave the same.

src/CoreBundle/Service/RequestInfoService.php
j6s, line 149: TODO: Fix `__rpeviewmode` typo

src/CoreBundle/Service/RequestInfoServiceInterface.php
```diff
@@ -38,12 +38,19 @@ public function getChameleonRequestType();
     public function isChameleonRequestType($requestType);
 
     /**
-     * Returns true if the current request is a backend request AND the current page is the template engine.
+     * Returns true if the current request is a frontend request and the current page is the template engine.
```
j6s, line 41: Both of these comments are a bit confusing, can we make clearer what the method does?

src/CoreBundle/Session/ChameleonSessionManager.php
j6s: This was the result of a discussion we had about this topic. Because these methods are triggered globally by the existence of certain GET-Parameters, collisions with parameters in custom logic is possible. This would lead to very weird session issues. To make debugging them easier, a notice is being logged for these instances.

src/CoreBundle/Service/RequestInfoService.php
bestform: fixed

src/CoreBundle/Service/RequestInfoService.php
bestform: We are not using the parameter in any way, we are just comparing it to a static string. So I don't see how an input filter would benefit us. Or do I miss something?

Yes, this is always a GET parameter. And I think it is very clear what is being checked here. I don't thing a separate method would benefit us much.

src/CoreBundle/Service/RequestInfoService.php
bestform: fixed

src/CoreBundle/Service/RequestInfoService.php
bestform: fixed

src/CoreBundle/Service/RequestInfoServiceInterface.php
bestform: reworded and added usage so it should clearer what's going on.

CHANGELOG-7.1.md
```diff
@@ -3,8 +3,17 @@ CHANGELOG For 7.1.x
...
 
-This changes public methods - especially `GetMLTTableName`.
\ No newline at end of file
+This changes public methods - especially `GetMLTTableName`.
+
+# Changed Interfaces
+
+In v7.1.4 `\ChameleonSystem\CoreBundle\Service\RequestInfoServiceInterface` has an additional method:
+`isPreviewMode();`
+Implementations should return true if the current request is a frontend request and the current page is a preview page.
```
UlrichKu, line 19:
```suggestion
Implementations should return true if the current request is a frontend request and the preview mode is active.
```

src/CoreBundle/Service/RequestInfoService.php
UlrichKu: Specific: I meant $request->query->get().

src/CoreBundle/Session/ChameleonSessionManager.php
UlrichKu: Hm, not sure I would expect anything helping me debug as a notice in a log...

src/CoreBundle/Service/RequestInfoService.php
bestform: ah, i see. Yes, that makes sense. done

src/CoreBundle/Session/ChameleonSessionManager.php
bestform: done

CHANGELOG-7.1.md
bestform: done

src/CoreBundle/Session/ChameleonSessionManager.php
j6s: It doesn't hurt, does it? And potentially it saves us a bit of deuabugging in the future

PR #677, opened 2023-01-30
change 7.2 to 8.0 in the upgrade documentation

rnuernberg:


UPGRADE-8.0.md
```diff
@@ -112,7 +112,7 @@ Remove the file `app/autoload.php`. It is no longer used by the system (see belo
 The functionality "annotation support" was removed. This file was calling a
 deprecated function `AnnotationRegistry::registerLoader()`. If needed annotations can still be configured and used
 directly in a project.
-However with php 8 you should use attributes instead.
+However with php > 8 you should use attributes instead.
```
bestform, line 115: Technically it is php >= 8

bestform (changes requested):

UPGRADE-8.0.md
rnuernberg: we require php 8.1 in the new version - so I think this should be ok

PR #686, opened 2023-03-28
ref #815: fix interface to method mismatch for PortalDomainService, p…

pixeljunkie:
…revent unsetting an already set portal to prevent crash in ChangeActivePortalEvent

| Q             | A
| ------------- | ---
| Branch        | 7.1.x
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Fixed issues  | chameleon-system/chameleon-system#815
| License       | MIT

### DIFF
--- a/src/CoreBundle/Service/PortalDomainServiceInterface.php
+++ b/src/CoreBundle/Service/PortalDomainServiceInterface.php
@@ -86,11 +86,9 @@ interface PortalDomainServiceInterface
     /**
      * Sets the currently active portal.
      *
-     * @param TCMSPortal $portal
-     *
      * @return void
      */
-    public function setActivePortal(TCMSPortal $portal);
+    public function setActivePortal(TCMSPortal $portal = null);
 
     /**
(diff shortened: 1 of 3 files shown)


src/CoreBundle/Service/PortalDomainServiceInterface.php
```diff
@@ -86,11 +86,9 @@ public function hasPrimaryDomain($portalId = null, $languageId = null);
     /**
      * Sets the currently active portal.
      *
-     * @param TCMSPortal $portal
-     *
      * @return void
      */
-    public function setActivePortal(TCMSPortal $portal);
+    public function setActivePortal(TCMSPortal $portal = null);
```
rnuernberg, line 91: If you allow null, then the method should be able to handle a call with null. So the event should take null as a new portal.

pixeljunkie: i changed the event to handle null values

PR #693, opened 2023-08-10
827 add head include modification methods to event

pixeljunkie:
| Q             | A
| ------------- | ---
| Branch        | 7.1.x
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed issues  | chameleon-system/chameleon-system#827
| License       | MIT

### Summary by CodeRabbit

**New Feature:**
- Added `removeDataElement` method to `HtmlIncludeEvent` class and `HtmlIncludeEventInterface` interface for removing specific elements from the data structure.
- Introduced `updateDataElement` method to `HtmlIncludeEvent` class and `HtmlIncludeEventInterface` interface for updating specific elements in the data structure.

> 🎉🐇
> 
> "Hopping through the code, with a swift rabbit's grace,
> 
> We've added new methods, to update and erase.
> 
> In our data structure, changes we now can trace,
> 
> Celebrate this feature, it's no small chase!" 🥕🎉

### DIFF
--- a/src/CoreBundle/Event/HtmlIncludeEvent.php
+++ b/src/CoreBundle/Event/HtmlIncludeEvent.php
@@ -58,9 +58,15 @@ class HtmlIncludeEvent extends Event implements HtmlIncludeEventInterface
         return $this->data;
     }
 
-    public function removeDataElement(string $key): void
+    public function removeDataElement(string $key): bool
     {
+        if (!isset($this->data[$key]) {
+            return false;
+        }
+    
         unset($this->data[$key]);
+
+        return true;
     }
 
     public function updateDataElement(string $key, string $value): void
@@ -70,7 +76,6 @@ class HtmlIncludeEvent extends Event implements HtmlIncludeEventInterface
 
     /**
      * @param mixed $value
-     *
      * @psalm-assert-if-true string $value
      */
     private function isInteger($value): bool
--- a/src/CoreBundle/Event/HtmlIncludeEvent.php
+++ b/src/CoreBundle/Event/HtmlIncludeEvent.php
@@ -58,13 +58,22 @@ class HtmlIncludeEvent extends Event implements HtmlIncludeEventInterface
         return $this->data;
     }
 
+    public function removeDataElement(string $key): void
+    {
+        unset($this->data[$key]);
+    }
+
+    public function updateDataElement(string $key, string $value): void
+    {
+        $this->data[$key] = $value;
+    }
+
     /**
      * @param mixed $value
      *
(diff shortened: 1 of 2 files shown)


src/CoreBundle/Event/HtmlIncludeEvent.php
```diff
@@ -58,13 +58,22 @@ public function getData()
...
+    }
+
+    public function updateDataElement(string $key, string $value): void
+    {
+        $this->data[$key] = $value;
+    }
+
     /**
      * @param mixed $value
      *
```
finn-matti, line 73: If you remove the param here, the empty line in the comment can be removed as well.

src/CoreBundle/Event/HtmlIncludeEvent.php
```diff
@@ -58,13 +58,22 @@ public function getData()
         return $this->data;
     }
 
+    public function removeDataElement(string $key): void
+    {
+        unset($this->data[$key]);
```
finn-matti, line 63: unset does not produce an error when unsetting a key that doesn't exist, however: Is this the functionality we want? Does the user (or the program somewhere higher in the stack) need to know about a failed removeDataElement attempt?

finn-matti (commented):
Only two small things. Can be merged as is, but you may want to address them beforehand.

src/CoreBundle/Event/HtmlIncludeEvent.php
pixeljunkie: done

src/CoreBundle/Event/HtmlIncludeEvent.php
pixeljunkie: added boolen return value (this may be interesting for debugging if a key really existed or not)

src/CoreBundle/Event/HtmlIncludeEvent.php
pixeljunkie: done

src/CoreBundle/Event/HtmlIncludeEvent.php
```diff
@@ -58,13 +58,27 @@ public function getData()
...
+    
+        unset($this->data[$key]);
+
+        return true;
+    }
+
+    public function updateDataElement(string $key, string $value): void
+    {
+        $this->data[$key] = $value;
+    }
```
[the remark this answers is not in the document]
pixeljunkie: added boolean return value

PR #703, opened 2023-12-06
ref #803: remove window open on wysiwyg download links

pixeljunkie:
| Q             | A
| ------------- | ---
| Branch        | 7.1.x
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | yes
| Deprecations? | no
| Tests pass?   | yes 
| Fixed issues  | chameleon-system/chameleon-system#835
| License       | MIT


pixeljunkie:
I marked it as BC break, but only because it removes a feature. But this feature is outdated and produces problems, so i think we can live with it. You would not implement a window open as inline onclick event today.

PR #713, opened 2024-01-16
ref #62363: when copying mlt, we should do that in the order that the…

rnuernberg:


src/CoreBundle/private/library/classes/TCMSTableEditor/TCMSTableEditorEndPoint.class.php
```diff
@@ -1793,21 +1793,18 @@ public function CopyMLTRecords($oField, $sourceRecordID)
     {
         $mltTableName = $oField->GetMLTTableName();
         $oField->oTableRow->id = $sourceRecordID;
-        $oFieldType = $oField->oDefinition->GetFieldType();
-        if ('CMSFIELD_DOCUMENTS' == $oFieldType->sqlData['constname']) {
```
pixeljunkie, line 1797: This was removed by intention?

rnuernberg: yes - there is no need to load the foreign table, so we no longer need to know what it is called. We only need the ID, which we can get from the target_id field

PR #721, opened 2024-02-07
ref #841: add option to disable the post login hook and remove duplic…

pixeljunkie:
…ate call in DirectLogin

| Q             | A
| ------------- | ---
| Branch        | 7.1.x
| Bug fix?      | yes
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Fixed issues  | chameleon-system/chameleon-system#841
| License       | MIT

### DIFF
--- a/src/ExtranetBundle/objects/db/TDataExtranetUser.class.php
+++ b/src/ExtranetBundle/objects/db/TDataExtranetUser.class.php
@@ -91,6 +91,8 @@ class TDataExtranetUser extends TDataExtranetUserAutoParent
      * @var array
      */
     protected $pageAccessCache = array();
+    
+    protected bool $callPostLoginHook = true;
 
     /**
      * @param string|null $id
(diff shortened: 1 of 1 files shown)


src/ExtranetBundle/objects/db/TDataExtranetUser.class.php
```diff
@@ -598,9 +605,8 @@ public function DirectLogin($sLoginName, $sPassword, $callPostLoginHook = false)
         $this->AllowEditByAll(true);
         $bWasLoggedIn = $this->Login($sLoginName, $sPassword);
         $this->AllowEditByAll(false);
-        if ($bWasLoggedIn && $callPostLoginHook) {
-            $this->PostLoginHook();
-        }
+        
+        $this->setCallPostLoginHook($callPostLoginHook);
```
mhaendler, line 609: At this point, setting setCallPostLoginHook is too late, as the postLoginHook is already called within the login function. 

As a result, the value is "true" by default and the PostLoginHook is always executed although we get from the parameter that it should be false?

pixeljunkie: You are absolutely right. Fixed the position. Good catch! Thanks!

